PR #50, opened 2022-10-21
fix navigation task robot reset logic

ChengshuLi:
There is definitely a cleaner way of doing this, but I am currently just restoring what we have in iG 2 to make it functional for now.


Doitright124:
Hi

PR #85, opened 2023-01-14
add checking function for unfolded

ChengshuLi:


omnigibson/examples/object_states/folded_unfolded_state_demo.py
```diff
@@ -1,5 +1,5 @@
 from omnigibson.utils.constants import PrimType
-from omnigibson.object_states import Folded
+from omnigibson.object_states import Folded, Unfolded
```
cremebrule, line 2: high level question: what was the reason for having an explicit separate state rather than having `Folded` be tertiary?

omnigibson/utils/object_state_utils.py
```diff
@@ -220,3 +221,86 @@ def sample_kinematics(
...
+
+
+# Folded / Unfolded related utils
+m.DEBUG_CLOTH_PROJ_VIS = False
+m.NORMAL_Z_ANGLE_DIFF = np.deg2rad(30.0)
+
+def calculate_projection_area_and_diagonal_unfolded(obj):
+    """
+    Calculate the maximum projection area and the diagonal length along different axes in the unfolded state.
+    Should be called in the initialize function. Assume the object's default pose is unfolded.
```
cremebrule, line 233: outdated comment

cremebrule (approved):
minor comment, otherwise LGTM!

ChengshuLi:
Remember to call ConvextHull.volume instead of ConvexHull.area to get the actual area. 

Ref: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.ConvexHull.html

ChengshuLi:
Can we merge this?

cremebrule:
Yep!

PR #87, opened 2023-01-17
Refactor isaac sim 2022.2.0

cremebrule:
Minimal refactoring to unblock other development thrusts

### DIFF
--- a/omnigibson/sensors/vision_sensor.py
+++ b/omnigibson/sensors/vision_sensor.py
@@ -3,26 +3,39 @@ import numpy as np
 import time
 import gym
 
-from omnigibson import app
+import omnigibson as og
 from omnigibson.sensors.sensor_base import BaseSensor
 from omnigibson.utils.constants import MAX_CLASS_COUNT, MAX_INSTANCE_COUNT, MAX_VIEWER_SIZE, VALID_OMNI_CHARS
 from omnigibson.utils.python_utils import assert_valid_key, classproperty
+from omnigibson.utils.sim_utils import set_carb_setting
+from omnigibson.utils.ui_utils import dock_window
 from omnigibson.utils.usd_utils import get_camera_params
 from omnigibson.utils.transform_utils import euler2quat, quat2euler
 
-import carb
+import omni.ui
 from omni.isaac.core.utils.stage import get_current_stage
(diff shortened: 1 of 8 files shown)


omnigibson/sensors/vision_sensor.py
```diff
@@ -25,6 +28,16 @@
...
 
 
+# Duplicate of simulator's render method, used so that this can be done before simulator is created!
+def render():
+    """
+    Refreshes the Isaac Sim app rendering components including UI elements and view ports..etc.
+    """
+    set_carb_setting(og.app._carb_settings, "/app/player/playSimulations", False)
+    og.app.update()
+    set_carb_setting(og.app._carb_settings, "/app/player/playSimulations", True)
```
ChengshuLi, line 38: Sorry would you mind adding a more detailed comment explaining why this is needed?

omnigibson/utils/asset_utils.py
```diff
@@ -427,3 +427,5 @@ def encrypt_file(original_filename, encrypted_filename=None, encrypted_file=None
         download_og_dataset()
     elif args.change_data_path:
         change_data_path()
+
+    og.shutdown()
```
ChengshuLi, line 431: Why do we need to add this line? A better question to ask is: do we need to add this line to the end of every example?

ChengshuLi (approved):
LGTM! I left some minor comments. Thanks!!

omnigibson/sensors/vision_sensor.py
cremebrule: yeah -- when OG first starts up, we automatically create / initialize the `Simulator` singleton class. However, mid-initialization, we want to register the "main" viewer camera, which is a `VisionSensor` instance. During the `VisionSensor` initialization, it needs to call `og.sim.render()` in order to propagate some changes, but it can't do so because we're still in the middle of the simulator's initialization process.

omnigibson/utils/asset_utils.py
cremebrule: IIRC without this line the process simply hangs or raises an error. Calling shutdown actually terminates the process properly. I should have already added the shutdown call (equivalently, `env.close()`) to the end of all our examples

PR #89, opened 2023-01-20
implement OnFire as a child of HeatSourceOrSink

ChengshuLi:


omnigibson/object_states/heat_source_or_sink.py
```diff
@@ -97,37 +93,27 @@ def get_optional_dependencies():
...
-        # Run super first
-        super()._initialize()
-        self.initialize_link_mixin()
-
-    def _update(self):
-        self.status, self.position = self._compute_state_and_position()
-
-    def _get_value(self):
-        return self.status, self.position
+        return True
```
cremebrule, line 114: is the position not needed anymore because it's static (i.e.: it doesn't change?)

omnigibson/object_states/temperature.py
```diff
@@ -41,17 +44,30 @@ def _set_value(self, new_value):
...
         # Start at the current temperature.
         new_temperature = self.value
 
         # Find all heat source objects.
         affected_by_heat_source = False
-        for obj2 in self._simulator.scene.get_objects_with_state(HeatSourceOrSink):
-            # Obtain heat source position.
-            heat_source = obj2.states[HeatSourceOrSink]
-            heat_source_state, heat_source_position = heat_source.get_value()
+        heat_sources = og.sim.scene.get_objects_with_state(HeatSourceOrSink) | \
```
cremebrule, line 55: maybe make this an external utility call? That way we can easily find it and add to it if we add other heat sources in the future

omnigibson/object_states/temperature.py
```diff
@@ -41,17 +44,30 @@ def _set_value(self, new_value):
...
+                       og.sim.scene.get_objects_with_state(OnFire)
+        for obj2 in heat_sources:
+            # Only external heat sources will affect the temperature.
+            if obj2 == self.obj:
+                continue
+
+            heat_source = obj2.states[OnFire] if OnFire in obj2.states else obj2.states[HeatSourceOrSink]
```
cremebrule, line 62: Does this imply that objects will either be only OnFire or HeatSourceOrSink?

omnigibson/examples/object_states/onfire_demo.py
```diff
@@ -0,0 +1,103 @@
...
+    apples[0].set_position(stove.states[object_states.HeatSourceOrSink].get_link_position() + np.array([-0.15, 0, 0.1]))
+
+    # The second apple will NOT be affected by the stove, but will be affected by the first apple once it's on fire.
+    apples[1].set_position(stove.states[object_states.HeatSourceOrSink].get_link_position() + np.array([-0.3, 0, 0.1]))
+
+    steps = 0
+    max_steps = -1 if not short_exec else 1000
+
+    # Main recording loop
+    while steps != max_steps:
```
cremebrule, line 92: such a cool demo!! awesome

omnigibson/object_states/on_fire.py
```diff
@@ -0,0 +1,79 @@
...
+            ignition_temperature (float): The temperature threshold above which on fire will become true.
+            fire_temperature (float): The temperature of the fire (heat source) once on fire is true.
+            heating_rate (float): Fraction in [0, 1] of the temperature difference with the
+                heat source temperature should be received every step, per second.
+            distance_threshold (float): The distance threshold which an object needs
+                to be closer than in order to receive heat from this heat source.
+        """
+        assert fire_temperature > ignition_temperature, "fire temperature should be higher than ignition temperature."
+
+        super(OnFire, self).__init__(
```
cremebrule, line 48: should just be `super().__init__(...)` just to be consistent with python3?

cremebrule (approved):
A few minor comments, but overall LGTM!

omnigibson/object_states/heat_source_or_sink.py
ChengshuLi: We still need the position. It's just that we don't return position from the `get_value` call anymore. The code that uses the position will just do `state.get_link_position()`. I think this way we unify with other object states better (`HeatSourceOrSink` is the only state that returns two things from `get_value`).

omnigibson/object_states/on_fire.py
ChengshuLi: done

omnigibson/object_states/temperature.py
ChengshuLi: I added the get_objects_with_state_recursive function call. I think it's safe to assume (at least for now) that all heat source related states would be subclasses of `HeatSourceOrSink`.

omnigibson/object_states/temperature.py
ChengshuLi: Added a sanity check.

ChengshuLi:
@cremebrule I addressed your comments. PTAL. Thanks!

omnigibson/object_states/heat_source_or_sink.py
cremebrule: good point!

cremebrule (approved):
LGTM!

PR #102, opened 2023-02-22
Unit tests for object states

ChengshuLi:


omnigibson/object_states/__init__.py
```diff
@@ -19,40 +19,6 @@
...
 from omnigibson.object_states.particle_modifier import ParticleRemover, ParticleApplier
 from omnigibson.object_states.pose import Pose
-from omnigibson.object_states.robot_related_states import (
-    InFOVOfRobot,
-    InHandOfRobot,
-    InReachOfRobot,
-    InSameRoomAsRobot,
-    ObjectsInFOVOfRobot,
-)
-from omnigibson.object_states.room_states import (
```
cremebrule, line 29: cya!

omnigibson/object_states/folded.py
```diff
@@ -45,9 +45,10 @@ def _get_value(self):
...
 
     def _set_value(self, new_value):
-        """
-        Set the folded state. Currently, it's not supported yet.
-        """
-        raise NotImplementedError("_set_value of the Folded state has not been implemented")
+        if not new_value:
+            raise NotImplementedError("Folded does not support set_value(False)")
+
+        # TODO (eric): add this support
```
cremebrule, line 51: will we add support for programmatic unfolding too, or just folding?

omnigibson/object_states/open.py
```diff
@@ -76,37 +77,33 @@ def _get_relevant_joints(obj):
             - list of int: Joint directions for each corresponding relevant joint
     """
     global m
-    if not hasattr(obj, "metadata"):
-        return None, None, None
 
-    both_sides = obj.metadata[m.BOTH_SIDES_METADATA_FIELD] if m.BOTH_SIDES_METADATA_FIELD in obj.metadata else False
+    default_both_sides = False
+    default_relevant_joints = list(obj.joints.values())
+    default_joint_directions = [1] * len(default_relevant_joints)
```
cremebrule, line 83: can we add a comment here? i assume 1 --> open direction corresponds to positive joint angle change, and -1 --> open direction corresponds to negative joint angle change?

omnigibson/object_states/open.py
```diff
@@ -76,37 +77,33 @@ def _get_relevant_joints(obj):
...
     if m.METADATA_FIELD not in obj.metadata:
-        print("No openable joint metadata found for object %s" % obj.name)
-        return None, None, None
+        logging.warning("No openable joint metadata found for object %s" % obj.name)
+        return default_both_sides, default_relevant_joints, default_joint_directions
 
+    both_sides = obj.metadata[m.BOTH_SIDES_METADATA_FIELD] if m.BOTH_SIDES_METADATA_FIELD in obj.metadata else False
     joint_metadata = obj.metadata[m.METADATA_FIELD].items()
 
     # The joint metadata is in the format of [(joint_id, joint_name), ...] for legacy annotations and
```
cremebrule, line 97: do we still have legacy annotations?

omnigibson/object_states/open.py
```diff
@@ -76,37 +77,33 @@ def _get_relevant_joints(obj):
             - list of int: Joint directions for each corresponding relevant joint
     """
     global m
-    if not hasattr(obj, "metadata"):
-        return None, None, None
 
-    both_sides = obj.metadata[m.BOTH_SIDES_METADATA_FIELD] if m.BOTH_SIDES_METADATA_FIELD in obj.metadata else False
+    default_both_sides = False
```
cremebrule, line 81: nit -- shouldn't we put these lines within the if clause in L85, since the default values aren't used elsewhere?

omnigibson/object_states/overlaid.py
```diff
@@ -20,13 +24,57 @@
...
+
+        if not (self.obj.prim_type == PrimType.CLOTH and other.prim_type == PrimType.RIGID):
+            raise ValueError("Overlaid state requires obj1 is cloth and obj2 is rigid.")
+
+        sampling_success = False
+        state = self._simulator.dump_state(serialized=False)
+
+        # Get the top center of the object below
+        aabb_low, aabb_hi = other.states[AABB].get_value()
+        top_center = (aabb_low + aabb_hi) / 2 + np.array([0., 0., (aabb_hi - aabb_low)[2] / 2.0])
```
cremebrule, line 47: super nit, couldn't we just use aabb_hi[2] as the z value?

omnigibson/object_states/overlaid.py
```diff
@@ -20,13 +24,57 @@
...
+        for _ in range(10):
+            # Use a random orientation in the z-axis
+            orn = T.euler2quat(np.array([0., 0., np.random.uniform(0, np.pi * 2)]))
+            self.obj.set_position_orientation(pos, orn)
+            self.obj.root_link.reset()
+            self.obj.keep_still()
+            # Let it fall for 0.2 second
+            for _ in range(int(0.2 / og.sim.get_physics_dt())):
+                og.sim.step_physics()
+                if len(self.obj.states[ContactBodies].get_value()) > 0:
```
cremebrule, line 65: long-term, will we make Cloths part of ContactParticles, or maybe ContactCloths instead of ContactBodies just to disambiguate?

omnigibson/object_states/overlaid.py
```diff
@@ -20,13 +24,57 @@
...
+                    break
+
+            if self.get_value(other):
+                sampling_success = True
+                break
+            else:
+                if omnigibson.debug_sampling:
+                    print("Overlaid checking", sampling_success)
+                    embed()
```
cremebrule, line 74: it would be nice to make this functionality an explicit util so we don't have IPython import polluting a bunch of modules

e.g. in ui_utils (or some other module):

```python
def debug_breakpoint(msg):
    logging.error(msg)
    embed()
```

omnigibson/object_states/unfolded.py
```diff
@@ -45,9 +45,9 @@ def _get_value(self):
...
 
     def _set_value(self, new_value):
-        """
-        Set the unfolded state. Currently, it's not supported yet.
-        """
-        raise NotImplementedError("_set_value of the Unfolded state has not been implemented")
+        if not new_value:
+            raise NotImplementedError("Unfolded does not support set_value(False)")
+
+        self.obj.root_link.reset()
```
cremebrule, line 51: wow!! single line implementation :D

omnigibson/object_states/folded.py
cremebrule: oops ignore this, i forgot folded / unfolded is tertiary

omnigibson/objects/stateful_object.py
```diff
@@ -471,6 +471,9 @@ def _load_state(self, state):
                 else:
                     logging.warning("Missing object state [{}] in the state dump".format(state_name))
 
+        # Clear cache after loading state
+        self.clear_states_cache()
```
cremebrule, line 475: good catch -- in this case, would you remind removing the redundant line from `_load_state` in scene_base.py?

omnigibson/prims/cloth_prim.py
```diff
@@ -168,11 +218,30 @@ def update_handles(self):
...
+        else:
+            raise ValueError(f"Cannot compute volume for mesh of type: {mesh_type}")
+
+        mesh_volume *= np.product(self.get_world_scale())
+        return mesh_volume
 
     @volume.setter
     def volume(self, volume):
-        raise NotImplementedError("Cannot set volume for ClothPrim")
+        raise NotImplementedError("Cannot set volume directly for an link!")
```
cremebrule, line 244: super nit -- "a", not "an"

omnigibson/prims/cloth_prim.py
```diff
@@ -200,14 +269,199 @@ def density(self):
...
+            np.ndarray: current average linear velocity of the particles of the cloth prim. Shape (3,).
+        """
+        return np.array(self._prim.GetAttribute("velocities").Get()).mean(axis=0)
+
+    def get_angular_velocity(self):
+        """
+        Returns:
+            np.ndarray: zero vector as a placeholder because a cloth prim doesn't have an angular velocity. Shape (3,).
+        """
+        return np.zeros(3)
```
cremebrule, line 292: we could probably have some aggregate metric of angular velocity from individual particle velocities (e.g.: [HERE](https://arxiv.org/pdf/physics/0401146.pdf)), but probably not worth the effort?

omnigibson/prims/cloth_prim.py
```diff
@@ -200,14 +269,199 @@ def density(self):
...
 
     def wake(self):
         # TODO (eric): Just a pass through for now.
         return
+
+    @property
+    def bend_stiffness(self):
+        """
+        Returns:
+            bend_stiffness (float): spring bend stiffness of the particle system
```
cremebrule, line 323: nit: for all property getters, docstring should be

```
"""
Returns:
    float: ...
"""
```

omnigibson/prims/joint_prim.py
```diff
@@ -835,9 +835,13 @@ def _load_state(self, state):
                 self.set_vel(state["target_vel"], target=True)
 
     def _serialize(self, state):
-        # We serialize by iterating over the keys and adding them to a list that's concatenated at the end
-        # This is a deterministic mapping because we assume the state is an dict
-        return np.concatenate(list(state.values())).astype(float)
+        return np.concatenate([
```
cremebrule, line 838: was this just to make it explicit?

omnigibson/objects/stateful_object.py
cremebrule: nvm, you already did!

pyproject.toml
```diff
@@ -40,7 +40,6 @@ exclude = [
 testpaths = [
     "tests",
 ]
-addopts = '--cov=omnigibson --cov-report=xml'
```
cremebrule, line 43: what is this?

tests/test_object_states.py
```diff
@@ -0,0 +1,692 @@
...
+    for _ in range(5):
+        og.sim.step()
+
+    assert not bowl.states[OnTop].get_value(breakfast_table)
+    assert not dishtowel.states[OnTop].get_value(breakfast_table)
+
+    assert bowl.states[OnTop].set_value(breakfast_table, True)
+    assert dishtowel.states[OnTop].set_value(breakfast_table, True)
+
+    with pytest.raises(NotImplementedError):
```
cremebrule, line 42: this is cool! i like how we now explicit test for out-of-scope behavior

tests/test_object_states.py
```diff
@@ -0,0 +1,692 @@
...
+    for _ in range(5):
+        og.sim.step()
+
+    assert not bowl.states[Touching].get_value(breakfast_table)
+    assert not breakfast_table.states[Touching].get_value(bowl)
+    assert not dishtowel.states[Touching].get_value(breakfast_table)
+    assert not breakfast_table.states[Touching].get_value(dishtowel)
+
+    with pytest.raises(NotImplementedError):
+        bowl.states[Touching].set_value(breakfast_table, False)
```
cremebrule, line 137: should we also check for True in this case?

cremebrule: (same goes for other tests, e.g. NextTo that don't support setting True nor False)

tests/test_object_states.py
```diff
@@ -0,0 +1,692 @@
...
+
+    assert bagel.states[Frozen].set_value(True)
+    assert dishtowel.states[Frozen].set_value(True)
+    assert bagel.states[Frozen].get_value()
+    assert dishtowel.states[Frozen].get_value()
+    assert bagel.states[Temperature].get_value() <= bagel.states[Frozen].freeze_temperature
+    assert dishtowel.states[Temperature].get_value() <= dishtowel.states[Frozen].freeze_temperature
+
+@og_test
+def test_frozen():
```
cremebrule, line 606: should be called test_heated?

cremebrule (commented):
overall LGTM, a lot of nits and a few clarification questions

omnigibson/object_states/open.py
ChengshuLi: done

omnigibson/object_states/open.py
ChengshuLi: I am not sure. I am just maintaining the original behavior. Cem might change the annotation format again, and we will do a quick revamp then.

omnigibson/object_states/open.py
ChengshuLi: it's used in two different places actually.

omnigibson/object_states/overlaid.py
ChengshuLi: done

omnigibson/object_states/overlaid.py
ChengshuLi: Good question! I think we should have a ContactCloths class probably. We can do it in the next PR if the need arises?

omnigibson/object_states/overlaid.py
ChengshuLi: done! I also got rid of some of the debug breakpoint because I don't think they are useful there and they pollute the cleanness of the main object state codebase.

omnigibson/prims/cloth_prim.py
ChengshuLi: done

omnigibson/prims/cloth_prim.py
ChengshuLi: I think it's probably not worth the effort. Also, I am leaning towards not to do it because the object base link has no orientation change (obj.get_orientation() stays constant as simulation plays). It's better to have zero angular velocity to be consistent.

omnigibson/prims/joint_prim.py
ChengshuLi: Yes. It's less bug-prone. I actually need to do this for xform prim because otherwise it causes a bug.

pyproject.toml
ChengshuLi: I am actually not sure... I need to remove this line in order for pytest to work. Unclear why.. @cgokmen Do you have any idea?

tests/test_object_states.py
ChengshuLi: I changed it to `bowl.states[Touching].set_value(breakfast_table, None)` to test for both True and False. Hopefully it is sufficient.

tests/test_object_states.py
ChengshuLi: good catch! done.

omnigibson/object_states/open.py
cremebrule: sounds good!

omnigibson/object_states/overlaid.py
cremebrule: sg!

cremebrule (approved):
LGTM!

PR #108, opened 2023-02-25
Diced state

ChengshuLi:


omnigibson/object_states/contact_particles.py
```diff
@@ -34,6 +40,7 @@ def report_hit(hit):
             link_name = None if link is None else link.prim_path.split("/")[-1]
             base, body = "/".join(hit.rigid_body.split("/")[:-1]), hit.rigid_body.split("/")[-1]
             continue_traversal = True
+            # print("hit", base, body)
```
cremebrule, line 43: remove

omnigibson/object_states/contact_particles.py
```diff
@@ -67,23 +74,38 @@ def cache_info(self, get_value_args):
...
+
+                # If there are no instancers, skip
+                if len(info.keys()) == 0:
+                    continue
+
+                arg_pos= np.vstack([instancer.particle_positions for instancer in arg.particle_instancers.values()])
+                info_pos = np.vstack([particle_positions for particle_positions in info.values()])
+
+                # If any of the particles moved, return False
+                if np.any(np.linalg.norm(arg_pos - info_pos, axis=1) >= m.POSITIONAL_VALIDATION_EPSILON):
```
cremebrule, line 108: I'm guessing this check is indeed faster than just not caching anything?

omnigibson/systems/macro_particle_system.py
```diff
@@ -61,6 +60,9 @@ def initialize(cls, simulator):
         particle_template = cls._create_particle_template()
         simulator.import_object(obj=particle_template, register=False, auto_initialize=True)
 
+        assert len(particle_template.links) == 1, "MacroParticleSystem particle template has more than one link"
```
cremebrule, line 63: can we add a comment here as to why these asserts are needed? I'm guessing it's because we want to make sure the template mesh has no ambiguity, i.e.: (a) not articulated, and (b) only a single visual mesh that can be used

omnigibson/systems/micro_particle_system.py
```diff
@@ -329,31 +309,10 @@ def particle_prototype_ids(self, prototype_ids):
...
-        """
-        assert visibilities.shape[0] == self._n_particles, \
-            f"Got mismatch in particle setting size: {visibilities.shape[0]}, vs. number of particles {self._n_particles}!"
-        # We leverage the ids + invisibleIds prim fields to infer visibility
-        # id = 1 means visible, id = 0 means invisible
-        self.set_attribute(attr="ids", val=visibilities)
+    def state_size(self):
+        # idn, particle_group, n_particles (3),
+        # and the corresponding states for each particle (N * 14)
+        return 3 + len(self._n_particles) * (3 + 3 + 4 + 3 + 1)
```
cremebrule, line 315: maybe we just put 14 here and break it down in the comment (3 pos + 3 vel + 4 quat + 3 scale + 1 id)?

omnigibson/systems/micro_particle_system.py
```diff
@@ -422,29 +381,23 @@ def _deserialize(self, state):
...
+class MicroParticleSystem(BaseSystem):
     """
     Global system for modeling "micro" level particles, e.g.: water, seeds, cloth. This system leverages
     Omniverse's native physx particle systems
     """
     # Particle system prim in the scene, should be generated at runtime
-    prim = None
-
-    @classproperty
-    def prim_path(cls):
```
cremebrule, line 434: is this no longer needed?

cremebrule: nvm

omnigibson/systems/micro_particle_system.py
```diff
@@ -497,7 +450,7 @@ def use_isosurface(cls):
                 See https://docs.omniverse.nvidia.com/app_create/prod_extensions/ext_physics.html?highlight=isosurface#isosurface
                 for more information
         """
-        raise NotImplementedError()
+        return False
```
cremebrule, line 453: why are we returning false by default for all of these properties?

omnigibson/systems/micro_particle_system.py
```diff
@@ -672,13 +564,21 @@ def reset(cls):
...
         Updates the max instancer identification number based on the current internal state
         """
-        cls.max_instancer_idn = -1 if len(cls.particle_instancers) == 0 else \
-            int(np.max([cls.particle_instancer_name_to_idn(name) for name in cls.particle_instancers.keys()]))
+        if cls.n_instancers == 0:
+            return cls.default_instancer_idn
+        else:
+            for idn in range(max(cls.instancer_idns) + 2):
+                if idn not in cls.instancer_idns:
+                    return idn
```
cremebrule, line 577: i wonder if a set operation would be faster here:

`idn = np.min(list(set(range(max(cls.instancer_idns) + 2)) - set(cls.instancer_idns)))`

omnigibson/systems/micro_particle_system.py
```diff
@@ -890,9 +765,9 @@ def generate_particle_instancer(
 
         # Automatically generate an identification number for this instancer if none is specified
         if idn is None:
-            idn = cls.max_instancer_idn + 1
-            # Also increment this counter
-            cls.max_instancer_idn += 1
+            idn = cls.next_available_instancer_idn
+
+        assert idn not in cls.instancer_idns, f"instancer idn {idn} already exists."
```
cremebrule, line 770: this assert should maybe me moved to cls.next_available_instancer_idn?

omnigibson/systems/micro_particle_system.py
```diff
@@ -999,9 +874,10 @@ def generate_particles_from_link(
             extent = obj.aabb_extent
         # We sample the range of each extent minus
         sampling_distance = 2 * cls.particle_radius if sampling_distance is None else sampling_distance
-        n_particles_per_axis = ((extent - 2 * cls.particle_radius) / sampling_distance).astype(int) + 1
-        arrs = [np.linspace(lo + cls.particle_radius, hi - cls.particle_radius, n) for lo, hi, n in
-                zip(low, high, n_particles_per_axis)]
+        n_particles_per_axis = (extent / sampling_distance).astype(int)
+        assert np.all(n_particles_per_axis), f"link {link.name} is too small to sample any particle of radius {cls.particle_radius}."
+        arrs = [np.arange(lo + cls.particle_radius, hi - cls.particle_radius + 1e-10, cls.particle_radius * 2)
```
cremebrule, line 879: why is this 1e-10 needed?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1210,19 +991,14 @@ def remove_particle_instancer(cls, name):
...
         # Remove instancer from our tracking and delete its prim
         instancer = cls.particle_instancers.pop(name)
-        instancer.remove(simulator=cls.simulator)
-        cls.simulator.stage.RemovePrim(
-            f"{get_prototype_path_from_particle_system_path(particle_system_path=cls.prim_path)}/{name}")
-
-        # Update the max particle instancer idn
-        cls.update_max_instancer_idn()
+        instancer.remove(simulator=og.sim)
+        og.sim.stage.RemovePrim(f"{cls.prim_path}/{name}")
```
cremebrule, line 995: should this be system_prim_path, not prim_path?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1375,32 +1152,130 @@ def remove_all_particle_instancers(cls):
...
+    _color = np.array([0.0, 0.0, 1.0])
+
+    @classmethod
+    def initialize(cls, simulator):
+        # Run super first
+        super().initialize(simulator=simulator)
+
+        # Create the particle material (only if we're using high-quality rendering since this takes time)
+        # Set custom rendering settings if we're using a fluid isosurface
+        if gm.ENABLE_HQ_RENDERING:
```
cremebrule, line 1221: actually, technically we could have the water material applied but _without_ the isosurface rendering. Should we allow for this case? Might make it easier to interpret what fluid you're seeing, but without the additional overhead of the isosurface rendering

omnigibson/systems/micro_particle_system.py
```diff
@@ -1491,42 +1353,173 @@ def cm_customize_particle_material(cls):
...
         # Add to any other params specified
-        kwargs["_register_system"] = cp_register_system
-        kwargs["particle_contact_offset"] = cp_particle_contact_offset
-        kwargs["particle_density"] = cp_particle_density
         kwargs["_material_mtl_name"] = cp_material_mtl_name
         kwargs["_customize_particle_material"] = cm_customize_particle_material
 
         # Create and return the class
-        return subclass_factory(name=f"{fluid_name}System", base_classes=FluidSystem, **kwargs)
+        return super().create(
```
cremebrule, line 1360: nice!! so clean :D

omnigibson/systems/micro_particle_system.py
```diff
@@ -1491,42 +1353,173 @@ def cm_customize_particle_material(cls):
...
+    def is_fluid(cls):
+        return False
+
+    @classmethod
+    def _create_particle_prototypes(cls):
+        # Load the particle template
+        particle_template = cls._create_particle_template()
+        og.sim.import_object(obj=particle_template, register=False, auto_initialize=True)
+
+        assert len(particle_template.links) == 1, "GranularSystem particle template has more than one link"
```
cremebrule, line 1388: comment here? similar to macro particle system

omnigibson/systems/micro_particle_system.py
```diff
@@ -1491,42 +1353,173 @@ def cm_customize_particle_material(cls):
...
+        return super().create(
+            name=name,
+            particle_contact_offset=particle_contact_offset,
+            particle_density=particle_density,
+            **kwargs,
+        )
 
+def customize_particle_material_factory(attr, value):
+    def func(mat):
+        setattr(mat, attr, np.array(value))
```
cremebrule, line 1369: should we generalize to also consider single scalars (e.g: floats), not just arrays?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1491,42 +1353,173 @@ def cm_customize_particle_material(cls):
...
+        model="00_0",
+        visible=False,
+        fixed_base=False,
+        visual_only=True,
+        include_default_states=False,
+        abilities={},
+    ),
+)
+
+DangoSystem = GranularSystem.create(
```
cremebrule, line 1508: yummm

omnigibson/utils/physx_utils.py
```diff
@@ -226,11 +213,21 @@ def create_physx_particleset_pointinstancer(
     instancer.GetAngularVelocitiesAttr().Set(Vt.Vec3fArray.FromNumpy(angular_velocities))
     instancer.GetScalesAttr().Set(Vt.Vec3fArray.FromNumpy(scales))
 
+    # Take a render step to "lock" the visuals of the prototypes at the graveyard position
+    # This needs to happen AFTER setting particle states
+    og.sim.render()
+
+    # Then we move the prototypes back to zero offset because otherwise all the generated particles will be offset by
+    # the graveyard position. At this point, the prototypes themselves no longer appear at the zero offset (locked at
+    # the graveyard position), which is desirable because we don't want the agent to see the prototypes themselves.
```
cremebrule, line 222: holy, what is this black magic???

omnigibson/utils/physx_utils.py
```diff
@@ -242,11 +239,12 @@ def create_physx_particleset_pointinstancer(
     instancer_prim.GetAttribute("physxParticle:particleEnabled").Set(enabled)
 
     if is_isosurface:
+        og.log.warning("Creating an instancer that uses isosurface. The rendering of these particles will have a delay of one timestep.")
```
cremebrule, line 242: can we add the prim path to this instancer to this warning?

omnigibson/utils/physx_utils.py
```diff
@@ -242,11 +239,12 @@ def create_physx_particleset_pointinstancer(
...
         # We have to update the physics for a single step here, and then pause the simulator and take an additional sim
         # step in order for the isosurface to be rendered correctly! (empirically validated, idk why)
         # TODO: Follow-up / cleanup with omni team to see if this is expected BEHAVIOR
-        og.sim.step_physics()
-        with og.sim.paused():
-            og.app.update()
+        # og.sim.step_physics()
```
cremebrule, line 246: can we just remove these lines?

cremebrule (commented):
some minor comments, overall LGTM!

omnigibson/object_states/contact_particles.py
ChengshuLi: done

omnigibson/systems/macro_particle_system.py
ChengshuLi: done

omnigibson/systems/micro_particle_system.py
ChengshuLi: done

omnigibson/object_states/contact_particles.py
ChengshuLi: I am actually not sure. This check is slightly slower than 
```
inbound_idxs = ((lower < positions) & (positions < upper)).all(axis=-1).nonzero()[0]
```

So if no particles in inbound to further test with `get_physx_scene_query_interface`, then this check will actually make things slower.

omnigibson/systems/micro_particle_system.py
ChengshuLi: Because we have been setting them as False always? This is just to remove some redundancy. As far as I understand, only the fluid system set isosurface to be True.

omnigibson/systems/micro_particle_system.py
ChengshuLi: I tested it. The speed is pretty comparable. I will just leave it be because it's slightly more interpretable

omnigibson/systems/micro_particle_system.py
ChengshuLi: hmm I don't think so. `next_available_instancer_idn` is guaranteed to not exist. This is to safeguard if someone pass in an idn that already exists when calling `generate_particle_instancer`

omnigibson/systems/micro_particle_system.py
ChengshuLi: Oh this is to safeguard the tiny probability that N particles perfectly fit the extent, then the last particle won't be included because (hi - cls.particle_radius) is not inclusive.

omnigibson/systems/micro_particle_system.py
ChengshuLi: What does the water material look like without isosurface rendering?

omnigibson/systems/micro_particle_system.py
ChengshuLi: done!

omnigibson/systems/micro_particle_system.py
ChengshuLi: I think the value should be an array because it represents an RGB color? Down in the row if we want further customization, we can make this function more complex.

omnigibson/utils/physx_utils.py
ChengshuLi: This is black magic... I can't think of another way to fix this...

omnigibson/utils/physx_utils.py
ChengshuLi: done

omnigibson/utils/physx_utils.py
ChengshuLi: done

omnigibson/systems/micro_particle_system.py
cremebrule: gotcha. makes sense!

omnigibson/systems/micro_particle_system.py
cremebrule: sg!

omnigibson/systems/micro_particle_system.py
cremebrule: oh oh i see! makes sense

omnigibson/systems/micro_particle_system.py
cremebrule: Oh I see -- could we add a quick comment here?

omnigibson/systems/micro_particle_system.py
cremebrule: right, but there's many other values that are not just RGB values, e.g.: specular, or map_influence, etc. values that are a range from 0 to 1.0

But for now I'm fine just leaving it as is and implementing more robust stuff in the future

omnigibson/object_states/contact_particles.py
ChengshuLi: Commented out for now.

omnigibson/systems/micro_particle_system.py
ChengshuLi: done

omnigibson/systems/micro_particle_system.py
ChengshuLi: done

PR #118, opened 2023-03-07
Bug fixes for open state and cloth system

ChengshuLi:
bug fixes for open state (get relevant joints in _initialize) and cloth system (wrong system path)


cremebrule (approved):
LGTM

PR #124, opened 2023-03-10
Windows compatibility & utility improvements

wensi-ai:
- Add OG compatibility on windows
- Add installation support for zsh (linux) and cmd (windows)
- remove legacy config in pyproject.toml


cgokmen:
I have added Windows compatibility and tempfile cleanup fixes to this PR as well.

ChengshuLi (approved):
LGTM. @cremebrule Would you mind taking another quick look?

omnigibson/utils/asset_utils.py
```diff
@@ -11,26 +11,20 @@
...
-
 # Create module logger
 log = create_module_logger(module_name=__name__)
 
 
 def folder_is_hidden(p):
     """
-    Removes hidden folders from a list. Works on Linux, Mac and Windows
+    Check if a folder is hidden. This corresponds to checking if its filename starts with a dot.
+    Note that while this does not actually correspond to checking for hidden files on Windows, the
```
cremebrule, line 21: nit, maybe we should rename this function so it semantically matches this docstring? e.g.: `dot_folder`

omnigibson/__init__.py
```diff
@@ -98,6 +105,8 @@ def create_app():
     global app
     from omni.isaac.kit import SimulationApp
     app = SimulationApp({"headless": gm.HEADLESS})
+    from omni.isaac.core.utils.extensions import enable_extension
+    enable_extension("omni.kit.window.viewport")
```
cremebrule, line 109: why is this needed?

cremebrule (commented):
overall LGTM! A few comments, once addressed, feel free to merge!

omnigibson/__init__.py
wensi-ai: This is isaac sim bug on windows where this extension is not automatically loaded and will crash OG

omnigibson/utils/asset_utils.py
wensi-ai: done

omnigibson/__init__.py
```diff
@@ -71,6 +74,10 @@
 example_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "examples")
 example_config_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "configs")
 
+# Create and expose a temporary directory for any use cases. It will get destroyed *after* omni
+# shutdown by the _kill_omni function.
+tempdir = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
wensi-ai: On Linux the decrypted file will immediately get deleted after load so there shouldn't be an issue there.
On windows however, Isaac Sim won't let go of those usd files so we can only delete it upon shutdown, so I guess it should be put here so that og knows what the folder path is?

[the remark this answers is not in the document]
cremebrule: hmmm I see. @ChengshuLi what do you think? will turbosquid folks get mad at us if we do this? lol

PR #126, opened 2023-03-10
lazy loading of systems; remove simulator reference from various modules

ChengshuLi:


omnigibson/objects/object_base.py
```diff
@@ -82,7 +82,7 @@ def __init__(
             name = "{}_{}".format(category, address)
 
         # Store values
-        self.uuid = int(str(id(self))[-8:]) if uuid is None else uuid
+        self.uuid = abs(hash(name)) % (10 ** 8) if uuid is None else uuid
```
cremebrule, line 85: this is cool!

omnigibson/prims/cloth_prim.py
```diff
@@ -194,11 +190,7 @@ def contact_list(self):
             list of CsRawData: raw contact info for this cloth body
         """
         contacts = []
-
-        # Avoid circular import
-        from omnigibson.systems.system_base import get_system_from_element_name
-        cloth_system = get_system_from_element_name("Cloth")
-
+        cloth_system = get_system("cloth")
```
cremebrule, line 193: is it maybe better to have the cloth object store an internal reference to the (monolithic) cloth system instead of having to query get_system() every time?

omnigibson/scenes/scene_base.py
```diff
@@ -420,18 +416,10 @@ def reset(self):
...
-        for robot in self.robots:
-            robot.reset()
-
-        # Reset the pose and joint configuration of all scene objects.
-        if self._initial_state is not None:
-            self.load_state(self._initial_state)
-            og.app.update()
+        # Reset the states of all objects (including robots), including (non-)kinematic states and internal variables.
+        assert self._initial_state is not None
+        self.load_state(self._initial_state)
```
cremebrule, line 421: nice! this is so much cleaner

omnigibson/sensors/vision_sensor.py
```diff
@@ -130,11 +130,9 @@ def __init__(
             load_config=load_config,
         )
 
-    def _load(self, simulator=None):
+    def _load(self):
         # Define a new camera prim at the current stage
-        stage = get_current_stage()
-        prim = UsdGeom.Camera.Define(stage, self._prim_path).GetPrim()
-        return prim
+        return UsdGeom.Camera.Define(get_current_stage(), self._prim_path).GetPrim()
```
cremebrule, line 135: can we add a comment here as to why we don't directly do `og.sim.stage` like the other cases?

omnigibson/simulator.py
```diff
@@ -501,6 +493,10 @@ def play(self):
             # We need to do this because for some reason omniverse exhibits strange behavior if we do certain operations
             # immediately after playing; e.g.: syncing USD poses when flatcache is enabled
             if was_stopped:
+                if self.scene is not None and self.scene.initialized:
+                    for robot in self.scene.robots:
+                        robot.update_controller_mode()
```
cremebrule, line 498: can we add a comment here why this is needed? (feel free to add a todo comment here also since I'm following up on this)

omnigibson/transition_rules.py
```diff
@@ -638,8 +637,8 @@ def transition(self, individual_objects, group_objects):
     DicingRule(),
     # Strawberry smoothie
     BlenderRule(
-        output_system=StrawberrySmoothieSystem,
-        particle_requirements={MilkSystem: 10},
+        output_system=get_system("strawberry_smoothie"),
+        particle_requirements={get_system("milk"): 10},
```
cremebrule, line 641: meta question: should these top-level APIs take in the system string names or the systems themselves? it seems a bit bulky to always pass in `get_system(...)`, but perhaps this makes it less bug-prone and also keeps the same structure as our main object state hierarchy (which also maps classes rather than strings). what do you think?

omnigibson/utils/python_utils.py
```diff
@@ -678,7 +678,6 @@ def deserialize(cls, state):
         """
         # Sanity check the idx with the expected state size
         state_dict, idx = cls._deserialize(state=state)
-        assert cls.state_size is not None, "State size must be specified by subclass!"
```
cremebrule, line 681: why is this assert needed? doesn't this core implementation raise a `NotImplementedError` by default?

omnigibson/utils/registry_utils.py
```diff
@@ -336,10 +332,8 @@ def _dump_state(self):
 
     def _load_state(self, state):
         # Iterate over all objects and load their states
+        assert {obj.name for obj in self.objects} == state.keys(), "object mismatch when loading state for registry"
```
cremebrule, line 335: does this actually work? doesn't state.keys() return a generator object, not a set?

tests/test_object_states.py
```diff
@@ -340,7 +339,7 @@ def test_temperature():
     assert dishtowel.states[Temperature].get_value() == m.object_states.temperature.DEFAULT_TEMPERATURE
 
     # Open the microwave
-    microwave.joints["j_link_0"].set_pos(np.pi / 2)
+    microwave.joints["joint_0"].set_pos(np.pi / 2)
```
cremebrule, line 342: interesting. was this using the newest assets? i was testing on my local branch on friday and don't recall running into an error...(But maybe i was on the old assets?)

tests/utils.py
```diff
@@ -53,7 +52,7 @@ def assert_test_scene():
                 get_obj_cfg("bowl", "bowl", "ajzltc"),
                 get_obj_cfg("bagel", "bagel", "zlxkry", abilities=TEMP_RELATED_ABILITIES),
                 get_obj_cfg("cookable_dishtowel", "dishtowel", "Tag_Dishtowel_Basket_Weave_Red", prim_type=PrimType.CLOTH, abilities=TEMP_RELATED_ABILITIES),
-                get_obj_cfg("microwave", "microwave", "abzvij"),
+                get_obj_cfg("microwave", "microwave", "7128"),
```
cremebrule, line 55: isn't this the outdated microwave asset?

omnigibson/envs/env_base.py
```diff
@@ -207,6 +206,10 @@ def _load_robots(self):
         """
         # Only actually load robots if no robot has been imported from the scene loading directly yet
         if len(self.scene.robots) == 0:
+            assert og.sim.is_stopped()
+            og.sim.play()
```
cremebrule, line 210: Hmmm I would actually recommend importing the robots first before playing, since when the robot is loaded in it could be overlapping at (0, 0, 0)

omnigibson/envs/env_base.py
```diff
@@ -224,10 +227,19 @@ def _load_robots(self):
...
+
+            self.scene.update_initial_state()
+            og.sim.stop()
+
     def _load_objects(self):
         """
         Load any additional custom objects into the scene
         """
+        assert og.sim.is_stopped()
+        og.sim.play()
```
cremebrule, line 241: same thing here

cremebrule: though now that i think about it, i may see why you chose to do this way -- we want the scene to be restored properly, but we can't actually load the state if we add more objects to the scene, thereby invalidating the len(state) == num_objects check...hmmm....

cremebrule: i wonder if there's a clean way to resolve this

omnigibson/object_states/covered.py
```diff
@@ -94,19 +85,18 @@ def _set_value(self, system, new_value):
...
             # Check current state and only do something if we're changing state
             if self.get_value(system) != new_value:
                 if new_value:
                     # Generate particles
-                    success = system.generate_group_particles_on_object(group=self._visual_particle_group)
-                    # If we succeeded with generating particles (new_value = True), store additional info
-                    if success:
-                        # Store how many particles there are now -- this is the "maximum" number possible
-                        self._n_initial_visual_particles[name] = system.num_group_particles(group=self._visual_particle_group)
+                    success = system.generate_group_particles_on_object(
```
cremebrule, line 95: shouldn't we first clear all the pre-existing attached particles? i thought i might be added to `generate_group_particles()`, but I didn't see anything there either

omnigibson/object_states/particle_modifier.py
```diff
@@ -274,34 +273,6 @@ def overlap_callback(hit):
                 orientation=T.euler2quat([0, -np.pi / 2, 0]),
             )
 
-            # Generate the projection visualization
```
cremebrule, line 277: Hmmmm...this is gonna cause some serious merge conflicts with the PR I also made. Was the reason to move this to ParticleApplier to avoid creating it for the ParticleRemover? I kept it as part of ParticleModifier but added a "visualize" flag so that it won't necessarily be rendered, thereby providing flexibility for all downstream classes but keeping it as a core feature that can be leveraged. what do you think?

omnigibson/object_states/particle_modifier.py
```diff
@@ -378,14 +349,13 @@ def _generate_limit_condition(self, system):
             function: Limit checker function, with signature condition(obj) --> bool, where @obj is the specific object
                 that this ParticleModifier state belongs to
         """
+        assert system in self.modified_particle_count, f"System {system.name} is not defined in the conditions."
```
cremebrule, line 352: maybe more semantically clear to do `system in self.conditions`?

omnigibson/object_states/particle_modifier.py
```diff
@@ -441,12 +411,11 @@ def check_at_limit(self, system, verify_not_over_limit=False):
         Returns:
             bool: True if the object has reached its limit with objects from @system, otherwise False
         """
+        assert system in self.modified_particle_count, f"System {system.name} is not defined in the conditions."
```
cremebrule, line 414: same here, and elsewhere

omnigibson/object_states/saturated.py
```diff
@@ -14,7 +14,7 @@ def _get_value(self, system):
         return self.obj.states[ParticleRemover].check_at_limit(system=system)
 
     def _set_value(self, system, new_value):
-        assert_valid_key(key=system, valid_keys=ParticleRemover.supported_systems, name="particle system")
+        assert_valid_key(key=system, valid_keys=ParticleRemover.supported_active_systems, name="particle system")
```
cremebrule, line 17: why does the system need to be active? couldn't it be initialized dynamically during the get() call?

omnigibson/systems/system_base.py
```diff
@@ -79,14 +36,20 @@ def __init_subclass__(cls, **kwargs):
         if cls._register_system:
             REGISTERED_SYSTEMS[cls.__name__] = cls
 
-    # Simulator reference
-    simulator = None
+        cls._uuid = abs(hash(cls.__name__)) % (10 ** 8)
```
cremebrule, line 39: can we make this a standalone utility function in python utils or something? so we don't have to duplicate our hashing code

omnigibson/systems/system_base.py
```diff
@@ -149,20 +94,21 @@ def reset(cls):
         """
         pass
 
+    @classmethod
+    def get_active_systems(cls):
+        """
+        Returns:
+            dict: Mapping from system name to system for all systems that are subclasses of this system AND active (initialized)
+        """
+        return {system.name: system for system in REGISTERED_SYSTEMS.values() if issubclass(system, cls) and system.initialized}
```
cremebrule, line 103: shouldn't we just directly query from og.sim.scene.system_registry? (and of course wrap in in the if scene exists, initialized, etc...)

omnigibson/systems/micro_particle_system.py
```diff
@@ -1528,7 +1534,7 @@ def cm_create_particle_template(cls):
 )
 
 
-class ClothSystem(MicroParticleSystem):
+class cloth(MicroParticleSystem):
```
cremebrule, line 1537: super nit, but it's a bit weird to have lower case classes. Could we maybe have simple logic that does the following:
- takes the first letter and capitalizes it. 
- takes any underscore, removes it, and then capitalizes the following letter

Just to maintain camelcasing for our classes :p

omnigibson/systems/macro_particle_system.py
```diff
@@ -67,6 +62,26 @@ def initialize(cls, simulator):
         # Class particle objet is assumed to be the first and only visual mesh belonging to the root link
         cls.set_particle_template_object(obj=list(particle_template.root_link.visual_meshes.values())[0])
 
+    @classproperty
+    def particle_idns(cls):
+        """
+        Returns:
+            int: Number of active particles in this system
```
cremebrule, line 69: update docstring

omnigibson/systems/macro_particle_system.py
```diff
@@ -67,6 +62,26 @@ def initialize(cls, simulator):
...
+        """
+        return [cls.particle_name2idn(particle_name) for particle_name in cls.particles]
+
+    @classproperty
+    def next_available_particle_idn(cls):
+        """
+        Updates the max particle identification number based on the current internal state
```
cremebrule, line 76: update docstring

omnigibson/systems/macro_particle_system.py
```diff
@@ -801,10 +793,10 @@ def _sync_particle_groups(cls, group_objects, particle_idns, particle_attached_l
             "link_names": link_names,
         }
             for obj, p_idns, link_names in
-            zip(group_objects, particle_idns, particle_attached_link_names) if obj is not None}
+            zip(group_objects, particle_idns, particle_attached_link_names)}
```
cremebrule, line 796: did we verify this works under our tests / examples?

cremebrule: especially sim_save_load demo script

omnigibson/systems/macro_particle_system.py
```diff
@@ -850,7 +842,6 @@ def create(cls, name, n_particles_per_group, create_particle_template, min_scale
 
         Args:
             name (str): Name of the visual particles
-            n_particles_per_group (int): Number of particles to generate per group of these particles
```
cremebrule, line 853: is there a reason why this was removed? are we okay assuming all macro particle systems will have the same number of particles sampled for each type?

cremebrule (commented):
Awesome work!!! At a high level, these changes are fantastic -- really like how this is unifying things end-to-end.

I had a few clarifying questions, and a couple open-ended ones. Happy to discuss over a call.

omnigibson/prims/cloth_prim.py
ChengshuLi: done

omnigibson/sensors/vision_sensor.py
ChengshuLi: done

omnigibson/simulator.py
ChengshuLi: done

omnigibson/transition_rules.py
ChengshuLi: I think they should take in the system string names, to be consistent with the category argument in `DatasetObject(category="novel_category", ...)`. This assumes we will have infinite number of systems and we won't be able to enumerate them.

omnigibson/utils/python_utils.py
ChengshuLi: Right. I am not sure why this is needed. That's why I removed it.

omnigibson/utils/registry_utils.py
ChengshuLi: You are right. But it actually works. I guess python is smart about it.

tests/test_object_states.py
ChengshuLi: this is using the old asset. Once we merge our two PRs. I will test Cem's new asset and make tests work.

tests/utils.py
ChengshuLi: Yes, but this "abzvij" microwave causes issue for me because it has too many collision meshes. In any case, I will create a follow-up PR to get the tests working with the new asset.

omnigibson/envs/env_base.py
ChengshuLi: sorry what do you mean? in both cases, the robot could be overlapping at (0, 0, 0) right? Let's discuss in person.

omnigibson/envs/env_base.py
ChengshuLi: Let's discuss in person.

omnigibson/object_states/covered.py
ChengshuLi: It's already included in `generate_group_particles_on_object`

omnigibson/object_states/particle_modifier.py
ChengshuLi: I am not sure why we need to create visualization if it's invisible. But let's discuss in person.

omnigibson/object_states/particle_modifier.py
ChengshuLi: done!

omnigibson/object_states/saturated.py
ChengshuLi: I am being a bit more conservative here. Essentially when you call `obj.states[Saturated].set_value(get_system("XYZ"))`, `get_system` itself already initializes the system.

omnigibson/systems/system_base.py
ChengshuLi: done

omnigibson/systems/system_base.py
ChengshuLi: done

omnigibson/systems/micro_particle_system.py
ChengshuLi: done!

omnigibson/systems/macro_particle_system.py
ChengshuLi: done

omnigibson/systems/macro_particle_system.py
ChengshuLi: done

omnigibson/systems/macro_particle_system.py
ChengshuLi: I removed it because I felt like it should belong to `Covered` state, rather than `MacroParticleSystem`. This is my attempt to unify with the implementation of `MicroParticleSystem` so that `generate_group_particles_on_object` take in `max_samples` as an external argument to decide on the number of particles to sample, rather than using an internal variable like `n_particles_per_group`.

omnigibson/transition_rules.py
ChengshuLi: Make output_system take in `"strawberry_smoothie"` rather than `get_system("strawberry_smoothie")`

omnigibson/envs/env_base.py
ChengshuLi: Try og.sim.pause()

omnigibson/object_states/particle_modifier.py
ChengshuLi: I will move it back the original location with a `pass` for ParticleRemover for future implementation

omnigibson/transition_rules.py
ChengshuLi: done

omnigibson/object_states/particle_modifier.py
```diff
@@ -194,18 +194,18 @@ def __init__(self, obj, method, conditions, projection_mesh_params=None):
...
+        for system_name, conds in conditions.items():
             # Make sure the system is supported
-            assert_valid_key(key=system, valid_keys=ParticleModifier.supported_systems, name="particle system")
+            assert is_visual_particle_system(system_name) or is_physical_particle_system(system_name), f"Unsupported system for ParticleModifier {system_name}"
             # Make sure conds isn't empty and is a list
             conds = [] if conds is None else list(conds)
             # Add the condition to avoid limits
-            conds.append(self._generate_limit_condition(system=system))
-            conditions[system] = conds
+            conds.append(self._generate_limit_condition(system_name))
```
cremebrule, line 207: can we update the docstring for this class so it's clear we're mapping strings to conditions now instead of classes to conditions?

omnigibson/prims/cloth_prim.py
```diff
@@ -83,11 +83,14 @@ def _post_load(self):
         if "mass" in self._load_config and self._load_config["mass"] is not None:
             self.mass = self._load_config["mass"]
 
+        # Internal reference to the monolithic cloth system
+        self._cloth_system = get_system("cloth")
```
cremebrule, line 87: this is global right? should we just make this a class property / class attribute?

omnigibson/systems/macro_particle_system.py
```diff
@@ -19,6 +19,12 @@
 log = create_module_logger(module_name=__name__)
 
 
+def is_visual_particle_system(system_name):
+    assert system_name in REGISTERED_SYSTEMS, f"System {system_name} not in REGISTERED_SYSTEMS."
+    system = REGISTERED_SYSTEMS[system_name]
+    return issubclass(system, VisualParticleSystem)
```
cremebrule, line 25: why is this extra check needed?

omnigibson/systems/micro_particle_system.py
```diff
@@ -32,6 +32,12 @@
 log = create_module_logger(module_name=__name__)
 
 
+def is_physical_particle_system(system_name):
+    assert system_name in REGISTERED_SYSTEMS, f"System {system_name} not in REGISTERED_SYSTEMS."
+    system = REGISTERED_SYSTEMS[system_name]
+    return issubclass(system, PhysicalParticleSystem)
```
cremebrule, line 38: same here

omnigibson/systems/system_base.py
```diff
@@ -1,50 +1,40 @@
...
-
 class BaseSystem(SerializableNonInstance, UniquelyNamedNonInstance):
     """
     Base class for all systems. These are non-instance objects that should be used globally for a given environment.
     This is useful for items in a scene that are non-discrete / cannot be distinguished into individual instances,
     e.g.: water, particles, etc.
     """
     def __init_subclass__(cls, **kwargs):
+        # While class names are camel case, we convert them to snake case to be consistent with object categories.
+        cls._snake_case_name = camel_case_to_snake_case(cls.__name__)
```
cremebrule, line 19: not a huge fan of this, but i guess this is necessary in order to maintain both snake case string reference and camel case class naming?

omnigibson/envs/env_base.py
ChengshuLi: og.sim.pause does not work.

omnigibson/systems/system_base.py
ChengshuLi: Explain why we want the snake case for systems.

omnigibson/systems/macro_particle_system.py
ChengshuLi: done

omnigibson/object_states/particle_modifier.py
ChengshuLi: done

omnigibson/prims/cloth_prim.py
ChengshuLi: done

ChengshuLi:
- share the same MaterialPrim
- add if check for FORCE_ROUGHNESS

omnigibson/systems/system_base.py
ChengshuLi: done

cremebrule (approved):
LGTM!

PR #125, opened 2023-03-10
Cleanup object states

cremebrule:

PR #127, opened 2023-03-15
fix/deps

mjlbach:
Add imageio


cremebrule:
this is already taken care of in #125 !

PR #128, opened 2023-03-15
Add support for remote streaming

mjlbach:
REMOTE_ENABLED=1 python test.py, click on the link


cremebrule (approved):
this is super cool! do we have a demo script for this so people can try this out of the box?

mjlbach:
Nope, it's a global variable so should just be documented and work with any method?

omnigibson/__init__.py
```diff
@@ -97,7 +98,34 @@ def print_save_usd_warning(_):
...
+        from omni.isaac.core.utils.extensions import enable_extension
+
+        # Note: Only one livestream extension can be enabled at a time
+        # Enable Native Livestream extension
+        # Default App: Streaming Client from the Omniverse Launcher
+        # enable_extension("omni.kit.livestream.native")
+
+        # Enable WebSocket Livestream extension
+        # Default URL: http://localhost:8211/streaming/client/
+        # enable_extension("omni.services.streamclient.websocket")
```
cremebrule, line 120: when would these other options be used?

mjlbach: I mostly left the code unmodified from `standalone_examples/api/omni.isaac.kit/livestream.py` to ease diff'ing in the future if nvidia changes something. From my testing, WebRTC is the best for low-latency connections, websocket is a little more robust for constrained bandwidth, and native (mirror omni on a local omni) has little use since it essentially limits you to a windows/linux computer, and in our situation we are usually only on a windows/linux computer if we are testing something.

omnigibson/__init__.py
```diff
@@ -72,6 +73,38 @@ def print_save_usd_warning(_):
...
+        from omni.isaac.core.utils.extensions import enable_extension
+
+        hostname = socket.gethostname()
+
+        # Note: Only one livestream extension can be enabled at a time
+        if gm.REMOTE_STREAMING == "native":
+            # Enable Native Livestream extension
+            # Default App: Streaming Client from the Omniverse Launcher
+            enable_extension("omni.kit.livestream.native")
+            print(f"Now streaming on {hostname} via Omniverse Streaming Client\n")
```
ChengshuLi, line 94: super nit: why "\n" in the end

ChengshuLi (approved):

omnigibson/__init__.py
cgokmen: Great question. I guess it must have been to leave space after this but I think it's OK to remove.

cgokmen:
This now is tested to work, I am merging.

mjlbach:
lol glad this got merged

PR #130, opened 2023-03-15
Apply renderer default settings & force light intensity

cgokmen:
To get the default rendering behavior to a reasonable value, we:

- Force all usd-based lights to the same, reasonable intensity (this will need to be fixed in the assets later with a principled solution)
- Change direct lighting to sampling mode (this is important for two reasons: it adds performance in many-light settings, and it allows reflections to not be buggy)
- Reduce ambient lighting such that most of the pixel value comes from actual lights
- Make lights visible (otherwise only their reflections are visible which is weird)
- For now, disable both indirectDiffuse and reflections.
  - indirectDiffuse can be enabled with a 5fps penalty to get corners etc. better lit. Maybe this should be enabled in the long-run default setting.
  - reflections can be enabled with a 10fps penalty

I also left a todo to later allow the user to configure the rendering to one of a few presets that we can provide (otherwise the renderer setting rabbit hole is rather deep).

cc @mjlbach , @ChengshuLi

### DIFF
--- a/omnigibson/macros.py
+++ b/omnigibson/macros.py
@@ -68,6 +68,9 @@ gm.USE_ENCRYPTED_ASSETS = True
 # (Demo-purpose) Whether to activate Assistive Grasping mode for Cloth (it's handled differently from RigidBody)
 gm.AG_CLOTH = False
 
+# Forced light intensity for all USD-loaded objects
+gm.FORCE_USD_LIGHT_INTENSITY = 500000
+
 
 # Create helper function for generating sub-dictionaries
 def create_module_macros(module_path):
--- a/omnigibson/macros.py
+++ b/omnigibson/macros.py
@@ -68,8 +68,11 @@ gm.USE_ENCRYPTED_ASSETS = True
 # (Demo-purpose) Whether to activate Assistive Grasping mode for Cloth (it's handled differently from RigidBody)
 gm.AG_CLOTH = False
 
-# Forced light intensity for all USD-loaded objects
-gm.FORCE_USD_LIGHT_INTENSITY = 500000
+# Forced light intensity for all DatasetObjects. None if the USD-provided intensities should be respected.
+gm.FORCE_LIGHT_INTENSITY = 500000
+
+# Forced roughness for all DatasetObjects. None if the USD-provided roughness maps should be respected.
+gm.FORCE_ROUGHNESS = 0.7
 
 
 # Create helper function for generating sub-dictionaries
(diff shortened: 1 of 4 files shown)


omnigibson/macros.py
```diff
@@ -68,6 +68,9 @@
 # (Demo-purpose) Whether to activate Assistive Grasping mode for Cloth (it's handled differently from RigidBody)
 gm.AG_CLOTH = False
 
+# Forced light intensity for all USD-loaded objects
+gm.FORCE_USD_LIGHT_INTENSITY = 500000
```
cremebrule, line 72: nit, can we rename to `DEFAULT_LIGHT_INTENSITY`

omnigibson/macros.py
```diff
@@ -68,6 +68,9 @@
 # (Demo-purpose) Whether to activate Assistive Grasping mode for Cloth (it's handled differently from RigidBody)
 gm.AG_CLOTH = False
 
+# Forced light intensity for all USD-loaded objects
```
cremebrule, line 71: seems like also None is supported? can we also mention that here in the comment?

omnigibson/utils/usd_utils.py
```diff
@@ -637,6 +637,17 @@ def add_asset_to_stage(asset_path, prim_path):
     # Make sure prim was loaded correctly
     assert prim, f"Failed to load {asset_type.upper()} object from path: {asset_path}"
 
+    # Apply any forced light intensity updates.
+    if gm.FORCE_USD_LIGHT_INTENSITY:
```
cremebrule, line 641: I don't think this should be done here in `add_asset_to_stage` -- imo it should be handled at the object level, e.g.: `entity_prim.py` or `base_object.py`. What do you think?

cremebrule (commented):
Thanks for putting together these changes!! Some comments about (mainly cosmetic) stuff

omnigibson/objects/dataset_object.py
```diff
@@ -209,6 +209,29 @@ def _initialize(self):
...
+
+                for child_child_prim in child_prim.GetChildren():
+                    recursive_light_update(child_child_prim)
+
+            recursive_light_update(self._prim)
+
+        # Apply any forced roughness updates
+        if gm.FORCE_ROUGHNESS is not None:
+            def recursive_roughness_update(child_prim):
+                if child_prim.GetPrimTypeInfo().GetTypeName() == "Shader":
```
cremebrule, line 226: oh this should be done more programmatically, like so:

```python
for mat in self.materials:
    mat.reflection_roughness_constant = gm.FORCE_ROUGHNESS
    mat.reflection_roughness_texture_influence = 0.0
```

Ideally it'd be the same for lights but i haven't had time to that yet.

PR #137, opened 2023-03-28
Attached to

ChengshuLi:


omnigibson/simulator.py
```diff
@@ -603,26 +608,56 @@ def _on_contact(self, contact_headers, contact_data):
...
+        if gm.ENABLE_OBJECT_STATES:
+            if event.type == int(SimulationEvent.JOINT_BREAK):
+                joint_path = str(PhysicsSchemaTools.decodeSdfPath(event.payload["jointPath"][0], event.payload["jointPath"][1]))
+                obj = None
+                # TODO: recursively try to find the parent object of this joint
+                tokens = joint_path.split("/")
+                for i in range(2, len(tokens) + 1):
+                    obj = self._scene.object_registry("prim_path", "/".join(tokens[:i]))
+                    if obj is not None:
+                        break
```
ChengshuLi, line 651: Given a descendant joint, is there a good way to find the corresponding object in object_registry?

cremebrule: mmmmm...not really, unless we want to register all the objects under every joint it owns also. I think your recursive way probably makes the most sense

omnigibson/__init__.py
```diff
@@ -97,7 +97,7 @@ def print_save_usd_warning(_):
 def create_app():
     global app
     from omni.isaac.kit import SimulationApp
-    app = SimulationApp({"headless": gm.HEADLESS})
+    app = SimulationApp({"headless": gm.HEADLESS, "multi_gpu": False})
```
cremebrule, line 100: Can we add a comment here why we're forcing multi GPU to be false? (Also, should we bring this up with NV? This seems like a major regression)

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+m.ATTACHMENT_LINK_PREFIX = "attachment"
+
+m.DEFAULT_POSITION_THRESHOLD = 0.05  # 5cm
+m.DEFAULT_ORIENTATION_THRESHOLD = np.deg2rad(5.0)  # 5 degrees
+m.DEFAULT_JOINT_TYPE = JointType.JOINT_FIXED
+m.DEFAULT_BREAK_FORCE = 10000  # Newton
+m.DEFAULT_BREAK_TORQUE = 10000  # Newton-Meter
+
+
+class AttachedTo(RelativeObjectState, BooleanState, ContactSubscribedStateMixin, JointBreakSubscribedStateMixin, LinkBasedStateMixin):
```
cremebrule, line 36: Wow!! A record for class inheritance lol

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+        self.initialize_link_mixin()
+
+        # Reference to the parent object (DatasetObject)
+        self.parent = None
+
+        # Reference to the female meta link of the parent object (RigidPrim)
+        self.parent_link = None
+
+        # Mapping from the female meta link names of self.obj to their children (Dict[str, Optional[DatasetObject] = None])
+        self.children = {link_name: None for link_name in self.links if link_name.split("_")[1].endswith("F")}
```
cremebrule, line 71: why does it end with F? Is that implying a female link or something?

cremebrule: will this be pre-annotated from the asset team?

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+        # Note that when this function is invoked when a joint break event happens, @self.obj is the parent of the
+        # attachment joint, not the child. We access the child of the broken joint, and call the setter with False
+        child = self.children[joint_prim_path.split("/")[-2]]
+        child.states[AttachedTo].set_value(self.obj, False)
+
+    # Attempts to attach two objects when a CONTACT_FOUND event happens
+    def on_contact(self, other, contact_headers, contact_data):
+        for contact_header in contact_headers:
+            if contact_header.type == ContactEventType.CONTACT_FOUND:
+                self.set_value(other, True)
```
cremebrule, line 87: hmmm...this could be slightly problematic though, right? if an object happens to touch two other objects simulatenously, but the ordering the contact headers is such that the first object in the queue is not attachable, but the second object is attachable, the object won't be attached to the second object right? maybe the logic should be something like:

```
if self.set_value(other, True):
    break
```

so we keep trying until we find a successful attachment

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def _find_attachment_links(self,
+                               other,
+                               bypass_alignment_checking=False,
+                               pos_thresh=m.DEFAULT_POSITION_THRESHOLD,
+                               orn_thresh=m.DEFAULT_ORIENTATION_THRESHOLD):
+        """
+        Args:
+            other (DatasetObject): parent object to find potential attachment links.
+            bypass_alignment_checking (bool): whether to bypass alignment checking when finding attachment links.
```
cremebrule, line 131: can this be a bit more descriptive?

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+            pos_thresh (float): position difference threshold to activate attachment, in meters.
+            orn_thresh (float): orientation difference threshold to activate attachment, in radians.
+
+        Returns:
+            RigidPrim or None: link belonging to @self.obj that should be aligned to that corresponding link of @other
```
cremebrule, line 136: super nit, doc style:

```
Returns:
    2-tuple:
        - RigidPrim or None: ...
        - RigidPrim or None: ...
```

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+            Dict[str, str] or None: mapping from the male meta link names of self.obj to the correspounding female meta
+            link names of @other. Returns None if @other does not have the AttachedTo state.
+        """
+        if AttachedTo not in other.states:
+            return None
+
+        if other not in self.parent_link_candidates:
+            parent_link_names = defaultdict(set)
+            for child_link_name, child_link in self.links.items():
+                child_category = child_link_name.split("_")[1]
```
cremebrule, line 169: This seems a bit brittle, but I can understand that this is probably the most efficient way to infer critical info about the attachment links.

Can we add explicit comments here, as to the expected format of the links? From what I gather it sounds like it should be something like `attachment_<CATEGORY>_<M/F>`

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+m.DEFAULT_BREAK_FORCE = 10000  # Newton
+m.DEFAULT_BREAK_TORQUE = 10000  # Newton-Meter
+
+
+class AttachedTo(RelativeObjectState, BooleanState, ContactSubscribedStateMixin, JointBreakSubscribedStateMixin, LinkBasedStateMixin):
+    """
+        Handles attachment between two rigid objects, by creating a fixed/spherical joint between self.obj (child) and
+        other (parent). At any given moment, an object can only be attached to at most one other object, i.e.
+        a parent can have multiple children, but a child can only have one parent. Note that only
+        child.states[AttachedTo].get_value(parent) will return True.
```
cremebrule, line 41: Just to clarify, parents == female, and children == male? Is this always true? We should maybe be explicit here

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+        self.parent = None
+
+        # Reference to the female meta link of the parent object (RigidPrim)
+        self.parent_link = None
+
+        # Mapping from the female meta link names of self.obj to their children (Dict[str, Optional[DatasetObject] = None])
+        self.children = {link_name: None for link_name in self.links if link_name.split("_")[1].endswith("F")}
+
+        # Cache of parent link candidates for other objects (Dict[DatasetObject, Dict[str, str]])
+        # Map from the male meta link names of self.obj to the correspounding female meta link names of other
```
cremebrule, line 74: Isn't this `map from other object to male meta link names of self.obj to the correspounding female meta link names of other`?

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+            return None, None
+
+        for child_link_name, parent_link_names in parent_candidates.items():
+            child_link = self.links[child_link_name]
+            for parent_link_name in parent_link_names:
+                parent_link = other.states[AttachedTo].links[parent_link_name]
+                if other.states[AttachedTo].children[parent_link_name] is None:
+                    pos_diff = np.linalg.norm(child_link.get_position() - parent_link.get_position())
+                    orn_diff = T.get_orientation_diff_in_radian(child_link.get_orientation(), parent_link.get_orientation())
+                    if bypass_alignment_checking or (pos_diff < pos_thresh and orn_diff < orn_thresh):
```
cremebrule, line 150: for efficiency sake, should the bypass_alignment_checking check happen before we actually compute the pos / orn diff?

omnigibson/object_states/attached_to.py
cremebrule: maybe it would be better to have the M/F separate, so it's `attachment_<CATEGORY>_<M/F>`, and so we can split the information more cleanly:

```python
child_category, child_type = child_link_name.split("_")[-2:]
if child_type == "F":
    continue
else:
    assert child_type == "M"
    parent_category_type = f"{child_category}_{child_type}"
    ...
```

omnigibson/object_states/attached_to.py
```diff
@@ -0,0 +1,293 @@
...
+
+        # Actually move the object and also keep it still for stability purposes.
+        self.obj.set_position_orientation(new_child_root_pos, new_child_root_quat)
+        self.obj.keep_still()
+
+        if joint_type == JointType.JOINT_FIXED:
+            # For FixedJoint: the parent link, the child link and the joint frame align
+            parent_local_quat = np.array([0.0, 0.0, 0.0, 1.0])
+        else:
+            # SphericalJoint: the same except that the rotation of the parent link doesn't align with the joint frame
```
cremebrule, line 229: so just to clarify, joint frame with always align with child frame irregardless of joint type right?

omnigibson/examples/object_states/attachment_demo.py
```diff
@@ -1,335 +1,112 @@
...
-    obj1.set_linear_velocity(velocity=np.array([-2.0, 0, 1.0]))
-    for i in range(100):
-        env.step(np.array([]))  # empty action array since action space is 0 (no robots in the env)
+    for _ in range(10):
+        env.step([])
 
+    shelf_baseboard = og.sim.scene.object_registry("name", "shelf_baseboard")
+    shelf_baseboard.set_position_orientation([0, -0.979, 0.26], [0, 0, 0, 1])
+    shelf_baseboard.keep_still()
+    shelf_baseboard.set_linear_velocity([-0.2, 0, 0])
```
cremebrule, line 100: what's the goal of this demo? If it's for the user to interact there should probably be explicit instructions printed into the terminal, otherwise we should probably inform the user what is happening and then execute the attach / detach programmatically

omnigibson/object_states/link_based_state_mixin.py
```diff
@@ -28,6 +29,14 @@ def link(self):
         """
         return self._default_link if self._link is None else self._link
 
+    @property
+    def links(self):
+        """
+        Returns:
+            None or RigidPrim: The link associated with this link-based state, if it exists
```
cremebrule, line 36: Update docstring -- this is a dict now, right?

omnigibson/object_states/link_based_state_mixin.py
```diff
@@ -43,12 +52,13 @@ def _default_link(self):
...
-        self._link = None
         # TODO: Extend logic to account for multiple instances of the same metalink? e.g: _0, _1, ... suffixes
         for name, link in self.obj.links.items():
             if self.metalink_prefix in name:
-                self._link = link
-                break
+                self._links[name] = link
+
+        if len(self._links) > 0:
+            self._link = list(self._links.values())[0]
```
cremebrule, line 61: hmmm...I feel like the current paradigm of having `link` and `links` is a bit confusing. I wonder if we can unify? maybe we make `link` --> `default_link` or something (or just deprecate `link` completely, which is probably the proper solution long-term, but requires another refactoring of all other link-based states). happy to discuss.

omnigibson/object_states/attached_to.py
cremebrule: also, just realized, how will this work wrt multiple instances of the same type of attachment link? We should probably sit down with @cem and confirm the most general metalink structure, and maybe even make a a general purpose API that will programmatically decompose a given metalink name in a standardized way (e.g.: `get_metalink_info()` or something)

omnigibson/objects/object_base.py
```diff
@@ -171,14 +172,6 @@ def _post_load(self):
             PhysxSchema.PhysxArticulationAPI.Apply(root_prim)
             self.self_collisions = self._load_config["self_collisions"]
 
-        # TODO: Do we need to explicitly add all links? or is adding articulation root itself sufficient?
-        # Set the collision group
-        CollisionAPI.add_to_collision_group(
```
cremebrule, line 176: Why is this removed? I know the actual implementation in `CollisionAPI` is all commented out, but that was because (at the time, which was like pre-CoRL so almost a year ago) omni was buggy and giving us slowdowns + crashes when we were adding a bunch of objects to the collision group.

omnigibson/simulator.py
```diff
@@ -373,6 +380,11 @@ def _non_physics_step(self):
         if len(self._objects_to_initialize) > 0 and self.is_playing():
             for obj in self._objects_to_initialize:
                 obj.initialize()
+                if len(obj.states.keys() & self.object_state_types_on_contact) > 0:
+                    self._objects_require_contact_callback = True
```
cremebrule, line 384: this is cool! We should probably also make sure these flags get reset every time a new stage is opened / sim is cleared. (actually, in general, maybe put a bunch of stuff in an explicit function like `reset_variables` or something)

omnigibson/simulator.py
```diff
@@ -602,27 +614,57 @@ def _on_contact(self, contact_headers, contact_data):
...
+                # If any of the objects is not initialized, skip
+                if not actor0_obj.initialized or not actor1_obj.initialized:
+                    continue
+                # If any of the objects is not stateful, skip
                 if not isinstance(actor0_obj, StatefulObject) or not isinstance(actor1_obj, StatefulObject):
                     continue
+                # If any of the objects doesn't have states that require on_contact callbacks, skip
+                if len(actor0_obj.states.keys() & self.object_state_types_on_contact) == 0 or len(actor1_obj.states.keys() & self.object_state_types_on_contact) == 0:
+                    continue
+                headers[tuple(sorted((actor0_obj, actor1_obj), key=lambda x: x.name))].append(contact_header)
```
cremebrule, line 637: oops, i guess we will resolve this merge conflict later?

omnigibson/simulator.py
cremebrule: btw, did you see how slow this call was (how much overhead it adds) if we run it in a full scene like Rs_int?

omnigibson/utils/usd_utils.py
```diff
@@ -193,58 +197,37 @@ def create_joint(prim_path, joint_type, body0=None, body1=None, enabled=True, st
...
-        # if col_group not in cls.ACTIVE_COLLISION_GROUPS:
-        #     # Raise error if we don't explicitly want to create a new group
-        #     if not create_if_not_exist:
-        #         raise ValueError(f"Collision group {col_group} not found in current registry, and create_if_not_exist"
-        #                          f"was set to False!")
-        #     # Otherwise, create the new group
-        #     col_group_name = f"/World/collisionGroup_{col_group}"
-        #     group = UsdPhysics.CollisionGroup.Define(get_current_stage(), col_group_name)
-        #     group.GetFilteredGroupsRel().AddTarget(col_group_name)  # Make sure that we can collide within our own group
-        #     cls.ACTIVE_COLLISION_GROUPS[col_group] = group
```
cremebrule, line 236: Oh I see the CollisionAPI was removed altogether lol...we should check to see if this API has been fixed from the omni side.

cremebrule: unless it's unnecessary? maybe we don't need it after all

omnigibson/utils/usd_utils.py
```diff
@@ -193,58 +197,37 @@ def create_joint(prim_path, joint_type, body0=None, body1=None, enabled=True, st
...
-    @classmethod
-    def add_to_collision_group(cls, col_group, prim_path, create_if_not_exist=False):
-        """
-        Adds the prim and all nested prims specified by @prim_path to the global collision group @col_group. If @col_group
-        does not exist, then it will either be created if @create_if_not_exist is True, otherwise will raise an Error.
+    if joint_frame_in_parent_frame_pos is not None:
+        joint_prim.GetAttribute("physics:localPos0").Set(Gf.Vec3f(*joint_frame_in_parent_frame_pos))
+    if joint_frame_in_parent_frame_quat is not None:
+        joint_frame_in_parent_frame_quat = joint_frame_in_parent_frame_quat[[3, 0, 1, 2]]
+        joint_prim.GetAttribute("physics:localRot0").Set(Gf.Quatf(*joint_frame_in_parent_frame_quat))
```
cremebrule, line 209: nit, can prob combine these two lines into one call, e.g.:

```python
joint_prim.GetAttribute("physics:localRot0").Set(Gf.Quatf(*joint_frame_in_parent_frame_quat[[3, 0, 1, 2]]))
```

omnigibson/utils/usd_utils.py
```diff
@@ -193,58 +197,37 @@ def create_joint(prim_path, joint_type, body0=None, body1=None, enabled=True, st
...
-        #     # Otherwise, create the new group
-        #     col_group_name = f"/World/collisionGroup_{col_group}"
-        #     group = UsdPhysics.CollisionGroup.Define(get_current_stage(), col_group_name)
-        #     group.GetFilteredGroupsRel().AddTarget(col_group_name)  # Make sure that we can collide within our own group
-        #     cls.ACTIVE_COLLISION_GROUPS[col_group] = group
-        #
-        # # Add this prim to the collision group
-        # cls.ACTIVE_COLLISION_GROUPS[col_group].GetCollidersCollectionAPI().GetIncludesRel().AddTarget(prim_path)
+    # Possibly (un-/)enable this joint
+    joint_prim.GetAttribute("physics:jointEnabled").Set(enabled)
```
cremebrule, line 222: previously this had to happen after the update_simulation() call. is it okay that this happens before?

tests/test_envs.py
```diff
@@ -24,14 +24,18 @@ def task_tester(task_type):
...
     # Create the environment
     env = og.Environment(configs=cfg, action_timestep=1 / 60., physics_timestep=1 / 60.)
 
     env.reset()
     for _ in range(5):
         env.step(env.robots[0].action_space.sample())
 
-    og.sim.stop()
+    # Clear the sim
+    og.sim.clear()
```
cremebrule, line 38: oh lol this was also solved in my local changes, this is also good!

tests/test_object_states.py
```diff
@@ -695,3 +696,66 @@ def test_toggled_on():
...
+
+    force_dir = np.array([0, 0, 1])
+    # A small force will not break the attachment
+    force_mag = 10
+    apply_force_at_pos(shelf_shelf.root_link, force_dir * force_mag, shelf_shelf.get_position())
+    og.sim.step()
+    assert shelf_shelf.states[AttachedTo].get_value(shelf_back_panel)
+
+    # A large force will break the attachment
+    force_mag = 1000
```
cremebrule, line 741: was the force - breakage correlation issue ever resolved?

cremebrule (commented):
Great work!!! Overall LGTM, left some clarification questions and few discussion points we can go over.

omnigibson/__init__.py
ChengshuLi: Done!

omnigibson/object_states/attached_to.py
ChengshuLi: done!

omnigibson/object_states/attached_to.py
ChengshuLi: done

omnigibson/object_states/attached_to.py
ChengshuLi: done

omnigibson/object_states/attached_to.py
ChengshuLi: done

omnigibson/object_states/attached_to.py
ChengshuLi: Correct.

omnigibson/examples/object_states/attachment_demo.py
ChengshuLi: Added an explicit instruction!

omnigibson/object_states/link_based_state_mixin.py
ChengshuLi: done

omnigibson/object_states/link_based_state_mixin.py
ChengshuLi: Discussed in person. We will address this in the next PR that properly allows object states to use more than one meta link per type.

omnigibson/object_states/attached_to.py
ChengshuLi: Discussed in person. We will address this in the next PR that properly allows object states to use more than one meta link per type.

omnigibson/objects/object_base.py
ChengshuLi: added back

omnigibson/simulator.py
ChengshuLi: will do this after you merge your PR.

omnigibson/utils/usd_utils.py
ChengshuLi: added back

omnigibson/utils/usd_utils.py
ChengshuLi: done

omnigibson/utils/usd_utils.py
ChengshuLi: yep

tests/test_object_states.py
ChengshuLi: no :(

omnigibson/simulator.py
ChengshuLi: We could potentially register these mapping in `JointBreakSubscribedStateMixin` but it will make things stateful. Since the overhead here is very small (joint break event is very rare), it should be fine for now.

omnigibson/simulator.py
ChengshuLi: done

PR #140, opened 2023-03-30
change ParticleModifier mesh orientation, fix mass/density assignment

ChengshuLi:


omnigibson/examples/object_states/particle_source_sink_demo.py
```diff
@@ -51,16 +51,16 @@ def check_toggledon(obj):
...
+                "source_height": 0.05,
                 "initial_speed": 0.0,               # Water merely falls out of the spout
             },
             "particleSink": {
                 "conditions": {
                     "water": None,  # No conditions, always sinking nearby particles
                 },
-                "sink_radius": 0.20,
-                "sink_height": 0.20,
+                "sink_radius": 0.05,
```
cremebrule, line 62: In this case it's 4x smaller -- was this just empirically tuned?

omnigibson/object_states/particle_modifier.py
```diff
@@ -724,22 +716,20 @@ def _compute_particle_spawn_information(self, system):
...
-        # Flip the x and z values so that the positions are in the LINK frame, not the MESH frame
-        # This is super ugly, but necessary because omni requires the native visualization XForm to have its
-        # local x-axis pointing in the direction where particles will be spawned!
-        points = points[:, [2, 1, 0]]
         pos, quat = self.link.get_position_orientation()
         points_in_world_frame = get_particle_positions_from_frame(
             pos=pos,
             quat=quat,
-            scale=np.ones(3),
+            scale=self.link.scale,
```
cremebrule, line 732: we should assert self.obj.scale == self.link.scale

cremebrule: and then prob just use self.obj.scale since that's generally what we use elsewhere

omnigibson/object_states/particle_modifier.py
```diff
@@ -888,7 +878,7 @@ def _apply_particles_in_projection_volume(self, system):
         points = get_particle_positions_from_frame(
             pos=pos,
             quat=quat,
-            scale=np.ones(3),
+            scale=self.link.scale,
```
cremebrule, line 881: same here

omnigibson/prims/rigid_prim.py
```diff
@@ -101,7 +101,7 @@ def _post_load(self):
         self.update_meshes()
 
         # Possibly set the mass / density
-        if len(self._collision_meshes) == 0:
+        if not self.has_collision_meshes:
             # We have no collision meshes, this is a meta (virtual) link, so set a negligible mass
             self.mass = 1e-6
```
cremebrule, line 106: maybe we should also explicitly set density = 0 here?

cremebrule (approved):
LGTM! A couple small comments

omnigibson/examples/object_states/particle_source_sink_demo.py
ChengshuLi: Yeah I think it's more realistic that way. Otherwise the drainage hole is going to be really large.

omnigibson/prims/rigid_prim.py
ChengshuLi: done

omnigibson/object_states/particle_modifier.py
ChengshuLi: added to link based state mixin

PR #153, opened 2023-04-07
fix installation bug

wensi-ai:


omnigibson/__init__.py
```diff
@@ -182,25 +182,26 @@ def start():
...
-        shutil.rmtree(tempdir)
-    except PermissionError:
-        log.info("Permission error when removing temp files. Ignoring")
-    from omnigibson.utils.ui_utils import suppress_omni_log
-    log.info(f"{'-' * 10} Shutting Down {logo_small()} {'-' * 10}")
-
-    # Suppress carb warning here that we have no control over -- it's expected
-    with suppress_omni_log(channels=["carb"]):
-        app.close()
+    if not (os.getenv("OMNIGIBSON_NO_OMNIVERSE", 'False').lower() in {'true', '1', 't'}):
```
cgokmen, line 190: Nit: put this in a single place instead of checking it in two places. `NO_OMNIVERSE = os.getenv...` and use the NO_OMNIVERSE variable later (in fact maybe put that in gm)

cgokmen (approved):

omnigibson/__init__.py
wensi-ai: done

wensi-ai: @ChengshuLi @cremebrule Do you think I should put it in gm?

cremebrule (approved):
LGTM!

omnigibson/__init__.py
cremebrule: No, i think this is fine. It's not really a macro you'd set imho!

PR #155, opened 2023-04-07
fix api doc formatting

cremebrule:


wensi-ai (approved):
Tested on my end, works well!

PR #158, opened 2023-04-08
Final release fixes

cremebrule:


omnigibson/tasks/behavior_task.py
```diff
@@ -952,14 +953,17 @@ def _get_obs(self, env):
         low_dim_obs["robot_ori_cos"] = np.cos(env.robots[0].get_rpy())
         low_dim_obs["robot_ori_sin"] = np.sin(env.robots[0].get_rpy())
 
+        # Batch rpy calculations for much better efficiency
+        objs_rpy = T.quat2euler(np.array([v.states[Pose].get_value()[1] for v in self.object_scope.values()]))
```
ChengshuLi, line 957: nit: why is this faster? I am sure you have validated this, but just curious: this is still a for loop, right?

ChengshuLi (approved):

omnigibson/tasks/behavior_task.py
cremebrule: the bottleneck is the T.quat2euler() call -- batching it as a single (B, 4) array vs (4,) x B calls is a lot faster

PR #163, opened 2023-04-11
Add the distance to image plane modality of the vision sensor

zhuoranli23:
Add the "dist_to_image_plane" modality of the vision sensor. This new modality provides the exact distance from the scene to the image plane of the camera (e.g., depth data).


ChengshuLi:
Already implemented with ReplicatorAPI.

PR #165, opened 2023-04-18
Refactor transition rules

cremebrule:


omnigibson/transition_rules.py
```diff
@@ -369,213 +517,192 @@ def transition(self, individual_objects, group_objects):
...
+        """
+        # Default is empty dictionary
+        return dict()
 
-        # Call super method
-        super().__init__(individual_filters=individual_filters, group_filters=group_filters)
+    @classproperty
+    def individual_filters(cls):
+        # We want to filter for any object that is included in @obj_requirements and also separately for blender
+        return {"blender": CategoryFilter("blender")}
```
ChengshuLi, line 696: I wonder whether this should be configurable as well, e.g. mixer, other types of containers, etc.

omnigibson/transition_rules.py
```diff
@@ -369,213 +517,192 @@ def transition(self, individual_objects, group_objects):
...
 
-    def condition(self, individual_objects, group_objects):
+    @classmethod
+    def condition(cls, individual_objects, group_objects):
         # TODO: Check blender if both toggled on and lid is closed!
```
ChengshuLi, line 700: I think we should check if the blender is toggledOn here. Otherwise I am not sure what is the trigger event? 

I am okay with leaving the "lid is closed" requirement out because some kitchen appliance might not have a lid.

omnigibson/transition_rules.py
```diff
@@ -615,39 +743,35 @@ def transition(self, individual_objects, group_objects):
...
             system = get_system(system_name)
             # No need to check for whether particle instancers exist because they must due to @self.condition passing!
             for inst in system.particle_instancers.values():
-                indices = self._check_in_volume[blender.name](inst.particle_positions).nonzero()[0]
+                indices = cls._CHECK_IN_VOLUME[blender.name](inst.particle_positions).nonzero()[0]
                 inst.remove_particles(idxs=indices)
 
         # Spawn in blended physical particles!
-        blender.states[Filled].set_value(get_system(self.output_system), True)
+        blender.states[Filled].set_value(get_system(cls.output_system), True)
```
ChengshuLi, line 754: Sanjana, Sharon and I had a brief discussion about this. Four modifications are probably needed in the BlenderRule.
1. BDDL will not provide the quantity of substance (e.g. milk) in the blender rule. So we should essentially check for `ContainsAny` for each requires substance.
2. For rigid bodies (e.g. ice cubes, strawberries), we should check if the right proportion is in the blender when it's toggled on. E.g. if the recipe requires 1 strawberry + 2 ice cubes, then 2 strawberries + 4 ice cubes should also work.
3. If the proportion is right, we should generate the output substance that has the same rough volume of the input rigid bodies and substance. E.g. 2 strawberries + 4 ice cubes should produce more smoothie than 1 strawberry + 2 ice cubes. Right now, we are just setting Filled = True for the container, which can be a little unrealistic. 
4. If the proportion is wrong (including missing ingredient), a special substance ("garbage") should be outputted. 
Sorry this is a bit complicated, happy to discuss in person!

omnigibson/transition_rules.py
```diff
@@ -615,39 +743,35 @@ def transition(self, individual_objects, group_objects):
...
+    @classproperty
+    def _do_not_register_classes(cls):
+        # Don't register this class since it's an abstract template
+        classes = super()._do_not_register_classes
+        classes.add("BlenderRuleTemplate")
+        return classes
+
+
+# Create strawberry smoothie blender rule
+StrawberrySmoothieRule = BlenderRuleTemplate.create(
```
ChengshuLi, line 772: Aside from the `BlenderRuleTemplate`, we also need to add a `magic stirring rod template` maybe? Or maybe these two can be combined because much logic is shared? Happy to discuss more to flesh out.

omnigibson/object_states/contains.py
```diff
@@ -0,0 +1,89 @@
...
+        self.check_in_volume, calculate_volume = \
+            generate_points_in_volume_checker_function(obj=self.obj, volume_link=self.link, mesh_name_prefixes="container")
+
+        # Calculate volume
+        self._volume = calculate_volume()
+
+    def cache_info(self, get_value_args):
+        # Call super first
+        info = super().cache_info(get_value_args=get_value_args)
+        info.update(self._compute_info)
```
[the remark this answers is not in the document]
cremebrule: this is caching the intermediate computated values so it can be leveraged by other object states without re-computing.

concretely in this case, `Contains` is storing its computed `particle_positions` and `particles_in_volume` arrays so that downstream transition rule `CookingPhysicalParticleRule` can grab these necessary values without recomputing them.

omnigibson/object_states/contains.py
```diff
@@ -0,0 +1,89 @@
...
+from omnigibson.utils.geometry_utils import generate_points_in_volume_checker_function
+from omnigibson.utils.python_utils import classproperty
+
+# Create settings for this module
+m = create_module_macros(module_path=__file__)
+
+m.CONTAINER_LINK_PREFIX = "container"
+
+
+class Contains(RelativeObjectState, LinkBasedStateMixin):
```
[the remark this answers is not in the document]
cremebrule: This is a great idea! done

omnigibson/simulator.py
```diff
@@ -381,43 +375,57 @@ def _reset_variables(self):
...
+
+                self._objects_to_initialize = self._objects_to_initialize[n_objects_to_initialize:]
+
+                # Also refresh the transition rules that are currently active
+                TransitionRuleAPI.refresh_active_rules(objects=self.scene.objects)
+
+            # Propagate states if the feature is enabled
+            if gm.ENABLE_OBJECT_STATES:
+                # Step the object states in global topological order (if the scene exists).
+                if self.scene is not None:
```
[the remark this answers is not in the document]
cremebrule: done!

omnigibson/transition_rules.py
```diff
@@ -20,9 +27,135 @@
...
+
+        # Clear all active rules
+        cls.ACTIVE_RULES = dict()
+
+        # Iterate over all registered rules, and parse their filtering requirements
+        for rule in REGISTERED_RULES.values():
+            individual_objects, group_objects = rule.get_object_candidates(objects=objects)
+            n_individual_filters_satisfied, n_group_filters_satisfied = len(individual_objects), len(group_objects)
+            # Skip if the rule requirements are not met
+            if (rule.requires_individual_filters and n_individual_filters_satisfied != len(rule.individual_filters)) \
```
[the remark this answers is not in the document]
cremebrule: Yeah! This is very common -- basically, this is if some core static requirement (e.g.: objects that are "sliceable" or of category "table knife") is not met. In this case, the number of filters will be greater than the number of filters satisfied

Note that this process assumes the filter outputs are consistent across all timesteps, assuming no objects are added / deleted.

omnigibson/transition_rules.py
```diff
@@ -20,9 +27,135 @@
...
+            new_obj = added_obj_attr.obj
+            og.sim.import_object(new_obj)
+            # By default, added_obj_attr is populated with all Nones -- so these will all be pass-through operations
+            # unless pos / orn (or, conversely, bb_pos / bb_orn) is specified
+            if added_obj_attr.pos is not None or added_obj_attr.orn is not None:
+                new_obj.set_position_orientation(position=added_obj_attr.pos, orientation=added_obj_attr.orn)
+            elif isinstance(new_obj, DatasetObject) and \
+                    (added_obj_attr.bb_pos is not None or added_obj_attr.bb_orn is not None):
+                new_obj.set_bbox_center_position_orientation(position=added_obj_attr.bb_pos,
+                                                             orientation=added_obj_attr.bb_orn)
```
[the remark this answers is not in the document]
cremebrule: done!

omnigibson/transition_rules.py
```diff
@@ -301,17 +462,14 @@ def transition(self, individual_objects, group_objects):
...
-
-            part_obj_name = f"{sliced_obj.name}_part_{part_idx}"
+            part_obj_name = f"{sliced_obj.name}_part_{i}"
             part_obj = DatasetObject(
                 prim_path=f"/World/{part_obj_name}",
                 name=part_obj_name,
                 category=part["category"],
                 model=part["model"],
                 bounding_box=part["bb_size"] * scale,   # equiv. to scale=(part["bb_size"] / self.native_bbox) * (scale)
+                bddl_object_scope=None if sliced_obj.bddl_object_scope is None else f"half_{sliced_obj_scope_prefix}_{2 * sliced_obj_id - i}",
```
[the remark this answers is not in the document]
cremebrule: Oh....we should sync with @sanjanasrivastava then. From her BDDL rules it seemed like the expected scope name output should be "half_xxx"

omnigibson/transition_rules.py
```diff
@@ -369,213 +517,192 @@ def transition(self, individual_objects, group_objects):
...
+    def transition(cls, individual_objects, group_objects):
         t_results = TransitionResults()
+        fillable_obj = individual_objects["fillable"]
+        contains_state = fillable_obj.states[Contains]
+
+        # Iterate over all active physical particle systems, and for any non-cooked particles inside,
+        # convert into cooked particles
+        for name, system in PhysicalParticleSystem.get_active_systems().items():
+            # Skip any systems that are already cooked or do not contain any particles from this system
+            if "cooked" in name or not fillable_obj.states[ContainsAny].get_value():
```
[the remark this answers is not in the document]
cremebrule: duh! great catch. fixed

omnigibson/transition_rules.py
```diff
@@ -369,213 +517,192 @@ def transition(self, individual_objects, group_objects):
...
-
-    This rule is used as a catch-all rule for containers to turn objects inside
-    the container that did not match any other legitimate rules all into a
-    single garbage object.
+    Transition rule to apply to meltable objects to simulate melting
     """
+    @classproperty
+    def individual_filters(cls):
+        # We want to find all meltable object / heatsource combos
+        # TODO: This does not handle flammable objects currently
```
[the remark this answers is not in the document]
cremebrule: originally, i thought that it removes the assumption that all filters are static (i.e.: unchanging over time given the same set of objects).

Though actually, now that I think of it, I think this is actually fine, since the number of objects that have the OnFire state is static, and we'll just check for OnFire == True. Updated!

omnigibson/transition_rules.py
```diff
@@ -369,213 +517,192 @@ def transition(self, individual_objects, group_objects):
...
+        return {ability: AbilityFilter(ability) for ability in ("meltable", "heatSource")}
 
-    def __init__(self, garbage_obj_attrs, container_filter):
-        """Ctor for ContainerGarbageRule.
+    @classmethod
+    def condition(cls, individual_objects, group_objects):
+        # Return True if the heat source is touching the meltable object and the heat source is either toggled on or
+        # non-toggleable
+        meltable_obj, heatsource_obj = individual_objects["meltable"], individual_objects["heatSource"]
+        if ToggledOn in heatsource_obj.states and not heatsource_obj.states[ToggledOn].get_value():
```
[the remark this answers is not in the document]
cremebrule: done!

omnigibson/transition_rules.py
```diff
@@ -369,213 +517,192 @@ def transition(self, individual_objects, group_objects):
...
-        self._cached_contained_objs = None
-        self._counter = 0
-
-    def condition(self, container_obj):
-        if (
-            ToggledOn in container_obj.states and
-            not container_obj.states[ToggledOn].get_value()
-        ):
+        if not heatsource_obj.states[Touching].get_value(meltable_obj):
             return False
```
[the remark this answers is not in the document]
cremebrule: Hmmm...I think I agree. I was just converting the melting rule I saw from sanjana's spreadsheet. 

In this case, we don't even need any melter objects right? It's simply a function of the meltable's temperature?

omnigibson/transition_rules.py
cremebrule: Agreed! This is a good idea

cremebrule: btw, I want to discuss with you -- certain abilities now seem to map to no object states, e.g.: `meltable`, `slicer`, `sliceable`, `blender`, etc. This is expected, right?

omnigibson/transition_rules.py
cremebrule: Good point! Added.

omnigibson/transition_rules.py
cremebrule: these are good points. let's discuss in person once you get back

cremebrule: a related implementation detail i think we may need to consider is how to pass relevant information from `condition` into `transition`, since `condition` may do some additional object filtering, e.g.: parsing which objects are actually inside the blender or not

omnigibson/transition_rules.py
cremebrule: this is a good point! i think this is analogous to `ToggledOn` requirement, right? there might be a way to unify. let me think about this

omnigibson/object_states/contains.py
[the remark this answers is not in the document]
ChengshuLi: maybe we can return these in the `get_value()` function so that it re-uses the native object state caching system.

omnigibson/transition_rules.py
ChengshuLi: Yes, the mapping code is incomplete.

omnigibson/transition_rules.py
ChengshuLi: Sort of related: we need to check it's ToggledOn is True in the current timestep AND False in the previous timestep. 
For magic stirring rod rule, it's Touching.

omnigibson/object_states/contains.py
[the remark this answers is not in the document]
cremebrule: done!

omnigibson/simulator.py
```diff
@@ -381,43 +375,57 @@ def _reset_variables(self):
...
+                    for state_type in self.object_state_types_requiring_update:
+                        for obj in self.scene.get_objects_with_state(state_type):
+                            # Only update objects that have been initialized so far
+                            if obj.initialized:
+                                obj.states[state_type].update()
+
+                for obj in self.scene.objects:
+                    # Only update visuals for objects that have been initialized so far
+                    if isinstance(obj, StatefulObject) and obj.initialized:
```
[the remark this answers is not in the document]
cremebrule: handle in future pr

omnigibson/transition_rules.py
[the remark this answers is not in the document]
cremebrule: try for now, if breaks, update in future

omnigibson/transition_rules.py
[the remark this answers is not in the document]
cremebrule: done

omnigibson/transition_rules.py
cremebrule: done

omnigibson/simulator.py
```diff
@@ -43,7 +42,6 @@
 
 m.DEFAULT_VIEWER_CAMERA_POS = (-0.201028, -2.72566 ,  1.0654)
 m.DEFAULT_VIEWER_CAMERA_QUAT = (0.68196617, -0.00155408, -0.00166678,  0.73138017)
-m.OBJECT_GRAVEYARD_POS = (100.0, 100.0, 100.0)
```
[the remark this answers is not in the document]
cremebrule: answer: no explicit crash, but we get concerning physx "nphase null elements" error

omnigibson/transition_rules.py
cremebrule: done

omnigibson/object_states/factory.py
```diff
@@ -55,13 +16,15 @@
     "freezable": [Frozen],
     "heatable": [Heated],
     "heatSource": [HeatSourceOrSink],
+    "meltable": [],
+    "mixing_wand": [],
```
ChengshuLi, line 20: nit: maybe we should find a more formal name for this?

omnigibson/object_states/touching.py
```diff
@@ -21,4 +23,8 @@ def _get_value(self, other):
         elif other.prim_type == PrimType.CLOTH:
             return self._check_contact(self.obj, other)
         else:
-            return self._check_contact(other, self.obj) and self._check_contact(self.obj, other)
+            # Use optimized check for rigid bodies
+            return RigidContactAPI.in_contact(
+                prim_paths_a=[link.prim_path for link in self.obj.links.values()],
+                prim_paths_b=[link.prim_path for link in other.links.values()],
+            )
```
ChengshuLi, line 30: Can we discuss this in person? I guess I am not sure when we should use ContactBodies vs RigidContactAPI.

omnigibson/simulator.py
```diff
@@ -376,149 +367,78 @@ def remove_object(self, obj):
...
+                    if len(obj.states.keys() & self.object_state_types_on_joint_break) > 0:
+                        self._objects_require_joint_break_callback = True
+
+                self._objects_to_initialize = self._objects_to_initialize[n_objects_to_initialize:]
+
+                # Re-initialize the physics view because the number of objects has changed
+                RigidContactAPI.initialize_view()
+
+                # Also refresh the transition rules that are currently active
+                TransitionRuleAPI.refresh_all_rules(objects=self.scene.objects)
```
ChengshuLi, line 415: nit: should we package these two lines into a helper function?

omnigibson/simulator.py
```diff
@@ -558,12 +478,12 @@ def play(self):
...
 
-            # Additionally run non physics things if we have a valid scene
-            if self._scene is not None:
-                self._omni_update_step()
-                self._non_physics_step()
-                if gm.ENABLE_TRANSITION_RULES:
-                    self._transition_rule_step()
+                # Initialize physics view
+                self._physics_sim_view = omni.physics.tensors.create_simulation_view(self.backend)
+                self._physics_sim_view.set_subspace_roots("/")
```
ChengshuLi, line 483: nit: should `self._physics_sim_view = None` be in the `__init__` function?

omnigibson/simulator.py
```diff
@@ -589,32 +509,26 @@ def n_physics_timesteps_per_render(self):
...
-            force_playing (bool): If True, will force physics to propagate (i.e.: set simulation, if paused / stopped,
-                to "play" mode)
         """
-        # Possibly force playing
-        if force_playing and not self.is_playing():
-            self.play()
+        # If we have imported any objects within the last timestep, we render the app once, since otherwise calling
+        # step() may not step physics
+        if len(self._objects_to_initialize) > 0:
+            self.render()
```
ChengshuLi, line 522: Maybe we should add a TODO here to check with Nvidia.

omnigibson/utils/usd_utils.py
```diff
@@ -228,6 +228,139 @@ def create_joint(prim_path, joint_type, body0=None, body1=None, enabled=True,
...
+        in_contact = None if cls._CONTACT_CACHE is None else cls._CONTACT_CACHE.get(key, None)
+        if in_contact is not None:
+            return in_contact
+
+        # In contact if any of the matrix values representing the interaction between the two groups is non-zero
+        in_contact = np.any(cls.get_impulses(prim_paths_a=prim_paths_a, prim_paths_b=prim_paths_b))
+
+        cls._CONTACT_CACHE[key] = in_contact
+
+        return in_contact
```
ChengshuLi, line 353: nit, a cleaner way to write this could be:
```
key = (tuple(prim_paths_a), tuple(prim_paths_b))
if key not in cls._CONTACT_CACHE:
    cls._CONTACT_CACHE[key]  = np.any(cls.get_impulses(prim_paths_a=prim_paths_a, prim_paths_b=prim_paths_b))
return cls._CONTACT_CACHE[key]
```

omnigibson/utils/usd_utils.py
```diff
@@ -228,6 +228,139 @@ def create_joint(prim_path, joint_type, body0=None, body1=None, enabled=True,
...
+                any of the rigid body prim path(s) defined by @prim_paths_a
+
+        Returns:
+            n-array: (N, M, 3) impulse array defining current impulses between N bodies from @prim_paths_a and M bodies
+                from @prim_paths_b
+        """
+        # Compute subset of matrix and return
+        idxs_a = [cls._PATH_TO_IDX[path] for path in prim_paths_a]
+        idxs_b = [cls._PATH_TO_IDX[path] for path in prim_paths_b]
+        return cls.get_all_impulses()[idxs_a][:, idxs_b]
```
ChengshuLi, line 326: nit: would this work? `cls.get_all_impulses()[idxs_a, idxs_b]`

Also, why is the shape [N, M, 3]? What is the `3`?

omnigibson/utils/usd_utils.py
```diff
@@ -228,6 +228,139 @@ def create_joint(prim_path, joint_type, body0=None, body1=None, enabled=True,
...
+        Grab all impulses at the current timestep
+
+        Returns:
+            n-array: (N, N, 3) impulse array defining current impulses between all N contact-sensor enabled rigid bodies
+                in the simulator
+        """
+        # Generate the contact matrix if it doesn't already exist
+        if cls._CONTACT_MATRIX is None:
+            cls._CONTACT_MATRIX = cls._CONTACT_VIEW.get_contact_force_matrix(dt=1.0)
+            cls._CONTACT_CACHE = dict()
```
ChengshuLi, line 303: nit: why do we have `cls._CONTACT_CACHE = dict()` here? Is it not sufficient to just have `cls._CONTACT_CACHE = dict()` in the `clear` function?

omnigibson/transition_rules.py
```diff
@@ -20,37 +40,205 @@
...
+        if (rule.requires_individual_filters and n_individual_filters_satisfied != len(rule.individual_filters)) \
+                or (rule.requires_group_filters and n_group_filters_satisfied != len(rule.group_filters)):
+            return
 
-        return super(BaseFilter, cls).__new__(cls)
+        return dict(individual=individual_candidates, group=group_candidates)
 
     @classmethod
-    def update(cls):
+    def prune_active_rules(cls, objects):
```
ChengshuLi, line 82: hmm, so what is the relationship between `prune_active_rules`, `refresh_rules` and `refresh_all_rules`? 

Is it possible for `prune_active_rules` to increase the set of active rules? Or will it strictly prune the set (i.e. reduce)?

omnigibson/transition_rules.py
```diff
@@ -20,37 +40,205 @@
...
+                rule.initialize()
+                conditions = []
+                combined_candidates = dict(**candidates["individual"], **candidates["group"])
+                for condition in rule.conditions:
+                    condition.initialize(combined_candidates)
+                    conditions.append(condition)
+                cls.ACTIVE_RULES[rule] = {
+                    "conditions": conditions,
+                    "candidates": candidates,
+                }
```
ChengshuLi, line 139: hmm is this logic correct? What is `rule` is already in the `ACTIVE_RULES`? 


```
for rule in rules:
    # Check if rule is still valid, if so, update its entry
    candidates = cls.get_rule_candidates(rule=rule, objects=objects)

    if candidates is None and rule in cls.ACTIVE_RULES:
        cls.ACTIVE_RULES.pop(rule)
    elif candidates is not None and rule not in cls.ACTIVE_RULES:
        rule.initialize()
        conditions = []
        combined_candidates = dict(**candidates["individual"], **candidates["group"])
        for condition in rule.conditions:
            condition.initialize(combined_candidates)
            conditions.append(condition)
        cls.ACTIVE_RULES[rule] = {
            "conditions": conditions,
            "candidates": candidates,
        }
```

Also, I am worried that a rule might be initialized more than once?

omnigibson/transition_rules.py
```diff
@@ -20,37 +40,205 @@
...
+            should_transition = True
+            for condition in conditions:
+                if not condition(object_candidates=combined_candidates):
+                    # Condition was not met, so break early
+                    should_transition = False
+                    break
+            if should_transition:
+                # Compile all valid transitions, and take the transition
+                pruned_individual_candidates = {k: combined_candidates[k] for k in candidates["individual"].keys()}
+                group_objects = {k: combined_candidates[k] for k in candidates["group"].keys()}
```
ChengshuLi, line 171: Sorry it's a bit hard to understand what these lines are doing.

omnigibson/transition_rules.py
```diff
@@ -103,115 +297,332 @@ def __call__(self, obj):
...
+        impulses = RigidContactAPI.get_all_impulses()
 
-            group_filters (None or dict): Group object filters that this filter cares about. For each name, filter
-                key-value pair, the global transition rule step will produce a single dictionary of valid filtered
-                objects.
-                For example, if the group filters are:
+        # Keep any object that has non-zero impulses between itself and any of the @filter_2_name's objects
+        objs = []
+        for obj in object_candidates[self._filter_1_name]:
+            if np.any(impulses[self._filter_1_idxs[obj]][:, self._filter_2_idxs]):
```
ChengshuLi, line 369: nit: `impulses[self._filter_1_idxs[obj], self._filter_2_idxs]`

omnigibson/transition_rules.py
```diff
@@ -103,115 +297,332 @@ def __call__(self, obj):
...
+            val (any): The value @state should be in order for this condition to be satisfied
+            op (function): Binary operator to apply between @state's getter and @val. Default is operator.eq,
+                which does state.get_value() == val.
+                Expected signature:
+                    def op(state_getter, val) --> bool
+        """
+        self._filter_name = filter_name
+        self._state = state
+        self._val = val
+        self._op = op
```
ChengshuLi, line 404: wow!! this is OP!

omnigibson/transition_rules.py
```diff
@@ -103,115 +297,332 @@ def __call__(self, obj):
...
+
+    def __call__(self, object_candidates):
+        # Iterate over all current candidates -- if there's a mismatch in last valid candidates and current, then
+        # we store it, otherwise, we don't
+        objs = [obj for obj in object_candidates[self._filter_name] if obj not in self._last_valid_candidates]
+        object_candidates[self._filter_name] = objs
+        self._last_valid_candidates = set(objs)
 
+        # Valid if any object conditions have changed
+        return len(objs) > 0
```
ChengshuLi, line 444: Sorry I am not sure I understand how this works. maybe we can chat in person?

omnigibson/transition_rules.py
```diff
@@ -296,358 +710,453 @@ def transition(self, individual_objects, group_objects):
...
-    This rule is used as a catch-all rule for containers to turn objects inside
-    the container that did not match any other legitimate rules all into a
-    single garbage object.
+    Transition rule to apply to mixing objects together
     """
+    # Maps recipe name to recipe information
+    # NOTE: By defining this class variable directly here, all subclasses SHARE this same dictionary!
+    # This is intentional because there may be multiple valid mixing methods,
+    # and all of them should share the same global recipe pool
+    _RECIPES = dict()
```
ChengshuLi, line 857: So are `BlenderRule` and `MixingWandRule` sharing the same `_RECIPES`? Would we lose some efficiency compared to keeping two separate lists of recipes?

omnigibson/transition_rules.py
```diff
@@ -296,358 +710,453 @@ def transition(self, individual_objects, group_objects):
...
+        Returns:
+            bool: True if all the input object quantities are contained
+        """
+        for obj_category, obj_quantity in recipe["input_objects"].items():
+            if np.sum(in_volume[cls._CATEGORY_IDXS[obj_category]]) < obj_quantity:
                 return False
+        return True
+
+    @classmethod
+    def _validate_nonrecipe_objects_not_contained(cls, recipe, in_volume):
```
ChengshuLi, line 941: Based on our yesterday's discussion, maybe we can relax this constraint for `MixingWandRule`?

omnigibson/transition_rules.py
```diff
@@ -296,358 +710,453 @@ def transition(self, individual_objects, group_objects):
...
+        t_results = TransitionResults()
 
+        # Compute all relevant object AABB positions
+        obj_positions = np.array([obj.aabb_center for obj in cls._OBJECTS])
+
+        # Iterate over all fillable objects, to execute recipes for each one
+        for container in individual_objects["container"]:
+            # Compute in volume for all relevant object positions
+            contained_particles_state = container.states[ContainedParticles]
+            in_volume = contained_particles_state.check_in_volume(obj_positions)
```
ChengshuLi, line 1037: Interesting, I guess this is far more efficient than checking `obj.states[Inside].get_value(container)` for each object.

omnigibson/transition_rules.py
```diff
@@ -296,358 +710,453 @@ def transition(self, individual_objects, group_objects):
...
+            for system in PhysicalParticleSystem.get_active_systems():
+                if container.states[Contains].get_value(system):
+                    container.states[Filled].set_value(system, False)
+
+            # Remove all objects inside the container
+            objs_to_remove = cls._OBJECTS[np.where(in_volume)[0]].tolist()
+            t_results.remove += objs_to_remove
+
+            # Spawn in garbage fluid
+            container.states[Filled].set_value(get_system(system_name=m.DEFAULT_GARBAGE_SYSTEM), True)
```
ChengshuLi, line 1101: nit: is it possible to include this as an actual recipe and we intentionally place this recipe in the last element of `cls._ACTIVE_RECIPES`? Then all the code above can be reused.

omnigibson/object_states/factory.py
cremebrule: change to `mixing_tool`

omnigibson/simulator.py
cremebrule: maybe something like `_sync_apis`?

omnigibson/utils/usd_utils.py
cremebrule: 3 - impulse vector. no contact = 0, 0, 0

omnigibson/utils/usd_utils.py
cremebrule: you're right! d'oh

omnigibson/transition_rules.py
cremebrule: make change condition a wrapper

omnigibson/transition_rules.py
cremebrule: make _RECIPES initialized separately since recipe list is different between mixing tool and blending rules

omnigibson/transition_rules.py
cremebrule: generate correct amount of garbage. maybe helper function?

omnigibson/transition_rules.py
cremebrule: - assert no input objects in mixing wand rule
- allow non-relevant objects to be in the container for mixing wand rule
    - do this by overlap_sphere in generate_particles_from_link call, and make the filtering a utility function
- handle visual particle systems as inputs as well, check for "CoveredAny"

omnigibson/simulator.py
cremebrule: can't do utility function, the two lines aren't the same

omnigibson/simulator.py
cremebrule: added doc we can add to

omnigibson/utils/usd_utils.py
cremebrule: done

omnigibson/transition_rules.py
cremebrule: - only have "filters", and have separate property individual_filter_names
- move ACTIVE_RULE information into the rule directly since it's monolithic

omnigibson/transition_rules.py
```diff
@@ -103,551 +263,952 @@ def __call__(self, obj):
...
 
-                    {"apple": CategoryFilter("apple"), "knife": CategoryFilter("knife")},
+        # If objs is empty, return False, otherwise, True
+        return len(objs) > 0
 
-                the transition rule step will produce the following dictionary:
+    @classproperty
+    def modifies_filter_names(self):
+        # Only modifies values from filter 1
+        return set((self._filter_1_name,))
```
ChengshuLi, line 355: why is this a set of tuple? Instead of a set?

omnigibson/transition_rules.py
```diff
@@ -103,551 +263,952 @@ def __call__(self, obj):
...
-    # Strawberry smoothie
-    BlenderRule(
-        output_system="strawberry_smoothie",
-        particle_requirements={"milk": 10},
-        obj_requirements={"strawberry": 5, "ice_cube": 5},
-    ),
+# Create strawberry smoothie blender rule
+BlenderRule.add_recipe(
+    name="strawberry_smoothie_recipe",
+    input_objects={"strawberry": 5, "ice_cube": 5},
```
ChengshuLi, line 1211: Are we going to check for "categorical correctness" only? I guess we can still use your code here, but set "strawberry": 1 and "ice_cube": 1.

ChengshuLi (approved):
LGTM. Great work!!!

omnigibson/transition_rules.py
```diff
@@ -103,551 +263,952 @@ def __call__(self, obj):
...
+        for system in VisualParticleSystem.get_active_systems():
+            group_name = system.get_group_name(container)
+            if group_name in system.groups and system.num_group_particles(group_name) > 0:
+                system.remove_all_group_particles(group=group_name)
+
+        # Remove all contained objects if requested
+        if not cls.ignore_nonrecipe_objects:
+            # Remove all objects inside the container
+            objs_to_remove = cls._OBJECTS[np.where(in_volume)[0]].tolist()
+            volume += sum(obj.volume for obj in objs_to_remove)
```
ChengshuLi, line 1119: for mixing_ward,

if matching recipe: the recipe-relevant object should be removed, other rigid objects stay intact
if not (garbage): all rigid bodies should stay intact

omnigibson/transition_rules.py
cremebrule: because `self._filter_1_name` is a str, so e.g.: `set("hello")` = `{'h', 'e', 'l', 'l', 'o'}` xD

omnigibson/transition_rules.py
cremebrule: exactly. at least this future-proofs us in case we do implement some more fine-grained, quantity-based requirement

omnigibson/transition_rules.py
cremebrule: whoops! great point. fixed (tho to be clear: mixing_wand rules should never have any recipe-relevant objects, right?)

ChengshuLi (approved):

PR #170, opened 2023-04-27
Fix rigid contact api

cremebrule:


omnigibson/prims/entity_prim.py
```diff
@@ -1160,7 +1160,8 @@ def kinematic_only(self, val):
                 body's pose / velocities. See https://docs.omniverse.nvidia.com/app_create/prod_extensions/ext_physics/rigid-bodies.html?highlight=rigid%20body%20enabled#kinematic-rigid-bodies
                 for more information
         """
-        self.root_link.kinematic_only = val
+        for link in self.links.values():
+            link.kinematic_only = val
```
ChengshuLi, line 1164: This is handling the case where the object has fixed joints?

omnigibson/systems/micro_particle_system.py
```diff
@@ -904,8 +906,9 @@ def generate_particles_from_link(
         # Check which points are inside the volume and only keep those
         particle_positions = particle_positions[np.where(check_in_volume(particle_positions))[0]]
 
-        # Also prune any that in contact with anything
-        particle_positions = particle_positions[np.where(cls.check_in_contact(particle_positions))[0]]
+        # Also prune any that in contact with anything if requested
+        if check_contact:
+            particle_positions = particle_positions[np.where(cls.check_in_contact(particle_positions) == 0)[0]]
```
ChengshuLi, line 911: why do we optionally `check_in_contact`? Is it for performance reasons?

tests/test_envs.py
```diff
@@ -28,9 +28,10 @@ def task_tester(task_type):
     # Make sure sim is stopped
     og.sim.stop()
 
-    # Make sure GPU dynamics are enabled (GPU dynamics needed for cloth)
+    # Make sure GPU dynamics are enabled (GPU dynamics needed for cloth) and no flatcache
     gm.ENABLE_OBJECT_STATES = True
     gm.USE_GPU_DYNAMICS = True
+    gm.ENABLE_FLATCACHE = False
```
ChengshuLi, line 34: why no flatcache?

tests/utils.py
```diff
@@ -67,9 +67,10 @@ def assert_test_scene():
         # Make sure sim is stopped
         og.sim.stop()
 
-        # Make sure GPU dynamics are enabled (GPU dynamics needed for cloth)
+        # Make sure GPU dynamics are enabled (GPU dynamics needed for cloth) and no flatcache
         gm.ENABLE_OBJECT_STATES = True
         gm.USE_GPU_DYNAMICS = True
+        gm.ENABLE_FLATCACHE = False
```
ChengshuLi, line 73: Same.

omnigibson/transition_rules.py
```diff
@@ -323,25 +323,48 @@ def __init__(self, filter_1_name, filter_2_name):
         # Will be filled in during self.initialize
         # Maps object to the list of rigid body idxs in the global contact matrix corresponding to filter 1
         self._filter_1_idxs = None
-        # List of rigid body idxs in the global contact matrix corresponding to filter 2
+        # Maps object to the list of rigid body idxs in the global contact matrix corresponding to filter 2
         self._filter_2_idxs = None
+        # Maps object to set of rigid bodies corresponding to filter 2
+        self._filter_2_bodies = None
```
ChengshuLi, line 329: nit: maybe add a comment here saying `self. _filter_2_idxs` will be used for the optimized mode and `self. _filter_2_bodies` will be used for the non-optimized mode.

ChengshuLi (approved):

omnigibson/prims/entity_prim.py
cremebrule: yep, exactly!

cremebrule: this was causing the code to break when loading a normal scene

omnigibson/systems/micro_particle_system.py
cremebrule: Actually, 2 reasons:
1. Performance, as you mentioned
2. To avoid inaccurate results. Case in point: Dicing: where we teleport / destroy the object, but physics doesn't get updated until you take a step, so the check_in_contact call will actually be invalid. So we don't run check_in_contact in that case

tests/test_envs.py
cremebrule: We get an assertionerror, because cloth isn't compatible with flatcache

omnigibson/transition_rules.py
cremebrule: done!

tests/test_envs.py
ChengshuLi: Got it!

ChengshuLi:
Feel free to merge! LGTM @cremebrule

PR #174, opened 2023-05-02
Improve systems

cremebrule:


omnigibson/systems/micro_particle_system.py
```diff
@@ -1269,6 +1269,9 @@ def initialize(cls):
         cls._material.shader_force_populate()
         # Potentially modify the material
         cls._customize_particle_material()
+        # Apply the physical material preset based on whether or not this fluid is viscous
+        apply_mat_physics = particleUtils.AddPBDMaterialViscous if cls.is_viscous else particleUtils.AddPBDMaterialWater
```
ChengshuLi, line 1273: So this is a physical material wrapper on top of an existing visual material?

ChengshuLi (approved):

omnigibson/systems/micro_particle_system.py
cremebrule: all the values in the physical material are set to 0 by default it seems. These actually set the parameters in the physical material, and I empirically noticed a difference in behavior when swapping between the two physical materials

PR #185, opened 2023-05-23
Improve contains

cremebrule:


ChengshuLi (commented):
LGTM with some minor comments!!

omnigibson/utils/transform_utils.py
```diff
@@ -468,10 +468,10 @@ def quat2mat(quaternion):
     Converts given quaternion to matrix.
 
     Args:
-        quaternion (np.array): (x,y,z,w) vec4 float angles
+        rmat (np.array): (..., 4) (x,y,z,w) float quaternion angles
```
ChengshuLi, line 471: minor typo `rmat`

omnigibson/object_states/contains.py
```diff
@@ -49,16 +53,33 @@ def _get_value(self, system):
...
+
+        # First, we check what type of system
+        # Currently, we support VisualParticleSystems and PhysicalParticleSystems
+        if issubclass(system, VisualParticleSystem):
+            if self._visual_particle_group in system.groups:
+                # Grab global particle poses and offset them in the direction of their orientation
+                raw_positions, quats = system.get_particles_position_orientation(group=self._visual_particle_group)
+                unit_z = np.zeros((len(raw_positions), 3, 1))
+                unit_z[:, -1, :] = m.VISUAL_PARTICLE_OFFSET
+                checked_positions = (T.quat2mat(quats) @ unit_z).reshape(-1, 3) + raw_positions
```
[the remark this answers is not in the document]
cremebrule: Right, great points!

- re enlarging the fillable volume itself: This itself isn't that trivial. I suppose we could add an offset for every face when construct the `check_volume` function, but that seems a bit more clunky and would require more overhead for this single niche use case.
- re proposed method: yes, exactly! It's moving each particle away from its sampled surface. This assumes the particle's normal is facing away from the object's surface, which I believe is an okay assumption assuming the original Covered sampling code works as expected

omnigibson/utils/transform_utils.py
cremebrule: fixed!

ChengshuLi (approved):

PR #198, opened 2023-06-20
Fix/debug sampling

cremebrule:


omnigibson/object_states/particle_modifier.py
```diff
@@ -654,7 +821,9 @@ def _initialize(self):
         system = get_system(system_name)
 
         if self.method == ParticleModifyMethod.PROJECTION and self._projection_mesh_params["visualize"]:
-            radius, height = self._projection_mesh_params["extents"][0] / 2.0, self._projection_mesh_params["extents"][2]
+            assert self._projection_mesh_params["type"] in {"Cylinder", "Cone"}, \
+                f"{self.__class__.__name__} visualization only supports Cylinder and Cone types!"
+            radius, height = np.mean(self._projection_mesh_params["extents"][:2]) / 2.0, self._projection_mesh_params["extents"][2]
```
ChengshuLi, line 826: is np.mean necessary? For "Cylinder" and "Cone", isn't `self._projection_mesh_params["extents"][0] == self._projection_mesh_params["extents"][1]`?

ChengshuLi: Two remaining issues:
- Check if meta meshes exist (link_prefix_0_0/mesh_0)
- Make sure meta meshes for ToggledOn and ParticleApplier are not stretched into different proportions when the object is scaled by different amount. Let's revisit this to see if this is still a good idea or not.

omnigibson/object_states/cloth.py
```diff
@@ -0,0 +1,35 @@
...
+    This class is a subclass of BaseObjectState that adds dependencies
+    on the default kinematics states.
+    """
+    def _initialize(self):
+        # Run super
+        super()._initialize()
+
+        # Make sure the owned object is a cloth type
+        assert self.obj.prim_type == PrimType.CLOTH, f"Cannot use ClothState {self.__class__.__name__}, " \
+                                                     f"make sure object is created with prim_type=PrimType.CLOTH!"
```
ChengshuLi, line 17: let's remove this.

omnigibson/objects/stateful_object.py
```diff
@@ -218,7 +218,16 @@ def prepare_object_states(self, abilities=None, include_default_states=True):
...
         for state_type, params in reversed(state_types_and_params):
-            self._states[state_type] = state_type(obj=self, **params)
+            relevant_params = extract_class_init_kwargs_from_dict(cls=state_type, dic=params, copy=False)
+            compatible, reason = state_type.is_compatible(obj=self, **relevant_params)
+            if compatible:
+                self._states[state_type] = state_type(obj=self, **relevant_params)
+            else:
+                log.warning(f"State {state_type.__name__} is incompatible with obj {self.name}. Reason: {reason}")
+                # Remove the ability if it exists
+                ability = state_type_to_ability.get(state_type, None)
```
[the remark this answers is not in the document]
ChengshuLi: add a comment

omnigibson/systems/micro_particle_system.py
```diff
@@ -430,8 +430,9 @@ def initialize(cls):
         cls.system_prim = cls._create_particle_system()
         # Create material
         cls._material = cls._create_particle_material_template()
-        # Load the material
-        cls._material.load()
+        # Load the material if not already loaded
+        if not cls._material.loaded:
+            cls._material.load()
```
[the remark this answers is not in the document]
ChengshuLi: add a TODO: remove system should completely remove the relevant prims from the stage

omnigibson/utils/bddl_utils.py
```diff
@@ -550,17 +562,20 @@ def _build_non_sampleable_object_scope(self):
...
+                obj_synset = self._object_instance_to_synset[obj_inst]
 
                 # We allow burners to be used as if they are stoves
-                categories = OBJECT_TAXONOMY.get_subtree_categories(obj_cat)
-                if obj_cat == "stove.n.01":
-                    categories += OBJECT_TAXONOMY.get_subtree_categories("burner.n.02")
+                categories = OBJECT_TAXONOMY.get_subtree_categories(obj_synset)
+                abilities = OBJECT_TAXONOMY.get_abilities(obj_synset)
+
+                # Grab all models that fully support all abilities for the corresponding category
```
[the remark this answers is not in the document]
ChengshuLi: not a bottleneck.

omnigibson/object_states/particle_modifier.py
ChengshuLi: - Support cube and sphere when creating visual meshes
- Set scale for these meshes based on extents, rather than setting size/radius/height

ChengshuLi: Add an additional comment:

```
    # Scale radius and height by the parent scale -- projection always points in the negative-z direction of the
    # parent frame
    # We do this AFTER we create the source sphere because the actual projection is scaled in the world frame, whereas
    # the source sphere is already scaled by its own parent frame
    projection_radius *= np.mean(parent_scale[:2])
    projection_height *= parent_scale[2]
```

saying the visualization will NOT match the meta mesh if the parent scale non-uniformally

docker/run_docker.sh
```diff
@@ -103,7 +103,7 @@ while true; do
     esac
 done
 
-docker pull stanfordvl/omnigibson:latest
```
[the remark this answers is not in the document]
cremebrule: fixed

omnigibson/object_states/heat_source_or_sink.py
```diff
@@ -81,10 +81,29 @@ def __init__(
...
+    def is_compatible(cls, obj, **kwargs):
+        # Run super first
+        compatible, reason = super().is_compatible(obj, **kwargs)
+        if not compatible:
+            return compatible, reason
+
+        # Check whether this state has toggledon if required or open if required
+        for kwarg, state_type in zip(("requires_toggled_on", "requires_closed"), (ToggledOn, Open)):
+            if kwargs.get(kwarg, False) and state_type not in obj.states:
+                return False, f"{cls.__name__} has requires_toggled_on but obj has no {state_type.__name__} state!"
```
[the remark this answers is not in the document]
cremebrule: fixed

omnigibson/objects/stateful_object.py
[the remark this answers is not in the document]
cremebrule: done

omnigibson/systems/micro_particle_system.py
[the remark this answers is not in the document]
cremebrule: done

tests/test_envs.py
```diff
@@ -58,3 +58,5 @@ def test_point_navigation_task():
 
 def test_behavior_task():
     task_tester("BehaviorTask")
+
+test_behavior_task()
```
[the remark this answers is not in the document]
cremebrule: done

tests/test_object_states.py
```diff
@@ -922,3 +923,7 @@ def test_covered():
 
 def test_clear_sim():
     og.sim.clear()
+
+# test_covered()
+# from IPython import embed; embed()
+# og.shutdown()
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/object_states/particle_source_or_sink.py
```diff
@@ -15,6 +15,9 @@
 m.SOURCE_LINK_PREFIX = "fluidsource"
 m.SINK_LINK_PREFIX = "fluidsink"
 
+# Default radius and height
+m.DEFAULT_RADIUS_HEIGHT = 0.2
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/object_states/particle_modifier.py
```diff
@@ -173,34 +172,40 @@ class ParticleModifier(AbsoluteObjectState, LinkBasedStateMixin, UpdateStateMixi
...
+            form above -- if all of its conditions evaluate to True and particles are detected within
             this particle modifier area, then we potentially modify those particles
+        method (ParticleModifyMethod): Method to modify particles. Current options supported are ADJACENCY (i.e.:
+            "touching" particles) or PROJECTION (i.e.: "spraying" particles)
         projection_mesh_params (None or dict): If specified and @method is ParticleModifyMethod.PROJECTION,
             manually overrides any metadata found from @obj.metadata to infer what projection volume to generate
             for this particle modifier. Expected entries are as follows:
 
-                "type": (str), one of {"Cylinder", "Cone"}
+                "type": (str), one of {"Cylinder", "Cone", "Cube"}
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/object_states/cloth.py
cremebrule: done

omnigibson/object_states/particle_modifier.py
cremebrule: done

omnigibson/object_states/particle_modifier.py
```diff
@@ -254,52 +272,64 @@ def overlap_callback(hit):
...
+                for shape_attr, default_val in shape_defaults.items():
+                    if shape_attr in property_names:
+                        mesh.GetAttribute(shape_attr).Set(default_val)
+            else:
+                # Potentially populate projection mesh params if the prim exists
+                if self._projection_mesh_params is None:
+                    self._projection_mesh_params = {
+                        "type": pre_existing_mesh.GetTypeName(),
+                        "extents": np.array(pre_existing_mesh.GetAttribute("xformOp:scale").Get()),
+                    }
```
ChengshuLi, line 299: So if pre_exsiting_mesh is True, does it take priority over `self._projection_mesh_params`? I thought it was the opposite based on our last discussion. I don't have a strong opinion on this tho but I just want to point this out.

omnigibson/object_states/toggle.py
```diff
@@ -41,30 +47,58 @@ def _set_value(self, new_value):
     def _initialize(self):
         super()._initialize()
         self.initialize_link_mixin()
-        mesh_prim_path = f"{self.link.prim_path}/visual_marker"
+        mesh_prim_path = f"{self.link.prim_path}/mesh_0"
+        pre_existing_mesh = get_prim_at_path(mesh_prim_path)
         # Create a primitive mesh if it doesn't already exist
-        if not get_prim_at_path(mesh_prim_path):
+        if not pre_existing_mesh:
```
ChengshuLi, line 53: hmm why are we creating a sphere mesh here rather than a sphere shape? I thought the latter will be better for consistency.

omnigibson/object_states/toggle.py
```diff
@@ -41,30 +47,58 @@ def _set_value(self, new_value):
...
+            # Continue traversal only if we don't have a valid hit yet
+            return not valid_hit
+
+        def check_overlap():
+            nonlocal valid_hit
+            valid_hit = False
+            og.sim.psqi.overlap_shape(*projection_mesh_ids, reportFn=overlap_callback)
+            return valid_hit
+
+        self._check_overlap = check_overlap
```
ChengshuLi, line 90: Do we not have a helper function like this in our codebase?

omnigibson/objects/object_base.py
```diff
@@ -207,7 +207,7 @@ def articulation_root_path(self):
         if self.kinematic_only or ((not has_articulated_joints) and (not has_fixed_joints)):
             # Kinematic only, or non-jointed single body objects
             return None
-        elif not self.fixed_base:
+        elif not self.fixed_base and has_articulated_joints:
```
ChengshuLi, line 210: hmmm I thought we debug this extensively before, is it still buggy? 

What behavior did this code change?

not kinematic_only
not fixed_base
not has_articulated_joints
yes has_fixed_joints

e.g. a cup with a fillable volume meta link.

In this case, what should be the articulation_root_path?

omnigibson/simulator.py
```diff
@@ -421,6 +421,8 @@ def _non_physics_step(self):
                 # Note that we don't explicitly do for obj in self._objects_to_initialize because additional objects
                 # may be added mid-iteration!!
                 # For this same reason, after we finish the loop, we keep any objects that are yet to be initialized
+                # First call zero-physics step update, so that handles are properly propagated
+                og.sim.pi.update_simulation(elapsedStep=0, currentTime=og.sim.current_time)
```
ChengshuLi, line 425: hmm how comes we didn't catch this before?

omnigibson/systems/system_base.py
```diff
@@ -1109,32 +1112,54 @@ def _customize_particle_material(mat: MaterialPrim): --> None
...
-                    include_default_states=False,
-                    abilities={},
-                )
+        asset_path = os.path.join(system_dir, f"{system_name}.usd")
+        has_asset = os.path.exists(asset_path)
+        if not has_asset:
+            if system_type == "macro_visual_particle":
+                # Fallback to stain asset
+                asset_path = os.path.join(gm.ASSET_PATH, "models", "stain", "stain.usd")
+                has_asset = True
```
ChengshuLi, line 1121: Let's put the stain particle to be under `og_dataset/objects`?

ChengshuLi (approved):

omnigibson/object_states/particle_modifier.py
cremebrule: Nope -- that's why there's the if statement in L295 -- it only gets populated by the externally-defined mesh if there's no params specified.

Note that this guarantees that `projection_mesh_params` are always populated, so that a few lines later (L315) it can be used to set the projection mesh scale

omnigibson/object_states/toggle.py
cremebrule: whoops! great point. fixed

omnigibson/object_states/toggle.py
cremebrule: nope...it's a bit tricky, because a lot of these variables can change between calls. e.g.: `self._robot_link_paths` might not always be the same each time.

omnigibson/objects/object_base.py
cremebrule: The interesting thing is that the comment in the immediately proceeding line suggests that the (fixed) code is actually how the logic _should_ be.

I think originally, non-articulated, non-fixed objects that had metalinks (fixed joints) were treated as kinematic-only objects, until I found out that that caused some USD issues. So I updated the logic so that this category of objects is not satisfied by that first `if` statement, but didn't change anything else -- so I think it, by default, went to the `else` clause. But really it belonged to the `elif` clause, since moving the articulation root path from the object prim to the root link prim is only done out of _necessity_ for articulated, non-fixed objects since their joint ordering gets messed up by omniverse otherwise.

To answer your specific questions, yes, that is the specific case I'm talking about, and the articulation_root_path should remain unchanged (i.e.: exist at the object prim level, NOT the root link prim level)

omnigibson/simulator.py
cremebrule: I think it's because we never tried to initialize the edge case of:

not kinematic_only
not fixed_base
not has_articulated_joints
yes has_fixed_joints

AFTER the sim starts playing. Note that if the sim is stopped and we import an object like the above and then play, we never run into issues. I only ran into this issue now because I was testing a manually-defined particleapplier which was imported after sim was playing.

omnigibson/systems/system_base.py
cremebrule: Yep, sg! I think we should only make this change once we have a finalized dataset to be released

omnigibson/simulator.py
cremebrule: test with GPU_DYNAMICS with articulated object, e.g.: robot or cabinet

omnigibson/object_states/particle_modifier.py
cremebrule: remove mesh if type mismatch beforehand

omnigibson/simulator.py
cremebrule: result: doesn't work )):

omnigibson/object_states/particle_modifier.py
cremebrule: result: can't delete mesh, instead, assert if projection_mesh_params type and pre-existing mesh type don't match

PR #200, opened 2023-06-30
Protect against planar objects being scaled to NaN

cgokmen:


omnigibson/objects/dataset_object.py
```diff
@@ -268,6 +268,7 @@ def _post_load(self):
         # Otherwise, if manual bounding box is specified, scale based on ratio between that and the native bbox
         elif self._load_config["bounding_box"] is not None:
             scale = self._load_config["bounding_box"] / self.native_bbox
+            scale[np.isnan(scale)] = 1.0
```
cremebrule, line 271: maybe this can be cleaner to avoid nan warnings?

```python
with np.errstate(divide='ignore', invalid='ignore'):
    scale = self._load_config["bounding_box"] / self.native_bbox
scale[np.isnan(scale)] = 1.0
```

cremebrule (approved):
lgtm with small comment

cremebrule (approved):
lgtm!

PR #202, opened 2023-07-01
Improve scale checking code, disable 90deg scaling rule temporarily

cgokmen:


ChengshuLi (approved):

omnigibson/objects/dataset_object.py
```diff
@@ -270,11 +270,14 @@ def _post_load(self):
         # Otherwise, if manual bounding box is specified, scale based on ratio between that and the native bbox
         elif self._load_config["bounding_box"] is not None:
             scale = np.ones(3)
-            valid_idxes = ~np.isclose(self.native_bbox, 0.0)
+            valid_idxes = self.native_bbox > 1e-4
```
ChengshuLi, line 273: These magic numbers seem a bit strange. 

Why not just do the following?
```
valid_idxes = np.nonzero(self.native_bbox)
```

cgokmen: the scale is many times not actually zero due to numerical errors applying transformations in trimesh but a very very small number (1e-7 etc) and I experimentally decided 1e-4 is a reasonable threshold. we can revisit all this after I have solved this issue to ensure there's a reasonable set of checks etc in the edited files here

PR #207, opened 2023-07-03
Fix/particle modifier

cremebrule:


omnigibson/object_states/particle_modifier.py
```diff
@@ -240,13 +240,6 @@ def is_compatible(cls, obj, **kwargs):
 
         return True, None
 
-    def initialize_link_mixin(self):
-        # Run super first
-        super().initialize_link_mixin()
-
-        # Make sure there's at most only a single metalink
-        assert len(self.links) <= 1, f"Can only have at most a single metalink for {self.__class__.__name__}!"
-
```
ChengshuLi, line 249: why remove?

omnigibson/prims/rigid_prim.py
```diff
@@ -175,11 +178,15 @@ def update_meshes(self):
...
-                    volume, com = get_mesh_volume_and_com(mesh_prim)
+                    is_volume, volume, com = get_mesh_volume_and_com(mesh_prim)
                     vols.append(volume)
                     # We need to translate the center of mass from the mesh's local frame to the link's local frame
                     local_pos, local_orn = mesh.get_local_pose()
                     coms.append(T.quat2mat(local_orn) @ (com * mesh.scale) + local_pos)
+                    # If we're not a valid volume, use bounding box approximation for the underlying collision approx
+                    if not is_volume:
+                        log.warning(f"Got invalid (non-volume) collision mesh: {mesh.name}")
+                        mesh.set_collision_approximation("boundingCube")
```
ChengshuLi, line 189: hmm if we don't do this, what would happen?

ChengshuLi (approved):

omnigibson/object_states/particle_modifier.py
cremebrule: there's objects that have multiple sources (e.g.: sink with two faucets), and I couldn't remember why I implemented this constraint originally.

Of course, the current code only functionally supports a single link anyways, but that's the same for all of our object states -- we need to handle multi-links properly anyways, it's a future todo

omnigibson/prims/rigid_prim.py
cremebrule: We get the non-compatible GPU mesh error

omnigibson/object_states/particle_modifier.py
ChengshuLi: Got it!

PR #223, opened 2023-07-20
Faster motion planning

sujaygarlanka:
### Major
- Implement faster motion planning by copying robot meshes and checking for mesh overlap during planning.
### Minor
- Fixed issue with IK where we were not correctly detecting when the position was not reachable
- Fixed precision error with planning where returned IK positions that were at the joint limits were sometimes marginally over or under, causing OMPL to fail
- Fixed issue with grasping due to the grasping frame for Tiago having a very large collision mesh
- Fixed issue with navigation where robot was sometimes navigating with a large margin of error to specified pose


omnigibson/utils/motion_planning_utils.py
```diff
@@ -133,7 +143,7 @@ def state_valid_fn(q):
 
     if solved:
         # try to shorten the path
-        ss.simplifySolution()
+        # ss.simplifySolution()
```
cgokmen, line 146: Why not?

omnigibson/utils/motion_planning_utils.py
```diff
@@ -159,22 +192,55 @@ def detect_robot_collision(robot, filter_objs=[]):
...
+
+    for link in arm_links:
+        pose = link_poses[link]
+        if link in context.robot_meshes_copy.keys():
+            for mesh, relative_pose in zip(context.robot_meshes_copy[link], context.robot_meshes_relative_poses[link]):
+                mesh_pose = T.pose_transform(*pose, *relative_pose)
+                translation = Gf.Vec3d(*np.array(mesh_pose[0], dtype=float))
+                mesh._prim.GetAttribute("xformOp:translate").Set(translation)
+                orientation = np.array(mesh_pose[1], dtype=float)[[3, 0, 1, 2]]
+                mesh._prim.GetAttribute("xformOp:orient").Set(Gf.Quatd(*orientation))
```
cgokmen, line 211: What's the type of `mesh` here, is it an OmniGibson object/prim? If so, you can use the set_position_orientation API directly rather than directly using the prim API. The cost of using the OmniGibson interface is wrapping the object at first which seems like it's already done by here.

omnigibson/utils/motion_planning_utils.py
```diff
@@ -159,22 +192,55 @@ def detect_robot_collision(robot, filter_objs=[]):
...
+                if valid_hit:
+                    return valid_hit
+                mesh_id = PhysicsSchemaTools.encodeSdfPath(mesh.prim_path)
+                mesh_hit = mesh.prim_path
+                if mesh._prim.GetTypeName() == "Mesh":
+                    og.sim.psqi.overlap_mesh(*mesh_id, reportFn=overlap_callback)
+                else:
+                    og.sim.psqi.overlap_shape(*mesh_id, reportFn=overlap_callback)
+            
+        return valid_hit
```
cgokmen, line 241: We can unwrap this function now and directly execute this code rather than calling it. It'll remove some of the annoying logic around the nonlocal stuff. Also, do the `if valid hit` part *after* the overlap call, not before. Also the logic around `mesh_hit` is very confusing too.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -88,41 +98,149 @@ def indented_print(msg, *args, **kwargs):
...
+            og.sim.remove_object(robot_to_copy)
+
+        self._disable_robot_colliders()
+        og.sim.step()
+
+    def _set_prim_pose(self, prim, pose):
+        translation = Gf.Vec3d(*np.array(pose[0], dtype=float))
+        prim.GetAttribute("xformOp:translate").Set(translation)
+        orientation = np.array(pose[1], dtype=float)[[3, 0, 1, 2]]
+        prim.GetAttribute("xformOp:orient").Set(Gf.Quatd(*orientation))
```
cgokmen, line 199: See later comment on one of the below files on this. If we have the prim as an OmniGibson object, there's no gain from calling this API manually

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -88,41 +98,149 @@ def indented_print(msg, *args, **kwargs):
...
+                    for mesh in link.collision_meshes.values():
+                        mesh.collision_enabled = False
+                        self.disabled_meshes.append(mesh)
+
+        # Disable object in hand
+        obj_in_hand = self.robot._ag_obj_in_hand[self.robot.default_arm] 
+        if obj_in_hand is not None:
+            for link in obj_in_hand.links.values():
+                    for mesh in link.collision_meshes.values():
+                        mesh.collision_enabled = False
```
cgokmen, line 223: We need comments on what we are disabling here and why

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -156,25 +274,15 @@ def __init__(self, task, scene, robot, teleport=False):
...
-                # print(damps)
-                # print("-------")
+            self._setup_tiago()
+
+    # Disable grasping frame for Tiago robot (Should be cleaned up in the future)
+    def _setup_tiago(self):
+        for link in self.robot.links.values():
+            for mesh in link.collision_meshes.values():
+                if "grasping_frame" in link.prim_path:
+                    mesh.collision_enabled = False
```
cgokmen, line 285: If we are doing this here, let's add the same thing for Fetch to update the base mass. Also let's open tickets on this repo for fixes to this if there's something for the OG team to do.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -453,12 +561,13 @@ def _convert_cartesian_to_joint_space(self, target_pose):
             )
         return joint_pos, control_idx
     
-    def _target_in_reach_of_robot(self, target_pose):
+    def _target_in_reach_of_robot(self, target_pose, is_relative=False):
+        if not is_relative:
+            target_pose = self._get_pose_in_robot_frame(target_pose)
```
cgokmen, line 566: Create a separate function (_target_in_reach_of_robot, _target_in_reach_of_robot_relative) for this instead of taking in an argument. Also document both.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -783,32 +896,33 @@ def _rotate_in_place(self, end_pose, angle_threshold = DEFAULT_ANGLE_THRESHOLD):
...
-        if pos_on_obj is None:
+    def _sample_pose_near_object(self, obj, pose_on_obj=None, **kwargs):
+        if pose_on_obj is None:
             pos_on_obj = self._sample_position_on_aabb_face(obj)
+            pose_on_obj = np.array([pos_on_obj, [0, 0, 0, 1]])
+
+        with UndoableContext(self.robot, "base") as context:
+            obj_rooms = obj.in_rooms if obj.in_rooms else [self.scene._seg_map.get_room_instance_by_point(pose_on_obj[0][:2])]
+            for _ in range(MAX_ATTEMPTS_FOR_SAMPLING_POSE_NEAR_OBJECT):
+                distance = np.random.uniform(0.2, 1.0)
```
cgokmen, line 907: Reflect update to 0 here for lower bound

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -783,32 +896,33 @@ def _rotate_in_place(self, end_pose, angle_threshold = DEFAULT_ANGLE_THRESHOLD):
...
         yield self._empty_action()
             
-    def _sample_pose_near_object(self, obj, pos_on_obj=None, **kwargs):
-        if pos_on_obj is None:
+    def _sample_pose_near_object(self, obj, pose_on_obj=None, **kwargs):
+        if pose_on_obj is None:
             pos_on_obj = self._sample_position_on_aabb_face(obj)
+            pose_on_obj = np.array([pos_on_obj, [0, 0, 0, 1]])
+
+        with UndoableContext(self.robot, "base") as context:
```
cgokmen, line 904: Why are we doing an undoablecontext here?

cgokmen (changes requested):

omnigibson/robots/fetch.py
```diff
@@ -382,6 +382,47 @@ def disabled_collision_pairs(self):
             ["torso_lift_link", "shoulder_lift_link"],
             ["torso_lift_link", "torso_fixed_link"],
         ]
+    
+    @property
+    def primitive_disabled_collision_pairs(self):
```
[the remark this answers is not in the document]
sujaygarlanka: Adding base link to disabled collision pairs is throwing an error, so had to create a new property that I could use to get the pairs. Issue in this slack conversation https://stanfordvl.slack.com/archives/C05DFPGND7G/p1689372142805389

omnigibson/robots/tiago.py
```diff
@@ -499,7 +499,9 @@ def camera_control_idx(self):
 
     @property
     def arm_control_idx(self):
-        return {"left": np.array([7, 10, 13, 15, 17, 19, 21]), "right": np.array([8, 11, 14, 16, 18, 20, 22])}
+        return {"left": np.array([7, 10, 13, 15, 17, 19, 21]),
+                "right": np.array([8, 11, 14, 16, 18, 20, 22]),
+                "combined": np.array([7, 8, 10, 11, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22])}
```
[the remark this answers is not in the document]
sujaygarlanka: @mj-hwang's FK work. Could you explain why?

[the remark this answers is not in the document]
sujaygarlanka: @mj-hwang

omnigibson/utils/motion_planning_utils.py
```diff
@@ -34,16 +32,16 @@ def state_valid_fn(q):
 
     # set lower and upper bounds
     bounds = ob.RealVectorBounds(2)
-    bounds.setLow(-7.0)
-    bounds.setHigh(7.0)
+    bounds.setLow(-3.0)
+    bounds.setHigh(3.0)
```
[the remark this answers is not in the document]
sujaygarlanka: ya, this should be changed. Shouldn't it be the full range of the floor the robot is on?

omnigibson/robots/tiago.py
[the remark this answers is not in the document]
mj-hwang: This combined mode is used when we do forward kinematics for mesh overlap during arm motion planning. Previously, we only figured out left arm and trunk link poses (using "tiago_dual_omnidirectional_stanford_left_arm_descriptor.yaml"), but we weren't able to get the correct link poses for right arm links, since FKSolver would only return default link poses for the uncontrolled arm. For the same reason, I updated the "robot_arm_descriptor_yamls" function and the configuration file in assets ("tiago_dual_omnidirectional_stanford_descriptor.yaml").

omnigibson/utils/motion_planning_utils.py
mj-hwang: Currently, if we simplify the solution, it potentially collides with an object. There may be two reasons.
1. OMPL's simplySolution does not guarantee collision-free simplification. It randomly samples points in trajectory and their intermediate points and removes those points. Perhaps @misoshiruseijin can explain more.
2. Interpolation in JointController and arm mass issue in Tiago.

omnigibson/action_primitives/starter_semantic_action_primitives.py
sujaygarlanka: faster way to test poses of when sampling

omnigibson/utils/motion_planning_utils.py
```diff
@@ -143,7 +153,30 @@ def state_valid_fn(q):
...
+def detect_robot_collision(context, pose):
+    translation = pose[0]
+    orientation = pose[1]
+    # context.robot_copy.prim.set_local_poses(np.array([translation]), np.array([orientation]))
+    translation = Gf.Vec3d(*np.array(translation, dtype=float))
+    context.robot_copy.prim.GetAttribute("xformOp:translate").Set(translation)
+
+    orientation = np.array(orientation, dtype=float)[[3, 0, 1, 2]]
+    context.robot_copy.prim.GetAttribute("xformOp:orient").Set(Gf.Quatd(*orientation)) 
+
```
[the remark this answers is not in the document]
sujaygarlanka: ya, robot copy is at the same joint config because it is set to that joint config when it is copied

omnigibson/utils/motion_planning_utils.py
sujaygarlanka: Using an Omnigibson object here because it has an easy api to remove it from the scene. However, in refactor coming up where the copy is created in the beginning, prims won't need to be removed so the wrapper will be gone.

omnigibson/robots/tiago.py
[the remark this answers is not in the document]
cgokmen: I see, interesting. This bimanual setup does raise quite a lot of headaches. SGTM.

omnigibson/robots/fetch.py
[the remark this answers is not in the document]
cgokmen: I'll follow up on this, thanks.

omnigibson/utils/motion_planning_utils.py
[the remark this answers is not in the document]
cgokmen: Sg, make sure you don't forget about this if/when you remove the undoablecontext later.

omnigibson/utils/motion_planning_utils.py
cgokmen: Aren't we re-interpolating to add waypoints now? @sujaygarlanka

omnigibson/utils/motion_planning_utils.py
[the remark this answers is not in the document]
cgokmen: yes get the union of the floor AABBs

omnigibson/utils/motion_planning_utils.py
```diff
@@ -34,16 +44,18 @@ def state_valid_fn(q):
 
     # set lower and upper bounds
     bounds = ob.RealVectorBounds(2)
-    bounds.setLow(-7.0)
-    bounds.setHigh(7.0)
+    map = og.sim.scene._seg_map
+    map_bound = (map.map_size / 2) * map.map_resolution
+    bounds.setLow(-map_bound)
+    bounds.setHigh(map_bound)
```
cgokmen, line 50: Let's not do this - the map is a static and potentially buggy source of data. We want this to work in arbitrary scenes. Replace this with the AABB union of all the "floors" objects in the scene

omnigibson/utils/motion_planning_utils.py
cgokmen: OK let's do that later. thanks

omnigibson/action_primitives/starter_semantic_action_primitives.py
cgokmen: @sujaygarlanka

cgokmen (changes requested):
See open conversations

omnigibson/utils/motion_planning_utils.py
sujaygarlanka: yes, removed this

omnigibson/action_primitives/starter_semantic_action_primitives.py
sujaygarlanka: Re ran tiago in the sim, and grasping frame is there but collisions are not enabled. Need to look into why it showed up earlier when I was testing. Going to leave this code it for now to make sure it is not an issue. Will investigate and if I can't replicate it again, will remove it. If I can replicate, I'll file an issue with the OG team.

sujaygarlanka: Fetch's base mass should only be higher when it moves its hand. Higher base mass interferes with its navigation, so it should probably not be in the setup, if that is what you mean by update the base mass.

cgokmen (approved):

PR #231, opened 2023-07-26
Refactor collision meshes

sujaygarlanka:
Copy meshes for the robot at the beginning of the creation of the Action Primitives class rather than everytime UndoableContext is called. This results in the following.

- No need to call take a simulator step on every call to UndoableContext
- Replace Omnigibson APIs with Omniverse APIs


omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -271,6 +234,10 @@ def __init__(self, task, scene, robot, teleport=False):
         self.robot_model = self.robot.model_name
         self.robot_base_mass = self.robot._links["base_link"].mass
         self.teleport = teleport
+
+        self.robot_copy = RobotCopy()
+        self._load_robot_copy()
```
cgokmen, line 239: `self.robot_copy = load_robot_copy(robot)`

i.e. let's make load_robot_copy into a static function and make it return a robot copy.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1003,7 +1035,8 @@ def _navigate_to_pose(self, pose_2d):
             self.robot.set_position_orientation(*robot_pose)
             yield from self._settle_robot()
         else:
-            with UndoableContext(self.robot, "base") as context:
+            robot_copy_type = "simplified" if self.robot_model == "Tiago" else "original"
```
cgokmen, line 1038: This is annoying to repeat. Let's put it inside the code for UndoableContext.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -93,117 +93,79 @@
...
         return self 
 
     def __exit__(self, *args):
-        for link in self.robot_meshes_copy:
-            for mesh in self.robot_meshes_copy[link]:
-                mesh.remove()
-        self.robot_copy.remove()
+        for meshes in self.robot_copy.meshes[self.robot_copy_type].values():
+            for mesh in meshes.values():
+                mesh.GetAttribute("physics:collisionEnabled").Set(False)
```
cgokmen, line 120: Two overarching concerns:

1. We leave the robot part in the scene at its last pose, which is possibly a bit annoying (e.g. you'll see the collision mesh there if you have that mode enabled right?) - let's maybe move it elsewhere (z=-100 or something)
2. What part of this make gravity not affect the robot?

dev_environment/test.py
cgokmen, line 1: Can you move this file to `omnigibson/examples/action_primitives/solve_task.py` instead of deleting it?

cgokmen (changes requested):
See comments

omnigibson/action_primitives/starter_semantic_action_primitives.py
sujaygarlanka: 1. You will never see the mesh because it is always disabled after UndoableContext.
2. No mass for the meshes

omnigibson/action_primitives/starter_semantic_action_primitives.py
sujaygarlanka: If I remove that parameter for Tiago and include that logic in UndoableContext, Tiago will always use the simplified model and we don't want that for its arm movement. An option is I use the mode parameter that is "base" or "arm" like I did before and figure out the mesh copy type and and whether we want self collisions in UndoableContext. I removed that because I thought this would make it more clear as to what is happening in UndoableContext. Let me know which you want.

cgokmen (approved):

PR #234, opened 2023-07-27
Add new sampling for pose logic

sujaygarlanka:
- Add new sampling for pose logic that is based on the robot's arm workspace
- Fix bug with grasping where the robot would not detect when the robot completed a grasp


omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -630,6 +632,7 @@ def _place_with_predicate(self, obj, predicate):
         
         obj_pose = self._sample_pose_with_object_and_predicate(predicate, obj_in_hand, obj)
         hand_pose = self._get_hand_pose_for_object_pose(obj_pose)
+        print(hand_pose)
```
cgokmen, line 635: Nit: Remove this

cgokmen (approved):

PR #239, opened 2023-08-01
Add head tracking

sujaygarlanka:
### Features/Updates:
- Add head tracking for an object
- Change arm planning algorithm from RRTConnect to BITstar
- Change Tiago workspace

### Bugs:
- Fix bug in collision detection where base planning was not using FK module to set collision meshes to construct the robot based on current joint positions


omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -345,12 +346,13 @@ def apply(self, action):
         action = StarterSemanticActionPrimitiveSet(action_idx)
         return self.controller_functions[action](target_obj)
     
-    def apply_ref(self, prim, *args, attempts=3):
+    def apply_ref(self, prim, track_object, *args, attempts=3):
```
cgokmen, line 349: Nit: put this as a kwarg before attempts, default it to False.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -368,6 +370,7 @@ def apply_ref(self, prim, *args, attempts=3):
             # Attempt
             success = False
             try:
+                self._tracking_object = args[0] if track_object else None
```
cgokmen, line 373: `if track_object and args`. We shouldn't crash if people call `release` etc

cgokmen (approved):

PR #255, opened 2023-08-11
Fix/behavior env errors

cremebrule:


omnigibson/object_states/particle_modifier.py
```diff
@@ -1193,6 +1205,11 @@ def _get_max_particles_limit_per_step(self, system):
         return m.MAX_VISUAL_PARTICLES_APPLIED_PER_STEP if is_visual_particle_system(system_name=system.name) else \
             m.MAX_PHYSICAL_PARTICLES_APPLIED_PER_STEP
 
+    @property
+    def requires_overlap(self):
+        # Overlap required only if sampling with raycast
+        return self._sample_with_raycast
```
ChengshuLi, line 1211: nit: should we just unify `sample_with_raycast` with `requires_overlap`? Seems weird that we have two different names referring to the exact same thing.

omnigibson/object_states/toggle.py
```diff
@@ -51,16 +52,21 @@ def _initialize(self):
         pre_existing_mesh = get_prim_at_path(mesh_prim_path)
         # Create a primitive mesh if it doesn't already exist
         if not pre_existing_mesh:
-            button = UsdGeom.Sphere.Define(og.sim.stage, Sdf.Path(mesh_prim_path))
             self.radius = m.DEFAULT_RADIUS if self.radius is None else self.radius
+            mesh = create_primitive_mesh(
```
ChengshuLi, line 56: Maybe add a comment here explaining we need to create a sphere mesh rather than a sphere shape because of the physx warning?

omnigibson/object_states/toggle.py
```diff
@@ -51,16 +52,21 @@ def _initialize(self):
...
+                prim_path=mesh_prim_path,
+                primitive_type="Sphere",
+                extents=1.0,
+            )
         else:
             # Infer radius from mesh if not specified as an input
-            self.radius = pre_existing_mesh.GetAttribute("radius").Get() if self.radius is None else self.radius
+            recompute_extents(prim=pre_existing_mesh)
+            scale = np.array(pre_existing_mesh.GetAttribute("xformOp:scale").Get())
```
ChengshuLi, line 64: This is assuming the preexisting meta mesh is sphere mesh, not sphere shape, right?

Have you made the change in ig_pipeline code and re-export the dataset?

omnigibson/object_states/toggle.py
```diff
@@ -51,16 +52,21 @@ def _initialize(self):
...
+            self.radius = np.cbrt(np.product(scale)) / 2.0 if self.radius is None else self.radius
```
ChengshuLi, line 65: Sorry why do we need to do `np.cbrt(np.product(scale))`? Is this trying to be more general and to accommodate arbitrary meshes (that are not spheres)?

omnigibson/object_states/toggle.py
```diff
@@ -51,16 +52,21 @@ def _initialize(self):
...
 
         # Create the visual geom instance referencing the generated mesh prim
         self.visual_marker = VisualGeomPrim(prim_path=mesh_prim_path, name=f"{self.obj.name}_visual_marker")
-        self.visual_marker.set_attribute("radius", self.radius)
-        recompute_extents(prim=self.visual_marker.prim)
+        self.visual_marker.scale = self.radius * 2.0
```
ChengshuLi, line 69: Oh I kind of see what you are trying to do here. So if the pre-existing mesh is non-uniformly scaled, we do the `np.cbrt(np.product(scale))` trick and force it to be uniformly scaled? This is also to prevent the physx warning, right? Maybe add a comment here to elaborate? I think it's a bit opaque currently.

omnigibson/scenes/interactive_traversable_scene.py
```diff
@@ -179,7 +179,7 @@ def _should_load_object(self, obj_info):
         valid_room = self.load_room_instances is None or len(set(self.load_room_instances) & set(in_rooms)) > 0
 
         # Check whether this is an agent and we allow agents
-        agent_ok = self.include_robots or category != robot_macros.ROBOT_CATEGORY
+        agent_ok = self.include_robots or (category != robot_macros.ROBOT_CATEGORY and obj_info["class_name"] not in REGISTERED_ROBOTS)
```
ChengshuLi, line 182: hmm sorry what is this line doing?

omnigibson/utils/usd_utils.py
```diff
@@ -682,25 +685,18 @@ def create_mesh_prim_with_default_xform(primitive_type, prim_path, u_patches=Non
...
-    else:
-        omni.kit.commands.execute(
-            "CreateMeshPrimWithDefaultXform",
-            prim_type=primitive_type,
-            prim_path=prim_path,
-        )
+        kwargs["u_patches"] = u_patches
+        kwargs["v_patches"] = v_patches
+
+    CreateMeshPrimWithDefaultXformCommand(**kwargs).do()
```
ChengshuLi, line 699: interesting syntax, what is the benefit of using this new API?

ChengshuLi (approved):
LGTM, left some minor comments.

omnigibson/object_states/particle_modifier.py
cremebrule: yeah, good point. I think semantically it's a bit more clear though with these two terms -- `requires_overlap` applies to all ParticleModifiers, but `sample_with_raycast` is specific to `ParticleAppliers`, since we're actually sampling particle positions if that is true.

I'm a bit ambivalent though, so don't mind changing it too.

omnigibson/object_states/toggle.py
cremebrule: done!

omnigibson/object_states/toggle.py
cremebrule: Yes, I have made the change in the ig_pipeline code. Waiting for Cem to re-export to verify that everything works as expected.

omnigibson/object_states/toggle.py
cremebrule: well now that we're using a mesh, we can scale stuff arbitrarily.

It's mainly to unify with the input kwarg `radius`, which is expected to be a scalar, not a 3-array. I suppose the reverse could also be done though, where `radius` can be a 3-array or a single number and we'd just apply it uniformly to the scale in the latter case.

But in general, I think can be a bit hard to infer scaling in xyz since it will be in the mesh's local frame -- i.e.: it might be difficult to figure out how the generated mesh will actually scale if you specify 3 different values for x,y,z. A single value for the radius is a lot more interpretable

omnigibson/scenes/interactive_traversable_scene.py
cremebrule: in the new RC we actually shouldn't run into this problem, but previously we weren't storing object categories if they weren't explicitly passed into the object constructor. So for robots, which have (internally) a hardcoded category of "agent", category wasn't getting stored (and the fallback in the code snippet above falls back to category `object`), so we could never filter out the robot because (`agent_ok` was always `True`).

This just adds an extra sanity check to make sure an actual robot class is not being loaded if we don't want to include_robots

omnigibson/utils/usd_utils.py
cremebrule: this (and all of the associated boilerplate code) is to allow us to leverage `create_primitive_mesh` and specify a stage explicitly. Omni's default code automatically uses the active stage, but for our post-processing pipeline where we now want to generate sphere meshes for the toggle button we should explicitly pass in the stage opened for a given object asset

omnigibson/object_states/particle_modifier.py
ChengshuLi: Got it. Let's just keep the status quo then.

omnigibson/object_states/toggle.py
ChengshuLi: Got it.

omnigibson/utils/usd_utils.py
ChengshuLi: I see. For the post-processing pipeline, why is the default active stage not sufficient?

omnigibson/scenes/interactive_traversable_scene.py
ChengshuLi: Got it.

omnigibson/object_states/toggle.py
ChengshuLi: hmmm I am not sure we should do this. 

Here is my suggestion:
```
        if not pre_existing_mesh:
            mesh = create_primitive_mesh(
                prim_path=mesh_prim_path,
                primitive_type="Sphere",
                extents=1.0,
            )
        else:
            # CAN YOU ADD A COMMENT HERE EXPLAINING WHY THIS IS NEEDED?
            recompute_extents(prim=pre_existing_mesh)

        # Create the visual geom instance referencing the generated mesh prim
        self.visual_marker = VisualGeomPrim(prim_path=mesh_prim_path, name=f"{self.obj.name}_visual_marker")
        if not pre_existing_mesh:
            self.radius = m.DEFAULT_RADIUS if self.radius is None else self.radius
            self.visual_marker.scale = self.radius * 2.0
        self.visual_marker.initialize()
```

cremebrule: agreed -- but i thought for now we would always have `pre_existing_mesh` but a sphere?

cremebrule: alternatively, we can have the input be `scale` so it's shape-agnostic

omnigibson/utils/usd_utils.py
cremebrule: you're right -- I didn't think about that. I was thinking about our add_default_metalinks pipeline, which actually has multiple stages open at once, and the active one (the OG one) is not the one we actually want to add metalinks to.

omnigibson/object_states/toggle.py
ChengshuLi: Right, I think I am just not a big fan of a mixture of things. If we have `self.radius`, let's make sure the mesh is actually a sphere (is there way to assert that?). Otherwise, we shouldn't have `self.radius` at all. Should be `self.scale` instead or `self.extent` or something like that.

omnigibson/utils/usd_utils.py
ChengshuLi: Sorry I am not sure I understand your response. So do we have the specify the stage, or is the default stage sufficient?

ChengshuLi: From what you said, it seems the default active stage is NOT sufficient?

omnigibson/object_states/toggle.py
cremebrule: yeah, we can't assert a mesh is a sphere. we should just have self.extent / scale. I will make this update

omnigibson/utils/usd_utils.py
cremebrule: sorry for not being clear. what i mean is:

- in the `generate_metadata` script in `ig_pipeline`, I think the default stage is fine
- in the `generate_default_metalinks` script in OG, the default stage is **not** fine

So I think the change actually wasn't technically needed, but it's good to be consistent in case we need to generate additional mesh-based metalinks in the future not in the ig_pipeline

omnigibson/object_states/toggle.py
cremebrule: done

PR #256, opened 2023-08-15
Make the decrypted context manager work on Windows

cgokmen:
Currently the USD library leaks file handles, meaning any files opened with either Omni or Usd.Stage cannot be deleted during the lifetime of the application. This seeks to avoid that issue by delegating it to the og tempdir (which may or may not be immediately removed, however, the OS should eventually remove the entire directory)


cremebrule (approved):
lgtm

PR #260, opened 2023-08-17
Add motion validator

sujaygarlanka:
Add motion validator for base planning that checks turn-straight-turn motion for the base is collision free.


omnigibson/utils/motion_planning_utils.py
```diff
@@ -27,6 +28,72 @@ def plan_base_motion(
...
+            self.space = space
+
+        def checkMotion(self, s1, s2):
+            if not self.si.isValid(s2):
+                return False
+
+            start = np.array([s1.getX(), s1.getY(), s1.getYaw()])
+            goal = np.array([s2.getX(), s2.getY(), s2.getYaw()])
+            segment = goal[:2] - start[:2]
+            segment_theta = np.arctan2(segment[1], segment[0])
```
cgokmen, line 48: nit: maybe use your self.get_angle_between_poses function

omnigibson/utils/motion_planning_utils.py
```diff
@@ -56,44 +123,54 @@ def state_valid_fn(q):
...
+                    return None
+            elif i == len(return_path) - 1:
+                start_ori =  CustomMotionValidator.get_angle_between_poses(return_path[i-1], pose)
+                if not CustomMotionValidator.is_valid_rotation(si, [pose[0], pose[1], start_ori], pose[2]):
+                    return None
+            else:
+                start_ori =  CustomMotionValidator.get_angle_between_poses(return_path[i-1], pose)
+                final_ori =  CustomMotionValidator.get_angle_between_poses(pose, return_path[i+1])
+                if not CustomMotionValidator.is_valid_rotation(si, [pose[0], pose[1], start_ori], final_ori):
+                    return None
```
cgokmen, line 169: This is not what I meant earlier - we should do our unnecessary rotation removal with this in mind (e.g. only remove rotations that don't make the path invalid). There is a deterministic way to do this that doesn't involve rejecting the path.

cgokmen: My approach essentially involves doing this:

```
def remove_unnecessary_rotations(path):
  # Start at the same starting pose
  new_path = [path[0]]

  # Process every intermediate waypoint
  for i in range(1, len(path) - 1):
    # compute the yaw you'd be at when arriving into path[i] and departing from it
    arriving_yaw = CustomMotionValidator.get_angle_between_poses(path[i-1], path[i])
    departing_yaw = CustomMotionValidator.get_angle_between_poses(path[i], path[i+1])

    # check if you are able to make that rotation directly. 
    arriving_state = (path[i][0], path[i][1], arriving_yaw)
    if CustomMotionValidator.is_valid_rotation(si, arriving_state, departing_yaw):
      # Then use the arriving yaw directly
      new_path.append(arriving_state)
    else:
      # Otherwise, keep the waypoint
      new_path.append(path[i])

  # Don't forget to add back the same ending pose
  new_path.append(path[-1])

  return new_path
```

cgokmen: If you do this, then you don't need to revalidate the path.

PR #269, opened 2023-09-05
fix object control so that zero efforts are applied for non-driven jo…

cremebrule:
…ints

### DIFF
--- a/omnigibson/objects/controllable_object.py
+++ b/omnigibson/objects/controllable_object.py
@@ -436,8 +436,8 @@ class ControllableObject(BaseObject):
             elif ctrl_type == ControlType.POSITION:
                 joint.set_pos(ctrl, normalized=norm, drive=True)
             elif ctrl_type == ControlType.NONE:
-                # Do nothing
-                pass
+                # Set zero efforts
+                joint.set_effort(0, normalized=False)
             else:
                 raise ValueError("Invalid control type specified: {}".format(ctrl_type))
 
--- a/omnigibson/prims/joint_prim.py
+++ b/omnigibson/prims/joint_prim.py
@@ -848,7 +848,6 @@ class JointPrim(BasePrim):
         # Sanity checks -- make sure that we're articulated (no control type check like position and velocity
         # because we can't set effort targets) and that we're driven
         self.assert_articulated()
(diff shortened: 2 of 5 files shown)


omnigibson/objects/controllable_object.py
```diff
@@ -436,8 +436,8 @@ def deploy_control(self, control, control_type, indices=None, normalized=False):
             elif ctrl_type == ControlType.POSITION:
                 joint.set_pos(ctrl, normalized=norm, drive=True)
             elif ctrl_type == ControlType.NONE:
-                # Do nothing
-                pass
+                # Set zero efforts
+                joint.set_effort(0, normalized=False)
```
ChengshuLi, line 440: Could you remind me when ctrl_type will be `NONE` again?

omnigibson/prims/joint_prim.py
```diff
@@ -866,6 +867,9 @@ def keep_still(self):
         Zero out all velocities for this prim
         """
         self.set_vel(np.zeros(self.n_dof))
+        # If not driven, set torque equal to zero as well
+        if not self.driven:
+            self.set_effort(np.zeros(self.n_dof))
```
ChengshuLi, line 872: why not always set torque to 0?

omnigibson/utils/usd_utils.py
```diff
@@ -124,7 +124,7 @@ def get_camera_params(viewport):
     view_proj_mat = helpers.get_view_proj_mat(view_params)
 
     return {
-        "pose": np.array(prim_tf),
+        "pose": np.array(prim_tf).T,
```
ChengshuLi, line 127: hmmm why the transpose?

ChengshuLi (approved):
@cremebrule So sorry for the belated reply. Left some minor comments/questions. Feel free to merge.

omnigibson/prims/joint_prim.py
cremebrule: looping in @cgokmen . You mentioned we actually shouldn't set efforts directly now?

omnigibson/utils/usd_utils.py
cremebrule: I think empirically I just observed that the prim_tf is a transposed homogeneous matrix wrt to what we normally expect I think

PR #270, opened 2023-09-05
Feat/sensor config

cremebrule:


omnigibson/configs/sensors/scan.yaml
```diff
@@ -0,0 +1,30 @@
+# Example ScanSensor sensor config
+# See omnigibson/sensors/__init__/create_sensor and omnigibson/sensors/scan_sensor for docstring of arguments
+# Arguments below are the arguments that should be specified by external user (other kwargs
+# used in constructor are generated automatically at runtime)
+robot:
+  sensor_config:
+    VisionSensor:
```
cgokmen, line 7: ScanSensor?

cgokmen (approved):

omnigibson/configs/sensors/scan.yaml
cremebrule: d'oh! thanks for the great catch. fixed

PR #278, opened 2023-09-22
Add assisted grasp information to simulator state saving and restoring

misoshiruseijin:
- Add AG state to simulator save and restore AG joint when restoring
- Resolved datatype mismatch when restoring Tiago from JSON


omnigibson/robots/manipulation_robot.py
```diff
@@ -1260,7 +1262,17 @@ def _load_state(self, state):
         if self.grasping_mode == "physical":
             return
 
-        # TODO: Include AG_state
+        # Include AG_state
+        # TODO: currently doese not take care of cloth objects
```
ChengshuLi, line 1266: typo "doese"

omnigibson/robots/manipulation_robot.py
```diff
@@ -1260,7 +1262,17 @@ def _load_state(self, state):
...
+        self._ag_obj_constraint_params = state["ag_obj_constraint_params"]
+        for arm in self._ag_obj_constraint_params.keys():
+            if len(self._ag_obj_constraint_params[arm]) > 0:
+                data = self._ag_obj_constraint_params[arm]
+                obj = og.sim.scene.object_registry("prim_path", data["ag_obj_prim_path"])
+                link = obj.links[data["ag_link_prim_path"].split("/")[-1]]
+                self._ag_data[arm] = (obj, link)
+                self._ag_obj_in_hand[arm] = obj
```
ChengshuLi, line 1274: Is this line needed? I think it might be handled by `self._establish_grasp` internally.

omnigibson/robots/tiago.py
```diff
@@ -576,7 +576,13 @@ def set_position_orientation(self, position=None, orientation=None):
             position = current_position
         if orientation is None:
             orientation = current_orientation
-
+        
+        # Position and orientation are lists when restoring scene from json. Cast to np.array
+        if isinstance(position, list):
+            position = np.array(position)
+        if isinstance(orientation, list):
+            orientation = np.array(orientation)
```
ChengshuLi, line 584: hmm I am not entirely sure why this would be happening. Either way, let's not fix this problem in `tiago.py` because this bug seems to be more general than the Tiago robot.

omnigibson/robots/manipulation_robot.py
misoshiruseijin: Thank you for pointing this out! The `self._establish_grasp` function sets `self._ag_obj_constraint_params` and `self._ag_obj_in_hand`, so I will remove those. `self._ag_obj_in_hand` does need to be set here since nothing else is updating it when the state is loaded. How does this look?
```
        for arm in state["ag_obj_constraint_params"].keys():
            if len(state["ag_obj_constraint_params"][arm]) > 0:
                data = state["ag_obj_constraint_params"][arm]
                obj = og.sim.scene.object_registry("prim_path", data["ag_obj_prim_path"])
                link = obj.links[data["ag_link_prim_path"].split("/")[-1]]
                self._ag_data[arm] = (obj, link)
                self._establish_grasp(arm=arm, ag_data=self._ag_data[arm], contact_pos=data["contact_pos"])
```

omnigibson/robots/tiago.py
misoshiruseijin: Got it. I will revert these changes.

omnigibson/robots/manipulation_robot.py
ChengshuLi: looks great!

misoshiruseijin:
I have made the suggested changes and pushed. Thank you!

omnigibson/robots/tiago.py
```diff
@@ -572,11 +572,7 @@ def get_position_orientation(self):
 
     def set_position_orientation(self, position=None, orientation=None):
         current_position, current_orientation = self.get_position_orientation()
-        if position is None:
-            position = current_position
-        if orientation is None:
-            orientation = current_orientation
```
ChengshuLi, line 578: we probably shouldn't remove these lines?

ChengshuLi (approved):
@cremebrule Would you mind taking a quick look at this PR as well? LGTM.

omnigibson/robots/tiago.py
misoshiruseijin: Sorry, this was an accident. I added these lines back in and re-pushed.

ChengshuLi:
@cremebrule Could you take brief look at this PR and merge it if you think it looks good? I already reviewed it.

cgokmen:
I'm going to merge this now to avoid conflicts with the other PR which also does stuff with this.

PR #280, opened 2023-09-25
Add collision detection for object in hand

sujaygarlanka:


cgokmen:
Whats the plan here? @sujaygarlanka

PR #293, opened 2023-10-18
Update particle utils with support for the 2023 version

cgokmen:
First part of the changes are simply copying over the new version of this class. Second part is adding a 2022/2023 difference since the utils object changes meaningfully, and switching between the two as needed.


cremebrule (approved):
lgtm!

PR #300, opened 2023-10-23
Add an option to use external camera as additional modalities in env obs

mj-hwang:


omnigibson/envs/env_base.py
```diff
@@ -80,6 +93,21 @@ def __init__(
...
 
+        # Set the initial camera if required.
+        self.use_external_camera = use_external_camera
+        if use_external_camera:
+            self.camera = VisionSensor(
+                prim_path="/World/viewer_camera",
+                name="camera",
+                modalities=external_camera_modalities, 
+            )
+            self.camera.initialize()
```
cremebrule, line 104: I think the camera has to be loaded first, right (`cam.load()`)? (i.e.: the prim needs to get created) I think this is why it didn't work when you tried to create a separate camera.

cremebrule (commented):
This is awesome! I think this is a super useful feature. It'd be nice to actually make this a full-fledge feature. I'm imagining support for any number of external cameras. I think a good way to do this could possibly be:

- Having the kwargs take in Iterables. Default could be an empty tuple or list, so no extra ones are used
- Keeping track of the external cameras internally with a property. e.g.: Write the created VisionSensor prims to `self._external_cameras` and expose them via a property `def external_cameras` or something like that
- Automatically appending each cam's obs to the outputted obs dict like you already do

What do you think?

mj-hwang:
Thank you for great feedback Josiah! I fixed and added those features you suggested; I added support for multiple camera with kwargs and create a class property for them.

omnigibson/envs/env_base.py
```diff
@@ -30,6 +31,8 @@ def __init__(
         automatic_reset=False,
         flatten_action_space=False,
         flatten_obs_space=False,
+        external_camera_kwargs=[],
+        external_camera_poses={},
```
cremebrule, line 35: maybe make this also a list so we can directly read out at the same time with external camera kwargs?

omnigibson/envs/env_base.py
```diff
@@ -80,6 +92,21 @@ def __init__(
...
 
+        # Set up external cameras
+        self._external_cameras = {}
+        for ec_kwargs in external_camera_kwargs:
+            camera = VisionSensor(**ec_kwargs)
+            camera.load()
+            camera.initialize()
+
+            camera_name = ec_kwargs["name"]
+            if camera_name in external_camera_poses:
```
cremebrule, line 103: maybe let's force a camera pose to be specified per camera, and assume the number of cam kwargs specified correspond to the number poses specified too.

So then we could do `for ec_kwargs, ec_pose in zip(external_camera_kwargs, external_camera_poses):`

omnigibson/envs/env_base.py
```diff
@@ -42,6 +45,15 @@ def __init__(
             automatic_reset (bool): whether to automatic reset after an episode finishes
             flatten_action_space (bool): whether to flatten the action space as a sinle 1D-array
             flatten_obs_space (bool): whether the observation space should be flattened when generated
+            external_camera_kwargs (list or tuple of dict): List of keyword-specific arguments for external cameras.
```
cremebrule, line 48: can we add support for this directly in the env config file? I believe @cgokmen requested this feature. Perhaps you could comment further?

omnigibson/envs/env_base.py
cgokmen: in fact can we just combine them into a single list or dict that has the pose & kwargs for each camera?

cremebrule:
Closing as this feature has already been implemented.

PR #309, opened 2023-10-31
Created auto-comment.yml

Killer2OP:


Killer2OP:
@cgokmen Review

cgokmen:
Hi, we do not require such a feature currently and do not want to expose our self-hosted runners to the myriad of custom actions used here. Also, we have recently been receiving a lot of these suspicious tiny contributions, so we are preemptively rejecting all of them. Thanks.

PR #323, opened 2023-11-20
Improve Behavior Task Flexibility

cremebrule:


omnigibson/utils/bddl_utils.py
```diff
@@ -521,6 +522,12 @@ def _build_sampling_order(self):
         # Sanity check kinematic objects -- any non-system must be kinematically sampled
         remaining_kinematic_entities = nonparticle_entities - unsampleable_obj_instances - \
             self._inroom_object_instances - set.union(*(self._object_sampling_orders["kinematic"] + [set()]))
+
+        # Possibly remove the agent entity if we're in an empty scene -- i.e.: no kinematic sampling needed for the
+        # agent
+        if self._scene_model is None:
+            remaining_kinematic_entities -= {"agent.n.01_1"}
```
ChengshuLi, line 529: hmmm seems kind of hacky.. but I don't know a better way either..

ChengshuLi (approved):
LGTM!

omnigibson/utils/bddl_utils.py
cremebrule: agreed )):

The most ideal way is to make the global ground plane a proper PrimitiveObject, but the problem is the last time I tried this physics became unstable )):

PR #327, opened 2023-11-28
Feat/control updates

cremebrule:


omnigibson/objects/controllable_object.py
```diff
@@ -350,8 +350,8 @@ def _actions_to_control(self, action):
             u_vec[idx] = ctrl["value"]
             u_type_vec[idx] = ctrl["type"]
 
-        # Return control
-        return u_vec, u_type_vec
+        # Deploy control signals
+        self.deploy_control(control=u_vec, control_type=u_type_vec, indices=None, normalized=False)
```
wensi-ai, line 354: I noticed that this is the only place where `deploy_control` gets called, should we fix `indices` and `normalized` since they are not used anyways?

wensi-ai (approved):
lgtm! Just tested on `robot_control_examples` and it worked.

PR #328, opened 2023-11-30
Feat/external sensors

cremebrule:


omnigibson/envs/env_base.py
```diff
@@ -280,6 +268,33 @@ def _load_objects(self):
...
+        Load any additional custom external sensors into the scene
+        """
+        assert og.sim.is_stopped(), "Simulator must be stopped before loading external sensors!"
+        sensors_config = self.env_config["external_sensors"]
+        if sensors_config is not None:
+            self._external_sensors = dict()
+            for i, sensor_config in enumerate(sensors_config):
+                # Add a name for the object if necessary
+                if "name" not in sensor_config:
+                    sensor_config["name"] = f"external_sensor{i}"
```
cgokmen, line 282: maybe an extra _ after sensor?

omnigibson/prims/entity_prim.py
```diff
@@ -806,6 +806,10 @@ def update_handles(self):
                 joint.initialize()
             joint.update_handles()
 
+        # Update the physics view if we're articulated
+        if self.articulated:
+            self._physics_view = og.sim.physics_sim_view.create_articulation_view(self.articulation_root_path)
+
```
cgokmen, line 812: Seems unrelated - remove?

omnigibson/simulator.py
```diff
@@ -495,6 +495,14 @@ def play(self):
             with suppress_omni_log(channels=channels):
                 super().play()
 
+            # If we're stopped, take a physics step and update the physics sim view. This must happen BEFORE the
```
cgokmen, line 498: Curious, why is this necessary to do here instead of where it was before?

cgokmen (approved):

omnigibson/configs/default_cfg.yaml
```diff
@@ -1,4 +1,7 @@
-env: {}
+env:
+  action_timestep: 0.01666              # (float): environment executes action per action_timestep second
+  physics_timestep: 0.01666             # (float): physics timestep for physx
+  device: null                          # (None or str): specifies the device to be used if running on the gpu with torch backend
```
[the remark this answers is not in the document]
cremebrule: It already does (see L6-7) -- or are you saying that it shouldn't?

omnigibson/prims/entity_prim.py
cremebrule: Oh sorry, yeah it's unrelated but it's a necessary hotfix. Mass matrix stuff breaks without this line

omnigibson/simulator.py
cremebrule: It's weird -- for whatever reason calling `self._physics_sim_view = omni.physics.tensors.create_simulation_view(self.backend)` causes all the current per-object physics views to break. So they need to be refreshed **after** the global physics sim view is created.

omnigibson/configs/fetch_behavior.yaml
```diff
@@ -1,5 +1,21 @@
 env:
+  action_timestep: 0.01666              # (float): environment executes action per action_timestep second
+  physics_timestep: 0.01666             # (float): physics timestep for physx
+  device: null                          # (None or str): specifies the device to be used if running on the gpu with torch backend
+  automatic_reset: false                # (bool): whether to automatic reset after an episode finishes
+  flatten_action_space: false           # (bool): whether to flatten the action space as a sinle 1D-array
+  flatten_obs_space: false              # (bool): whether the observation space should be flattened when generated
+  use_external_obs: false               # (bool): Whether to use external observations or not
   initial_pos_z_offset: 0.1
+  external_sensors:                     # (None or list): If specified, list of sensor configurations for external sensors to add. Should specify sensor "type" and any additional kwargs to instantiate the sensor. Each entry should be the kwargs passed to @create_sensor, in addition to position, orientation
```
ChengshuLi, line 10: external sensor as in not binded to any robot?

omnigibson/envs/env_base.py
```diff
@@ -280,6 +268,33 @@ def _load_objects(self):
...
+                # Determine prim path if not specified
+                if "prim_path" not in sensor_config:
+                    sensor_config["prim_path"] = f"/World/{sensor_config['name']}"
+                # Pop the desired position and orientation
+                local_position, local_orientation = sensor_config.pop("local_position", None), sensor_config.pop("local_orientation", None)
```
ChengshuLi, line 287: Should we assert local position and local orientation are not None? Or is it okay for them to be None?

omnigibson/envs/env_base.py
```diff
@@ -290,6 +305,18 @@ def _load_observation_space(self):
...
+
+        # Also load any external sensors
+        if self._external_sensors is not None:
+            external_obs_space = dict()
+            for sensor_name, sensor in self._external_sensors.items():
+                # Load the sensor observation space
+                external_obs_space[sensor_name] = sensor.load_observation_space()
+                # sensor_obs_space = sensor.load_observation_space()
+                # for obs_modality, obs_modality_space in sensor_obs_space.items():
+                #     external_obs_space[f"{sensor_name}_{obs_modality}"] = obs_modality_space
```
ChengshuLi, line 317: Why do we not need to iterate through `obs_modality`? Say there are RGB and depth, they should be two different keys, right?

omnigibson/robots/robot_base.py
```diff
@@ -253,9 +253,7 @@ def get_obs(self):
         # and grab their observations, processing them into a flat dict
         obs_dict = dict()
         for sensor_name, sensor in self._sensors.items():
-            sensor_obs = sensor.get_obs()
-            for obs_modality, obs in sensor_obs.items():
-                obs_dict[f"{sensor_name}_{obs_modality}"] = obs
+            obs_dict[sensor_name] = sensor.get_obs()
```
ChengshuLi, line 256: oh is `sensor.get_obs()` going to return a dict, keyed by the modality type?

omnigibson/sensors/vision_sensor.py
```diff
@@ -108,12 +110,16 @@ def __init__(
         load_config=None,
         image_height=128,
         image_width=128,
+        focal_length=17.0,
```
ChengshuLi, line 113: `17.0` seems a bit arbitary. Is there a reason behind this magic number? If so, maybe add a comment?

ChengshuLi (approved):
LGTM, left some minor comments!

omnigibson/configs/fetch_behavior.yaml
cremebrule: correct. actually, they could be implicitly bound to the robot, depending on the prim_path specified, so this `external_sensors` thing is quite flexible

omnigibson/envs/env_base.py
cremebrule: hmmm, good point. Originally i had expected them to be optional, similar to all the other objects / robots being specified. What do you think?

omnigibson/robots/robot_base.py
cremebrule: correct! it's a nested dict which is much more clean now, e.g.:

```
- robot0
    - eyes
        - rgb
        - depth
        - ...
```

vs. the old paradigm:

```
- robot0
    - eyes_rgb
    - eyes_depth
    - ...
```

omnigibson/sensors/vision_sensor.py
cremebrule: Yeah 17 is roughly the human focal length -- it's the same focal length we use for the viewer camera so I just made that the default. added a comment

omnigibson/envs/env_base.py
cremebrule: they're nested now! (see comment below)

omnigibson/envs/env_base.py
cremebrule: keeping None for now

omnigibson/envs/env_base.py
cremebrule: bit of a nit, keeping this consistent with our other naming convention (e.g.: `robot0` vs `robot_0`)

PR #334, opened 2023-12-07
Compute control lazily so that controllers do not need stepping

cgokmen:


ChengshuLi (approved):
@cremebrule Could you take a look at this as well?

PR #336, opened 2023-12-11
Replace dynamic control APIs with Isaac.Core View APIs

cgokmen:


omnigibson/utils/extended_articulation_view.py
```diff
@@ -0,0 +1,234 @@
...
+                    self._backend_utils.expand_dims(indices, 1) if self._backend != "warp" else indices, joint_indices
+                ],
+                device=self._device,
+            )
+            return result
+        else:
+            indices = self._backend_utils.resolve_indices(indices, self.count, self._device)
+            joint_indices = self._backend_utils.resolve_indices(joint_indices, self.num_dof, self._device)
+            max_velocities = np.zeros(shape=(indices.shape[0], joint_indices.shape[0]), dtype="float32")
+            indices = self._backend_utils.to_list(indices)
```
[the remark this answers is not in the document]
cgokmen: This code is identical to the pattern in the original articulationview class (in fact I am currently PRing it into the Isaac codebase too internally). As such I think it's good to leave it as-is so it's clear where it came from and how to remove this later.

omnigibson/utils/extended_articulation_view.py
```diff
@@ -0,0 +1,234 @@
...
+                self._backend_utils.resolve_indices(joint_indices, self.num_dof, self._device)
+            )
+            values = self._backend_utils.to_list(values)
+            articulation_read_idx = 0
+            for i in indices:
+                dof_read_idx = 0
+                for dof_index in joint_indices:
+                    dof_val = values[articulation_read_idx][dof_read_idx]
+                    if dof_types[dof_index] == omni.physics.tensors.DofType.Rotation:
+                        dof_val /= DEG2RAD
```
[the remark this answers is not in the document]
cgokmen: The array is not necessarily a numpy array but possibly a torch or warp tensor (in fact in a later version we'll force it to be a torch tensor here). This works for all three of those.

omnigibson/utils/extended_articulation_view.py
```diff
@@ -0,0 +1,234 @@
...
+import math
+
+from omni.isaac.core.articulations import ArticulationView
+
+DEG2RAD = math.pi / 180.0
+
+class ExtendedArticulationView(ArticulationView):
+    """ArticulationView with some additional functionality implemented."""
+
+    def set_joint_limits(
```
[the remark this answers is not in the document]
cgokmen: Exactly!

omnigibson/utils/extended_articulation_view.py
```diff
@@ -0,0 +1,234 @@
...
+                                                                                 to manipulate. Shape (M,).
+                                                                                 Where M <= size of the encapsulated prims in the view.
+                                                                                 Defaults to None (i.e: all prims in the view).
+            joint_indices (Optional[Union[np.ndarray, List, torch.Tensor, wp.array]], optional): joint indicies to specify which joints
+                                                                                 to manipulate. Shape (K,).
+                                                                                 Where K <= num of dofs.
+                                                                                 Defaults to None (i.e: all dofs).
+        """
+        if not self._is_initialized:
+            carb.log_warn("ArticulationView needs to be initialized.")
```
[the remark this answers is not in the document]
cgokmen: This code is identical to the pattern in the original articulationview class (in fact I am currently PRing it into the Isaac codebase too internally). As such I think it's good to leave it as-is so it's clear where it came from and how to remove this later.

omnigibson/utils/extended_articulation_view.py
```diff
@@ -0,0 +1,234 @@
...
+        joint_indices: Optional[Union[np.ndarray, List, torch.Tensor, wp.array]] = None,
+    ) -> None:
+        """Sets joint limits for articulation joints in the view.
+
+        Args:
+            values (Union[np.ndarray, torch.Tensor, wp.array]): joint limits for articulations in the view. shape (M, K, 2).
+            indices (Optional[Union[np.ndarray, List, torch.Tensor, wp.array]], optional): indicies to specify which prims
+                                                                                 to manipulate. Shape (M,).
+                                                                                 Where M <= size of the encapsulated prims in the view.
+                                                                                 Defaults to None (i.e: all prims in the view).
```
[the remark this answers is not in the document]
cgokmen: This code is identical to the pattern in the original articulationview class (in fact I am currently PRing it into the Isaac codebase too internally). As such I think it's good to leave it as-is so it's clear where it came from and how to remove this later.

omnigibson/prims/entity_prim.py
```diff
@@ -306,8 +274,8 @@ def _update_joint_limits(self):
             else:
                 assert not self.initialized, \
                     "Cannot update joint limits for a non-uniformly scaled object when already initialized."
-                for link_name, link in self.links.items():
-                    if joint.parent_name == link_name:
+                for link in self.links.values():
+                    if joint.body0 == link.prim_path:
```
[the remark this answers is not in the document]
cgokmen: parent_name was the non-fully-qualified name received from dynamic control, this is no longer available in the new API, so that parent/child name API is removed from the OG classes too.

omnigibson/prims/entity_prim.py
```diff
@@ -906,117 +799,28 @@ def get_angular_velocity(self):
...
-                    self._dc is not None and self._dc.is_simulating():
-                self.root_link.set_position_orientation(position, orientation)
-            else:
-                super().set_position_orientation(position=position, orientation=orientation)
-
+        if position is not None:
+            position = position[None, :]
+        if orientation is not None:
+            orientation = orientation[None, :]
+        self._articulation_view.set_world_poses(position, orientation)
```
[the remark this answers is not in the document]
cgokmen: yup

omnigibson/prims/entity_prim.py
```diff
@@ -906,117 +799,28 @@ def get_angular_velocity(self):
...
-            else:
-                super().set_local_pose(translation=translation, orientation=orientation)
-        else:
-            if self._root_handle is not None and self._root_handle != _dynamic_control.INVALID_HANDLE and \
-                    self._dc is not None and self._dc.is_simulating():
-                self._set_local_pose_when_simulating(translation=translation, orientation=orientation)
-            else:
-                super().set_local_pose(translation=translation, orientation=orientation)
+        if translation is not None:
+            translation = translation[None, :]
```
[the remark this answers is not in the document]
cgokmen: sure, why not. done.

omnigibson/prims/rigid_prim.py
```diff
@@ -272,99 +247,41 @@ def set_angular_velocity(self, velocity):
...
+        if orientation is not None:
             assert np.isclose(np.linalg.norm(orientation), 1, atol=1e-3), \
                 f"{self.prim_path} desired orientation {orientation} is not a unit quaternion."
-            pose = _dynamic_control.Transform(position, orientation)
-            self._dc.set_rigid_body_pose(self._handle, pose)
-        else:
-            # Call super method by default
-            super().set_position_orientation(position=position, orientation=orientation)
+            orientation = orientation[None, :]
+        self._rigid_prim_view.set_world_poses(positions=position, orientations=orientation)
```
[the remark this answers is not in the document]
cgokmen: Yup

omnigibson/prims/joint_prim.py
```diff
@@ -224,11 +206,13 @@ def set_control_type(self, control_type, kp=None, kd=None):
...
-                dof_property.stiffness = kp
-                dof_property.damping = kd
-                self._dc.set_dof_properties(dof_handle, dof_property)
+        kps = np.full((1, self._n_dof), kp)
+        kds = np.full((1, self._n_dof), kd)
+        self._articulation_view.set_gains(kps=kps, kds=kds, joint_indices=self.dof_indices, save_to_usd=True)
+
+        # Assert no weirdness happening around the save_to_usd business.
+        kps, kds = self._articulation_view.get_gains(joint_indices=self.dof_indices)
+        assert np.allclose(kps, kp) and np.allclose(kds, kd), "Something went wrong with setting the gains!"
```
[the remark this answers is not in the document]
cgokmen: there is a problem here that the getter/setter for gains has the unique save_to_usd feature that makes the gains update on the USD instead of in fabric. it is unclear if this will always update fabric too, hence the check. however, that feature is not aligned with all the other setters so I've gone and disabled it now so no need for the check

omnigibson/prims/joint_prim.py
```diff
@@ -187,15 +171,13 @@ def update_handles(self):
...
-            if joint_path == self._prim_path:
-                self._handle = joint_handle
-                break
-
+        # It's a bit tricky to get the joint index here. We need to find the first dof at this prim path
+        # first, then get the corresponding joint index from that dof offset.
+        self._joint_dof_offset = list(self._articulation_view.dof_paths[0]).index(self._prim_path)
+        self._joint_idx = list(self._articulation_view._metadata.joint_dof_offsets).index(self._joint_dof_offset)
+        self._joint_name = self._articulation_view._metadata.joint_names[self._joint_idx]
+        self._n_dof = self._articulation_view._metadata.joint_dof_counts[self._joint_idx]
```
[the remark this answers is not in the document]
cgokmen: Done.

omnigibson/robots/tiago.py
```diff
@@ -696,13 +697,9 @@ def arm_workspace_range(self):
...
-        # If the simulator is playing, return the pose of the base_footprint link frame
-        if self._dc is not None and self._dc.is_simulating():
-            return self.base_footprint_link.get_position_orientation()
-
-        # Else, return the pose of the robot frame
-        else:
-            return super().get_position_orientation()
+        # TODO: This is a bad idea. If we need the pose to be here, let's put the base there.
+        # We shouldn't have a different behavior for only this object.
+        return self.base_footprint_link.get_position_orientation()
```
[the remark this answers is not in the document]
cgokmen: Simplified this comment & leaving the other one unresolved for a discussion.

omnigibson/prims/entity_prim.py
```diff
@@ -49,17 +48,13 @@ def __init__(
         load_config=None,
     ):
         # Other values that will be filled in at runtime
-        self._dc = None                         # Dynamics control interface
```
cremebrule, line 52: BLESS -- SCREW DC

omnigibson/prims/entity_prim.py
```diff
@@ -237,51 +225,31 @@ def update_joints(self):
...
 
-        # Handle case separately based on whether the handle is valid (i.e.: whether we are actually articulated or not)
-        if (not self.kinematic_only) and self._handle is not None:
-            root_prim = get_prim_at_path(self._dc.get_rigid_body_path(self._root_handle))
-            n_dof = self._dc.get_articulation_dof_count(self._handle)
+        # Handle case separately based on whether we are actually articulated or not
+        if not self.kinematic_only:
+            self._n_dof = self._articulation_view.num_dof
 
+            # TODO: This still includes fixed joints for objects that also have non-fixed?
```
cremebrule, line 232: what is this TODO saying exactly? sorry i'm a bit confused

omnigibson/prims/entity_prim.py
```diff
@@ -471,6 +418,26 @@ def links(self):
             dict: Dictionary mapping link names (str) to link prims (RigidPrim) owned by this articulation
         """
         return self._links
+    
+    @property
+    def articulation_tree(self):
```
cremebrule, line 423: is this used for the scene graph?

omnigibson/prims/entity_prim.py
```diff
@@ -642,27 +575,15 @@ def set_joint_efforts(self, efforts, indices=None, normalized=False):
...
-        if indices is None:
-            new_efforts = efforts
-        else:
-            new_efforts = dof_states["effort"]
-            new_efforts[indices] = efforts
-
         # Set the DOF states
-        dof_states["effort"] = new_efforts
-        self._dc.set_articulation_dof_states(self._handle, dof_states, _dynamic_control.STATE_EFFORT)
+        self._articulation_view.set_joint_efforts(velocities=efforts, joint_indices=indices)
```
cremebrule, line 586: a couple Qs:

- shouldn't it be `efforts=...`, not `velocities=...`?
- so this can now handle variable-length arrays now, based on `joint_indices=...`?

omnigibson/prims/entity_prim.py
```diff
@@ -768,32 +689,14 @@ def _denormalize_efforts(self, efforts, indices=None):
         """
         return efforts * self.max_joint_efforts if indices is None else efforts * self.max_joint_efforts[indices]
 
-    def _verify_physics_view_is_valid(self):
```
cremebrule, line 771: is this no longer needed? iirc it becomes invalid sometimes after a sim step is taken or so

omnigibson/prims/entity_prim.py
```diff
@@ -906,117 +799,28 @@ def get_angular_velocity(self):
...
-                self._set_local_pose_when_simulating(translation=translation, orientation=orientation)
-            else:
-                super().set_local_pose(translation=translation, orientation=orientation)
-        else:
-            if self._root_handle is not None and self._root_handle != _dynamic_control.INVALID_HANDLE and \
-                    self._dc is not None and self._dc.is_simulating():
-                self._set_local_pose_when_simulating(translation=translation, orientation=orientation)
-            else:
-                super().set_local_pose(translation=translation, orientation=orientation)
+        positions, orientations = self._articulation_view.get_world_poses()
```
cremebrule, line 810: this is awesome!! so much cleaner. has this been verified to work when sim is stopped / playing, and for rigid / cloth objects?

omnigibson/prims/entity_prim.py
```diff
@@ -906,117 +799,28 @@ def get_angular_velocity(self):
...
+        return positions[0], orientations[0]
+
+    def set_local_pose(self, position=None, orientation=None):
```
cremebrule, line 813: this is intentionally overriding the super method, right? should we preserve the syntax (i.e.: use `translation` instead of `position`)?

omnigibson/prims/entity_prim.py
```diff
@@ -1353,7 +1157,8 @@ def wake(self):
         Enable physics for this articulation
         """
         if self.articulated:
-            self._dc.wake_up_articulation(self._handle)
+            prim_id = PhysicsSchemaTools.sdfPathToInt(self.prim_path)
+            og.sim.psi.wake_up(og.sim.stage_id, prim_id)
```
cremebrule, line 1161: does this work? I don't think I've tested this functionality on 2023.1.0

omnigibson/prims/joint_prim.py
```diff
@@ -146,31 +145,16 @@ def _initialize(self):
...
-            self._dof_handles = []
-            self._dof_properties = []
             control_types = []
-            for i in range(self._n_dof):
-                dof_handle = self._dc.get_joint_dof(self._handle, i)
-                dof_props = self._dc.get_dof_properties(dof_handle)
-                self._dof_handles.append(dof_handle)
-                self._dof_properties.append(dof_props)
+            stifnesses, dampings = self._articulation_view.get_gains(joint_indices=self.dof_indices)
+            for i, (kp, kd) in enumerate(zip(stifnesses[0], dampings[0])):
```
cremebrule, line 155: nit: typo, should be `stiffness` (two fs)

omnigibson/prims/joint_prim.py
```diff
@@ -413,8 +372,7 @@ def max_effort(self, force):
         """
         # Only support revolute and prismatic joints for now
         assert self.is_single_dof, "Joint properties only supported for a single DOF currently!"
-        self._dof_properties[0].max_effort = force
-        self._dc.set_dof_properties(self._dof_handles[0], self._dof_properties[0])
+        self._articulation_view.set_max_efforts(np.array([[force]]), joint_indices=self.dof_indices)
```
cremebrule, line 375: nit, can we rename force --> effort?

omnigibson/robots/tiago.py
```diff
@@ -713,8 +710,9 @@ def set_position_orientation(self, position=None, orientation=None):
         assert np.isclose(np.linalg.norm(orientation), 1, atol=1e-3), \
             f"{self.name} desired orientation {orientation} is not a unit quaternion."
 
+        # TODO: Reconsider the need for this. Why can't these behaviors be unified? Does the joint really need to move?
```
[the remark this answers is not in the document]
cremebrule: +1 . i will take a look at this too when debugging tiago

omnigibson/simulator.py
```diff
@@ -806,14 +805,6 @@ def __del__(self):
         Simulator._world_initialized = None
         return
 
-    @property
-    def dc(self):
```
cremebrule, line 810: SEE YA

omnigibson/utils/extended_articulation_view.py
```diff
@@ -0,0 +1,234 @@
...
+            joint_indices (Optional[Union[np.ndarray, List, torch.Tensor, wp.array]], optional): joint indicies to specify which joints
+                                                                                 to manipulate. Shape (K,).
+                                                                                 Where K <= num of dofs.
+                                                                                 Defaults to None (i.e: all dofs).
+        """
+        if not self._is_initialized:
+            carb.log_warn("ArticulationView needs to be initialized.")
+            return
+        if not omni.timeline.get_timeline_interface().is_stopped() and self._physics_view is not None:
```
cremebrule, line 40: rubber stamping the functionality -- i presume you copied most of the logic from other functions and then slightly modified it to work with the desired API for joint limits

cremebrule (commented):
Awesome work!! This is such a big improvement over what we had before.

My main concern is stability -- does this work if we:
- reset the env
- stop / play cycle the sim
- add an object
- remove an object

etc. It's the weird edge cases where physx tends to break a bit.

Other than that, overall LGTM!

omnigibson/prims/entity_prim.py
cgokmen: its used for the AG joint type decision (but sure, why not in the scene graph too!)

omnigibson/prims/entity_prim.py
cgokmen: nvm I fixed this issue later.

omnigibson/prims/entity_prim.py
cgokmen: Stop/play yes. I can check for cloth objects.

omnigibson/prims/entity_prim.py
cgokmen: Testing looks OK. But I'm opening an issue to reconsider the whole sleep thing anyway.

omnigibson/robots/tiago.py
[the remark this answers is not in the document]
cgokmen: resolving for now - this is our next work

omnigibson/prims/entity_prim.py
cgokmen: Yup, fixed. Thanks!

omnigibson/prims/entity_prim.py
cgokmen: We need to dig a bit deeper into this. it only ever gets invalidated in the same cases where we need to re-call update_handles, so if it ever happens it means we need to put more calls to update_handles in places instead of fixing it this way.

omnigibson/objects/object_base.py
```diff
@@ -136,9 +141,37 @@ def remove(self):
...
+        assert self._load_config["kinematic_only"] is None or kinematic_only == self._load_config["kinematic_only"], \
+            f"Kinematic only decision does not match! Got: {kinematic_only}, expected: {self._load_config['kinematic_only']}"
+        
+        # Actually apply the kinematic-only decision
+        self._load_config["kinematic_only"] = kinematic_only
+
         # Run super first
         super()._post_load()
 
+        # If the object is fixed_base but kinematic only is false, create the joint
```
cremebrule, line 166: Just rubber stamping this for now -- I assume you unit tested and verified that the new initialization ordering works?

omnigibson/prims/entity_prim.py
```diff
@@ -49,17 +48,14 @@ def __init__(
...
-        self._dofs_infos = None
-        self._n_dof = None                      # dof with dynamic control
+        self._n_dof = None
         self._links = None
         self._joints = None
         self._materials = None
         self._visual_only = None
-        self._physics_view = None
+        self._articulation_tree = None
+        self._articulation_view_direct = None
```
cremebrule, line 58: What does the "direct" imply?

omnigibson/prims/entity_prim.py
```diff
@@ -169,24 +160,24 @@ def _post_load(self):
...
+    def update_links(self):
         """
         Helper function to refresh owned joints. Useful for synchronizing internal data if
         additional bodies are added manually
-
-        Args:
-            load_config (None or dict): If specified, should contain keyword-mapped values that are relevant for
-            loading each of the link's rigid prims
         """
+        load_config = {
```
cremebrule, line 168: I think the original idea was to make this method more functional in nature, but you're right that this is better in the sense that if a user calls it externally, it's less error-prone

omnigibson/prims/entity_prim.py
```diff
@@ -207,22 +198,32 @@ def update_links(self, load_config=None):
...
 
         # Infer the correct root link name -- this corresponds to whatever link does not have any joint existing
         # in the children joints
-        valid_root_links = list(set(self._links.keys()) - joint_children)
+        valid_root_links = list(set(links_to_create.keys()) - joint_children)
 
-        # TODO: Uncomment safety check here after we figure out how to handle legacy multi-bodied assets like bed with pillow
-        # assert len(valid_root_links) == 1, f"Only a single root link should have been found for this entity prim, " \
-        #                                    f"but found multiple instead: {valid_root_links}"
+        assert len(valid_root_links) == 1, f"Only a single root link should have been found for {self.name}, " \
```
cremebrule, line 210: so just to clarify, we no longer have the bed + pillows multi-body-but-single-object case like in ig2?

omnigibson/prims/entity_prim.py
```diff
@@ -334,6 +320,20 @@ def _update_joint_limits(self):
             joint.lower_limit = joint.lower_limit * scale_along_axis
             joint.upper_limit = joint.upper_limit * scale_along_axis
 
+    @property
+    def _articulation_view(self):
```
cremebrule, line 324: should this property be public, and the `articulation_view_direct` simply be the protected version, i.e.: `_articulation_view`? Or do you think that the user should never access this explicitly externally?

omnigibson/prims/entity_prim.py
```diff
@@ -334,6 +320,20 @@ def _update_joint_limits(self):
...
+        if self._articulation_view_direct is None:
+            return None
+
+        # Validate that the articulation view is initialized and that if physics is running, the
+        # view is valid.
+        if og.sim.is_playing() and self.initialized:
+            assert self._articulation_view_direct.is_physics_handle_valid() and \
```
cremebrule, line 331: have we checked edge cases for this? I recall some stop / playing cycles or adding objects at runtime can cause potentially issues with some physics views becoming invalid, IIRC

omnigibson/prims/joint_prim.py
```diff
@@ -146,31 +141,16 @@ def _initialize(self):
...
 
-        # Initialize dynamic control references if this joint is articulated
+        # Update the joint indices etc.
+        self.update_handles()
+
+        # Get control type
         if self.articulated:
-            self._dc = _dynamic_control.acquire_dynamic_control_interface()
-            # TODO: A bit hacky way to get the joint handle, ideally we'd simply do dc.get_joint(), but this doesn't seem to work as expected?
-            for i in range(self._dc.get_articulation_joint_count(self._art)):
```
cremebrule, line 153: I guess now we assume all joints have exactly 1 DOF?

omnigibson/prims/prim_base.py
```diff
@@ -128,6 +128,9 @@ def remove(self):
         # Also clear the name so we can reuse this later
         self.remove_names(include_all_owned=True)
 
+        # Make the simulator refresh its handles
+        og.sim.update_handles()
```
cremebrule, line 132: shouldn't this occur from the sim side of things? (e.g.: within the Simulator class)?

omnigibson/prims/rigid_prim.py
```diff
@@ -82,20 +78,20 @@ def __init__(
         )
 
     def _post_load(self):
+        # Create the view
+        self._rigid_prim_view_direct = RigidPrimView(self._prim_path)
+
+        # Set it to be kinematic if necessary
+        if "kinematic_only" in self._load_config and self._load_config["kinematic_only"]:
```
cremebrule, line 85: should we also enforce the inverse in an `else` clause?

omnigibson/prims/rigid_prim.py
```diff
@@ -453,15 +398,11 @@ def mass(self):
...
-            mass = raw_usd_mass
-        elif self.dc_is_accessible:
-            mass = self.rigid_body_properties.mass
-        else:
-            mass = self.volume * self.density
+        mass = self._rigid_prim_view.get_masses()[0]
+
+        # Fallback to analytical computation of volume * density
+        if mass == 0:
+            return self.volume * self.density
```
cremebrule, line 405: is this how omni computes mass? If not, should we enforce consistency by immediately setting the mass value here?

omnigibson/utils/deprecated_utils.py
```diff
@@ -185,3 +195,284 @@ def get_compute_graph(self, selected_paths, create_new_graph=True, created_paths
...
+            indices = self._backend_utils.to_list(indices)
+            for i in indices:
+                if self._physx_rigid_body_apis[i] is None:
+                    if self._prims[i].HasAPI(PhysxSchema.PhysxRigidBodyAPI):
+                        rigid_api = PhysxSchema.PhysxRigidBodyAPI(self._prims[i])
+                    else:
+                        rigid_api = PhysxSchema.PhysxRigidBodyAPI.Apply(self._prims[i])
+                    self._physx_rigid_body_apis[i] = rigid_api
+                self._physx_rigid_body_apis[i].GetDisableGravityAttr().Set(True)
+            return
```
cremebrule, line 478: nit: extra line needed here

omnigibson/robots/manipulation_robot.py
```diff
@@ -1055,21 +1049,13 @@ def _get_assisted_grasp_joint_type(self, ag_obj, ag_link):
...
-            while joint_handle != 0:
-                # If this joint type is not fixed, we've encountered a valid moving joint
-                # So we create a spherical joint rather than fixed joint
-                if self._dc.get_joint_type(joint_handle) != JointType.JOINT_FIXED:
-                    joint_type = "SphericalJoint"
-                    break
-                # Grab the parent link and its parent joint for the link
-                link_handle = self._dc.get_joint_parent_body(joint_handle)
-                joint_handle = self._dc.get_rigid_body_parent_joint(link_handle)
+        for edge in nx.edge_dfs(ag_obj.articulation_tree, ag_link.body_name, orientation="reverse"):
```
cremebrule, line 1054: I'm not too familiar with nx / graphs -- so when a joint gets created during AG, I'm assuming the articulation tree is automatically updated so that when it's queried here a valid joint edge should be found? Is this an efficient computation?

cremebrule (approved):
overall LGTM!! Just some minor clarifying questions overall, once addressed, feel free to merge

omnigibson/prims/rigid_prim.py
ChengshuLi: I believe so, see https://docs.omniverse.nvidia.com/extensions/latest/ext_physics/rigid-bodies.html#mass-properties

omnigibson/prims/entity_prim.py
```diff
@@ -237,52 +238,37 @@ def update_joints(self):
...
 
-        # Handle case separately based on whether the handle is valid (i.e.: whether we are actually articulated or not)
-        if (not self.kinematic_only) and self._handle is not None:
-            root_prim = get_prim_at_path(self._dc.get_rigid_body_path(self._root_handle))
-            n_dof = self._dc.get_articulation_dof_count(self._handle)
+        # Handle case separately based on whether we are actually articulated or not
+        if self._articulation_view and not self.kinematic_only:
+            self._n_dof = self._articulation_view.num_dof
 
+            # TODO: This still includes fixed joints for objects that also have non-fixed?
```
ChengshuLi, line 245: the comment is still here. If you fixed the issue, maybe remove this comment?

omnigibson/prims/entity_prim.py
cgokmen: We do not, correct.

ChengshuLi (approved):

omnigibson/robots/manipulation_robot.py
cgokmen: This does not include the AG joint - it's the joints the object has at load time. It is an efficient computation because O(V+E) is on average ~1.2 or something I would guess.

omnigibson/prims/prim_base.py
cgokmen: Ideally, yes. Unfortunately right now the removal routines don't remove objects through the og.sim.remove_object interface but instead just by calling .remove() on the individual object. Agreed we should fix this. I filed #350 to track this, but out of scope here.

omnigibson/objects/object_base.py
cgokmen: Yes, it works on all unit tests and some examples I tried.

omnigibson/prims/entity_prim.py
cgokmen: I don't think the user should ever have explicit access to this, because it skips things like BoundingBoxAPI.clear() if you just use this. That said, the checks here are only important temporarily - I think after we test this for a bit we can safely remove this accessor and just access the field directly.

omnigibson/prims/entity_prim.py
cgokmen: Essentially a double-underscore for a really-private member :)

omnigibson/prims/entity_prim.py
cgokmen: Yes, I have checked those - I'm leaving the check in for now but I believe it's all covered (notice the new calls to update_handles I have added)

omnigibson/prims/entity_prim.py
cgokmen: Yes, thanks.

omnigibson/prims/rigid_prim.py
cgokmen: Good call. Done

omnigibson/prims/rigid_prim.py
cgokmen: Agreed that we should do something about this, but let's do it later. Filed #351

cgokmen:
Many thanks for the reviews. Merging.

PR #339, opened 2023-12-12
Fix some bugs in cloth generation

cgokmen:
This caps the max particle count to something we'll need to upper bound by later (due to GPU stuff), and also fixes an issue that causes each vertex to appear 6 times in the final mesh.

### DIFF
--- a/omnigibson/systems/micro_particle_system.py
+++ b/omnigibson/systems/micro_particle_system.py
@@ -44,7 +44,7 @@ m = create_module_macros(module_path=__file__)
 
 # TODO: Tune these default values!
 # TODO (eric): figure out whether one offset can fit all
-m.MAX_CLOTH_PARTICLES = 20000
+m.MAX_CLOTH_PARTICLES = 20000  # Comes from a limitation in physx - do not increase
 m.CLOTH_PARTICLE_CONTACT_OFFSET = 0.0075
 m.CLOTH_REMESHING_ERROR_THRESHOLD = 0.05
 m.CLOTH_STRETCH_STIFFNESS = 10000.0
@@ -1578,11 +1578,11 @@ class Cloth(MicroParticleSystem):
                 cm = ms.current_mesh()
                 if cm.vertex_number() > m.MAX_CLOTH_PARTICLES:
                     # We have too many vertices, so we will re-mesh again
-                    particle_distance *= 1.47  # halve the number of vertices
+                    particle_distance *= np.sqrt(2)  # halve the number of vertices
                 else:
                     break
             else:
-                raise ValueError("Could not remesh with less than MAX_CLOTH_PARTICLES vertices!")
+                raise ValueError(f"Could not remesh with less than MAX_CLOTH_PARTICLES ({m.MAX_CLOTH_PARTICLES}) vertices!")
 
             # Re-write data to @mesh_prim
             new_face_vertex_ids = cm.face_matrix().flatten()
--- a/omnigibson/systems/micro_particle_system.py
+++ b/omnigibson/systems/micro_particle_system.py
@@ -1,3 +1,4 @@
+import uuid
 import omnigibson as og
 from omnigibson.macros import gm, create_module_macros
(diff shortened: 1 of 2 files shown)


omnigibson/systems/micro_particle_system.py
```diff
@@ -43,6 +44,7 @@
 
 # TODO: Tune these default values!
 # TODO (eric): figure out whether one offset can fit all
+m.MAX_CLOTH_PARTICLES = 20000
```
ChengshuLi, line 47: is there a link to their doc that specify this magic number?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1540,44 +1542,58 @@ def clothify_mesh_prim(cls, mesh_prim, remesh=True, particle_distance=None):
...
+
+                avg_edge_percentage_mismatch = 1.0
+                # Loop re-meshing until average edge percentage is within error threshold or we reach the max number of tries
+                for _ in range(5):
+                    if avg_edge_percentage_mismatch <= m.CLOTH_REMESHING_ERROR_THRESHOLD:
+                        break
+
+                    ms.meshing_isotropic_explicit_remeshing(iterations=5, targetlen=pymeshlab.AbsoluteValue(particle_distance))
+                    avg_edge_percentage_mismatch = abs(1.0 - particle_distance / ms.get_geometric_measures()["avg_edge_length"])
+                else:
```
ChengshuLi, line 1573: all the else statements in this function seem weird, can you double check? the logic is a bit hard to follow.

omnigibson/systems/micro_particle_system.py
cgokmen: The actual magic number is something around 22600 or something. It shows up in an error when you exceed it with GPU pipeline on, docs search doesnt point at anything. I left a bit of margin too.

omnigibson/systems/micro_particle_system.py
cgokmen: just checked - the else statements run when the loop isnt broken out of via a `break` statement, which we do in the success cases. So in the failure cases we go into the elses.

cgokmen:
I have tested this and confirmed the cloth demos still work.

omnigibson/systems/micro_particle_system.py
```diff
@@ -1540,44 +1542,58 @@ def clothify_mesh_prim(cls, mesh_prim, remesh=True, particle_distance=None):
...
                     # Terminate anyways, but don't fail
                     log.warn("The generated cloth may not have evenly distributed particles.")
+
+                # Check if we have too many vertices
+                cm = ms.current_mesh()
+                if cm.vertex_number() > m.MAX_CLOTH_PARTICLES:
+                    # We have too many vertices, so we will re-mesh again
+                    particle_distance *= 1.47  # halve the number of vertices
```
cremebrule, line 1581: what does 1.47 signify?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1540,44 +1542,58 @@ def clothify_mesh_prim(cls, mesh_prim, remesh=True, particle_distance=None):
...
+                else:
                     break
+            else:
+                raise ValueError("Could not remesh with less than MAX_CLOTH_PARTICLES vertices!")
```
cremebrule, line 1585: maybe explicitly include the number here: `MAX_CLOTH_PARTICLES={m.MAX_CLOTH_PARTICLES}`

cremebrule (approved):
overall LGTM, a couple minor comments!

omnigibson/systems/micro_particle_system.py
cgokmen: Bad attempt at approximating sqrt(2). Fixed.

omnigibson/systems/micro_particle_system.py
cgokmen: Done.

PR #352, opened 2023-12-22
Fix AG bug when no candidates exist

cgokmen:


cgokmen:
This is harmless and a bad regression so I am merging it.

wensi-ai (approved):
lgtm

PR #353, opened 2023-12-22
bug fix

wensi-ai:


omnigibson/examples/robots/robot_control_example.py
```diff
@@ -120,8 +120,9 @@ def main(random_selection=False, headless=False, short_exec=False):
         orientation=np.array([0.56829048, 0.09569975, 0.13571846, 0.80589577]),
     )
 
-    # Reset environment
+    # Reset environment and robot
     env.reset()
+    robot.reset()
```
wensi-ai, line 125: This is needed since a bunch of og.sim.step() has been taken during load

omnigibson/scenes/scene_base.py
```diff
@@ -472,7 +472,7 @@ def reset(self):
         # Reset the states of all objects (including robots), including (non-)kinematic states and internal variables.
         assert self._initial_state is not None
         self.load_state(self._initial_state)
-        og.sim.step()
+        og.sim.step_physics()
```
wensi-ai, line 475: I don't think we need to render here, plus if physics dt is much smaller then action_dt it could bring down performance

setup.py
```diff
@@ -22,7 +22,7 @@
     packages=find_packages(),
     install_requires=[
         "gym>=0.26",
-        "numpy>=1.20.0",
+        "numpy~=1.23",
```
wensi-ai, line 25: compatibility for windows (c.f. @cgokmen)

omnigibson/objects/controllable_object.py
```diff
@@ -396,11 +395,13 @@ def deploy_control(self, control, control_type, indices=None, normalized=False):
...
         # Standardize normalized input
         n_indices = len(indices)
-        normalized = normalized if isinstance(normalized, Iterable) else [normalized] * n_indices
 
         # Loop through controls and deploy
         # We have to use delicate logic to account for the edge cases where a single joint may contain > 1 DOF
         # (e.g.: spherical joint)
+        pos_vec, pos_idxs, using_pos = [], [], False
+        vel_vec, vel_idxs, using_vel = [], [], False
+        eff_vec, eff_idxs, using_eff = [], [], False
```
wensi-ai, line 404: These change is needed for osc controllers to work (c.f. @cremebrule) because if we set the joint target sequentially the previous changes will be overwritten

omnigibson/prims/rigid_prim.py
```diff
@@ -89,6 +89,14 @@ def _post_load(self):
...
         super()._post_load()
 
+        # Apply rigid body and mass APIs
+        self._rigid_api = UsdPhysics.RigidBodyAPI(self._prim) if self._prim.HasAPI(UsdPhysics.RigidBodyAPI) else \
+            UsdPhysics.RigidBodyAPI.Apply(self._prim)
+        self._physx_rigid_api = PhysxSchema.PhysxRigidBodyAPI(self._prim) if \
+            self._prim.HasAPI(PhysxSchema.PhysxRigidBodyAPI) else PhysxSchema.PhysxRigidBodyAPI.Apply(self._prim)
+        self._mass_api = UsdPhysics.MassAPI(self._prim) if self._prim.HasAPI(UsdPhysics.MassAPI) else \
+            UsdPhysics.MassAPI.Apply(self._prim)
+
```
wensi-ai, line 99: This is needed to get the PrimitiveObject loading work (c.f. @ChengshuLi )

omnigibson/objects/controllable_object.py
```diff
@@ -428,21 +429,36 @@ def deploy_control(self, control, control_type, indices=None, normalized=False):
...
             # Finally, increment the current index based on how many DOFs were just controlled
             cur_indices_idx += joint_dof
+            
+        # set the targets for joints
+        if using_pos:
+            self.set_joint_positions(positions=np.array(pos_vec), indices=np.array(pos_idxs), drive=True, normalized=normalized)
+        if using_vel:
+            self.set_joint_velocities(velocities=np.array(vel_vec), indices=np.array(vel_idxs), drive=True, normalized=normalized)
+        if using_eff:
+            self.set_joint_efforts(efforts=np.array(eff_vec), indices=np.array(eff_idxs), normalized=normalized)
```
cgokmen, line 461: From what I can tell, this doesn't actually do much other than possibly a minor speedup?

omnigibson/prims/joint_prim.py
```diff
@@ -412,7 +412,7 @@ def damping(self):
         """
         # Only support revolute and prismatic joints for now
         assert self.is_single_dof, "Joint properties only supported for a single DOF currently!"
-        _, dampings = self._articulation_view.get_gains(joint_indices=self.dof_indices)[0]
+        dampings = self._articulation_view.get_gains(joint_indices=self.dof_indices)[1]
```
cgokmen, line 415: Ayy sorry for this. Thanks for the fix.

setup.py
cgokmen: Thanks. I hate to be a PITA here but can you freeze a ~= version for all of the dependencies here?

omnigibson/prims/rigid_prim.py
cgokmen: RigidPrimView already supposedly adds these so I don't know why this is happening, but if we need to keep it please go ahead and convert it into the below format:

```
if not self._prim_HasAPI(...):
   ...API.Apply(self._prim)
```

etc. (e.g. don't actually save those APIs as fields. I don't think we need them)

omnigibson/objects/controllable_object.py
wensi-ai: The change is made to get the osc working. Basically it seems like if we set the joint effort individually when we set the latter ones the previous values will get overwritten

setup.py
```diff
@@ -21,28 +21,28 @@
...
-        "pymeshlab",
-        "click",
-        "aenum",
+        "opencv-python~=4.8",
+        "nest_asyncio~=1.5",
+        "imageio~=2.22",
+        "imageio-ffmpeg~=0.4",
+        "termcolor~=2.4",
+        "progressbar~=2.5",
+        "pymeshlab~=2023.12",
```
cgokmen, line 43: I dont think this version works on Windows? Or have you fixed that? Otherwise maybe we can try the latest 2022

omnigibson/objects/controllable_object.py
cgokmen: That doesn't sound right. Please file an issue before you merge.

cgokmen (approved):

setup.py
wensi-ai: Just tested on windows and OG seems working fine

omnigibson/objects/controllable_object.py
wensi-ai: issue filed

Issue #350, opened 2023-12-21
Make removal use sim-based interface instead of prim-based


cgokmen:
Currently during object removal (e.g. w/ particles) we call .remove on the object directly instead of calling sim.remove_object. However, all handles need to be updated, so the obj remover needs to call the sim method. We should just use the sim version instead.

PR #467, opened 2024-01-09
Remove prim dependency on sim update_handles

hang-yin:
#350

### DIFF
--- a/omnigibson/objects/object_base.py
+++ b/omnigibson/objects/object_base.py
@@ -134,6 +134,9 @@ class BaseObject(EntityPrim, Registerable, metaclass=ABCMeta):
         return prim
 
     def remove(self):
+        """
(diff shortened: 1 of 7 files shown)


cgokmen (changes requested):
LGTM, just one request - let's go through BasePrim.remove and all of its overloaders and in each of their docstrings put a clear notice that says "Do NOT call this function directly to remove a prim - call og.sim.remove_prim(prim) for proper cleanup"

hang-yin:
Please check commit [6bba038](https://github.com/StanfordVL/OmniGibson/pull/467/commits/6bba038011037e44efe6fb98f0287821d647d4fc) for prim removal docstring

cgokmen (approved):
LGTM, thanks

PR #527, opened 2024-01-10
Fix non primitive related tests and update dependency packages

ChengshuLi:


setup.py
```diff
@@ -40,7 +40,7 @@
         "imageio-ffmpeg~=0.4.9",
         "termcolor~=2.4.0",
         "progressbar~=2.5",
-        "pymeshlab~=2022.2",
```
wensi-ai, line 43: I don't think this will works with the unit tests? (cc. @cgokmen )

ChengshuLi: reverted back. could you give me an approval?

cgokmen (approved):
Lgtm

Issue #233, opened 2023-07-27
Pickled traversability graph causes errors. Replace it with A* on the image


cgokmen:
Currently, we get these kinds of errors when a pickled traversability graph exists:

```
...
  File "/home/cremebrule/projects/repo/OmniGibson/omnigibson/scenes/interactive_traversable_scene.py", line 162, in _load
    self._trav_map.load_map(maps_path)
  File "/home/cremebrule/projects/repo/OmniGibson/omnigibson/maps/map_base.py", line 27, in load_map
    self.map_size = self._load_map(*args, **kwargs)
  File "/home/cremebrule/projects/repo/OmniGibson/omnigibson/maps/traversable_map.py", line 111, in _load_map
    self.floor_graph = self.build_trav_graph(map_size, maps_path, floor, trav_map)
  File "/home/cremebrule/projects/repo/OmniGibson/omnigibson/maps/traversable_map.py", line 170, in build_trav_graph
    trav_map[node[0], node[1]] = 255
IndexError: index 84 is out of bounds for axis 1 with size 84
```
The traversability graph doesn't really add much of a functionality besides exposing nx's a* function. We can either hack a quick dummy Graph class that can expose edges/nodes from the image, or simply do A* directly on the image, rather than creating a whole second copy of the traversability map as a much less efficient graph object.

cremebrule:
@ChengshuLi what do you think? I am fine either way. I assume it's not that expensive to re-compute the graph each time we load the scene?

PR #528, opened 2024-01-10
Replace Traversability Graph with in-place A*

hang-yin:
#233


hang-yin:
One thing I was considering was to put generic algorithms like A* and connected component analysis somewhere else instead of in this file. We can have an algorithm/util folder to keep these for future use. Let me know what you all think.

hang-yin:
Here's a list of major changes involved in this PR:

1. Instead of doing A* with nx on a new graph, we directly do A* search on the trav map.
2. Trav map erosion now uses robot radius (aabb extent) instead of a magic number. 
3. For mobile manipulators, we are now keeping track of their "tucked aabb extent" for trav map A* search and random point generation.
4. Unit of default erosion radius is now in meters instead of pixels.

omnigibson/maps/traversable_map.py
```diff
@@ -182,40 +107,64 @@ def n_floors(self):
...
             floor = np.random.randint(0, self.n_floors)
-        trav = self.floor_map[floor]
-        trav_space = np.where(trav == 255)
+        
+        # create a deep copy so that we don't erode the original map
+        trav_map = copy.deepcopy(self.floor_map[floor])
+        
+        # Erode the traversability map to account for the robot's size
+        if robot:
+            if hasattr(robot, 'tucked_aabb_extent'):
```
cremebrule, line 139: maybe we can make this more elegant?

use a single property, e.g.: `base_aabb_extent` (base referring to the physical robot locomotive base, not base class), and by default this is the pre-computed, static, cached aabb extent specified in the robot's frame.

And then for specific robots that has tucking capabilities (Fetch, Tiago), their robot class can extend the method to account for tucking.

This way we:
- only have a singled, unified property
- never have to actually compute `aabb_extent` on the fly
- the extent will be guaranteed to be consistent across calls for all robot classes

What do you think?

omnigibson/maps/traversable_map.py
```diff
@@ -226,30 +175,36 @@ def get_shortest_path(self, floor, source_world, target_world, entire_path=False
...
+        trav_map = copy.deepcopy(self.floor_map[floor])
 
-        if not g.has_node(source_map):
-            nodes = np.array(g.nodes)
-            closest_node = tuple(nodes[np.argmin(np.linalg.norm(nodes - source_map, axis=1))])
-            g.add_edge(closest_node, source_map, weight=T.l2_distance(closest_node, source_map))
-
-        path_map = np.array(nx.astar_path(g, source_map, target_map, heuristic=T.l2_distance))
+        # erode the traversability map to account for the robot's size
+        if robot:
```
cremebrule, line 192: same here -- might also be good to reduce duplicate code and put this block of functionality in another call, e.g.: `erode_trav_map(self, robot)`

omnigibson/maps/traversable_map.py
```diff
@@ -226,30 +175,36 @@ def get_shortest_path(self, floor, source_world, target_world, entire_path=False
...
+            robot_radius_pixel = int(np.ceil(robot_radius / self.map_resolution))
+            trav_map = cv2.erode(trav_map, np.ones((robot_radius_pixel, robot_radius_pixel)))
+        else:
+            # convert default_erosion_radius from meter to pixel
+            erosion_radius_pixel = int(np.ceil(self.default_erosion_radius / self.map_resolution))
+            trav_map = cv2.erode(trav_map, np.ones((erosion_radius_pixel, erosion_radius_pixel)))
 
+        path_map = astar(trav_map, source_map, target_map)
+        if path_map is None:
+            raise ValueError("No path found")
```
cremebrule, line 207: Perhaps a bit more descriptive error message -- e.g.: contextualizing that it was trying to find a traversable path, etc.

omnigibson/utils/motion_planning_utils.py
```diff
@@ -544,3 +545,76 @@ def detect_robot_collision_in_sim(robot, filter_objs=[], ignore_obj_in_hand=True
...
+    """
+    A* search algorithm for finding a path from start to goal on a grid map
+
+    Args:
+        search_map (Array): 2D Grid map to search on
+        start (Array): Start position on the map
+        goal (Array): Goal position on the map
+    
+    Returns:
+        Array of arrays: Array of positions that the robot should navigate to (None if no path is found)
```
cremebrule, line 559: Is this a (N, 3)-shaped array? If so, should specify explicitly here

omnigibson/utils/motion_planning_utils.py
```diff
@@ -544,3 +545,76 @@ def detect_robot_collision_in_sim(robot, filter_objs=[], ignore_obj_in_hand=True
...
     return len(collision_prims) > 0
+
+def astar(search_map, start, goal, eight_connected=True):
+    """
+    A* search algorithm for finding a path from start to goal on a grid map
+
+    Args:
+        search_map (Array): 2D Grid map to search on
+        start (Array): Start position on the map
+        goal (Array): Goal position on the map
```
cremebrule, line 556: what is `eight_connected`?

omnigibson/utils/motion_planning_utils.py
```diff
@@ -544,3 +545,76 @@ def detect_robot_collision_in_sim(robot, filter_objs=[], ignore_obj_in_hand=True
         if col_obj.category in filter_categories:
             collision_prims.remove(col_prim)
     return len(collision_prims) > 0
+
+def astar(search_map, start, goal, eight_connected=True):
```
cremebrule, line 549: this is awesome! one question: what is the slowdown here (running loop in python vs. c++)? There's some other open-source python packages that apparently implement A* in C++ (such as [this one](https://github.com/hjweide/pyastar2d)), so I'm curious if there was a reason for the custom implementation here.

I'm by no means an expert, so I'll defer to you. Also cc'ing @cgokmen if he has any further insight too

cremebrule (commented):
Overall LGTM! Mostly style-related comments, with one question regarding the design decision to implement A* manually in python vs. using an OTS implementation based in C++. Happy to discuss!

omnigibson/maps/traversable_map.py
cgokmen: +1

omnigibson/maps/traversable_map.py
```diff
@@ -182,40 +107,64 @@ def n_floors(self):
...
+        if floor and prev_point and floor != prev_point[0]:
+            raise ValueError("floor and prev_point are not on the same floor")
+        # If nothing is given, sample a random floor and a random point on that floor
+        if floor is None and prev_point is None:
             floor = np.random.randint(0, self.n_floors)
-        trav = self.floor_map[floor]
-        trav_space = np.where(trav == 255)
+        
+        # create a deep copy so that we don't erode the original map
+        trav_map = copy.deepcopy(self.floor_map[floor])
```
cgokmen, line 135: This is equal to self.floor_map[floor].copy() if self.floor_map[floor] is a numpy ndarray. Is it? If so, let's ditch the deepcopy.

omnigibson/utils/motion_planning_utils.py
cgokmen: we aren't slowing this down relative to what's already happening iiuc. Previously this was being done in networkx which concretely is slower than this implementation because it is in python and can't benefit from the domain specific knowledge of the 2d grid.

you are right that there are c++ implementations that could be used directly for this purpose. I am however hesitant to add those as a dependency - they all appear unmaintained with known issues. if this becomes an issue later we can take a look.

omnigibson/utils/motion_planning_utils.py
cgokmen: whether we consider just the sides as neighbors or the diagonals too. let's document here.

omnigibson/maps/traversable_map.py
cgokmen: lets discuss this in person quickly tomorrow. there's no perfect solution here and I'm hesitant to spend more time on this - but I do agree that the current approach is not great either so we should discuss

cgokmen (changes requested):

hang-yin:
Ping for review

omnigibson/maps/traversable_map.py
hang-yin: As suggested by @cremebrule and what we talked about during our meeting, I've updated this with a single unified property `base_aabb_extent`. Please check the recent commits for the updated code.

omnigibson/utils/motion_planning_utils.py
hang-yin: I was hesitant to use any external library for this cuz we would be essentially adding another dependency after removing networkx. I'm slightly inclined to keep it like this for now. If this causes any performance issue, it should be super simple to swap this implementation out. If we can reuse this A* implementation in any other part of our code base, that would also be great (I can try to make this more generic to fit more use cases if needed, e.g. support for higher dimension).

omnigibson/robots/robot_base.py
```diff
@@ -412,6 +414,14 @@ def remove(self):
 
         # Run super
         super().remove()
+    
+    @property
+    def base_aabb_extent(self):
```
ChengshuLi, line 419: I would call this `reset_joint_pos_aabb_extent`. The doc string should say something like "this is the aabb extent of the robot in the robot frame after resetting the joints".

omnigibson/robots/fetch.py
```diff
@@ -271,6 +275,10 @@ def _initialize(self):
             for joint in self.finger_joints[arm]:
                 if joint.joint_type != JointType.JOINT_FIXED:
                     joint.friction = 500
+        
+        # Keep track of tucked chassis footprint
+        self.tuck()
```
ChengshuLi, line 280: After thinking about this twice, I think we should remove the special casing for fetch and tiago. Essentially these two robots allow users to specify the reset_joint_pos to be tucked or untucked. The aabb extent should probably respect that argument accordingly.

cgokmen: @ChengshuLi correct me if I'm wrong, but I think the proposal is to not need to reset/tuck/etx the robot at all, just use the configuration at the end of initialize. This should already be the reset pose.

What happens on a robot whose joints might be saved in the JSON file though? Please make sure load_state doesnt get called before you record this because it will set the joints arbitrarily

ChengshuLi (approved):
LGTM! Great work!!

PR #531, opened 2024-01-11
new way to process the ability params of ParticleModifier

ChengshuLi:
allow for system synsets, and human readable strings for modifier method and condition type

### DIFF
--- a/omnigibson/object_states/particle_modifier.py
+++ b/omnigibson/object_states/particle_modifier.py
@@ -27,7 +27,6 @@ from omnigibson.utils.sampling_utils import sample_cuboid_on_object
 from omni.isaac.core.utils.prims import get_prim_at_path, delete_prim, is_prim_path_valid
 from pxr import PhysicsSchemaTools, UsdGeom, Gf, Sdf
 
-
 # Create settings for this module
 m = create_module_macros(module_path=__file__)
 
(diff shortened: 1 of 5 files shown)


omnigibson/object_states/particle_modifier.py
```diff
@@ -256,6 +255,24 @@ def is_compatible_asset(cls, prim, **kwargs):
...
 
+    @classmethod
+    def postprocess_ability_params(cls, params):
+        """
+        Post-processes ability parameters to ensure the system names (rather than synsets) are used for conditions.
+        """
+        # Import here to avoid circular imports
+        from omnigibson.utils.bddl_utils import get_system_name_by_synset
+
+        for sys in list(params["conditions"].keys()):
```
cremebrule, line 266: shouldn't this be wrapped with `if "conditions" in params`?

omnigibson/objects/stateful_object.py
```diff
@@ -32,16 +34,7 @@
 # Create module logger
 log = create_module_logger(module_name=__name__)
 
-
-# Optionally import bddl for object taxonomy.
```
cremebrule, line 36: we remove this because we force bddl to be an explicit OG dependency now, right?

omnigibson/utils/constants.py
```diff
@@ -50,17 +50,17 @@ class SemanticClass(IntEnum):
 
 
 # Specific methods for applying / removing particles
-class ParticleModifyMethod(IntEnum):
-    ADJACENCY = 0
-    PROJECTION = 1
+class ParticleModifyMethod(str, Enum):
```
cremebrule, line 53: whoaaa what is this python magic??

cremebrule (approved):
lgtm! one minor fix (i think), then feel free to merge!

omnigibson/object_states/particle_modifier.py
ChengshuLi: That was what I did initially, but then I realized conditions is a required positional arg for the particlemodifier class. So that key has to exist?

omnigibson/utils/constants.py
ChengshuLi: You did this magic before and I just copied your code lol.

omnigibson/object_states/particle_modifier.py
cremebrule: gotcha, sg!

omnigibson/objects/stateful_object.py
cremebrule: sg

omnigibson/utils/constants.py
cremebrule: oh LOOL

PR #536, opened 2024-01-12
Fix transition rule tests on CI

ChengshuLi:


cgokmen (approved):
Does this actually fix things?

ChengshuLi:
I believe the tests pass now. Feel free to check once it's done running.

Issue #316, opened 2023-11-06
Slicing demo doesn't work


cgokmen:
reported by Wensi

PR #545, opened 2024-01-16
Add slicer delay to fix slicing demo

hang-yin:
#316 
This PR tries to fix the slicing demo where the knife slices an apple and immediately dices it. 

To fix this, we've introduced a manual delay that temporarily deactivates the slicer after it gets into contact with a sliceable. 

Currently, the delay is set to 7 frames, which is the minimum required for this demo to function correctly. However, this delay is really arbitrarily and may not work in other scenarios. For example, in our demo, the knife is moved by gravity, which is significantly faster than a robot arm manipulating the knife. 

I could really use everyone's opinion on this. Let me know what you think the best way to do this is! Thanks.

### DIFF
--- a/omnigibson/object_states/slicer_active.py
+++ b/omnigibson/object_states/slicer_active.py
@@ -23,6 +23,7 @@ class SlicerActive(AbsoluteObjectState, UpdateStateMixin):
         self._previously_touching_sliceable = False
         self.value = True
         self.delay_counter = 0
+        self.steps_to_wait = max(1, m.REACTIVATION_DELAY / og.sim.get_rendering_dt())
 
     def _get_value(self):
         return self.value
@@ -58,8 +59,7 @@ class SlicerActive(AbsoluteObjectState, UpdateStateMixin):
             # If we are not touching any sliceable objects, we can revert to True.
             if not currently_touching_sliceable:
                 self.delay_counter += 1  # Increment the counter for each step we're not touching a sliceable object
-                steps_to_wait = max(1, m.REACTIVATION_DELAY / og.sim.get_rendering_dt())
-                if self.delay_counter >= steps_to_wait:  # Check if the counter has reached the delay
+                if self.delay_counter >= self.steps_to_wait:  # Check if the counter has reached the delay
                     self.set_value(True)
             else:
                 # If we are touching a sliceable object, reset the counter.
@@ -70,18 +70,21 @@ class SlicerActive(AbsoluteObjectState, UpdateStateMixin):
 
     @property
     def state_size(self):
-        return 2
+        return 3
 
     # For this state, we simply store its value.
     def _dump_state(self):
-        return dict(value=self.value, previously_touching_sliceable=self._previously_touching_sliceable)
+        return dict(value=self.value,
+                    previously_touching_sliceable=self._previously_touching_sliceable,
+                    delay_counter=self.delay_counter)
 
     def _load_state(self, state):
         self.value = state["value"]
         self._previously_touching_sliceable = state["previously_touching_sliceable"]
+        self.delay_counter = state["delay_counter"]
 
     def _serialize(self, state):
-        return np.array([state["value"], state["previously_touching_sliceable"]], dtype=float)
+        return np.array([state["value"], state["previously_touching_sliceable"], state["delay_counter"]], dtype=float)
 
     def _deserialize(self, state):
-        return dict(value=state[0], previously_touching_sliceable=state[1]), 2
+        return dict(value=state[0], previously_touching_sliceable=state[1], delay_counter=state[2]), 3
--- a/omnigibson/object_states/slicer_active.py
+++ b/omnigibson/object_states/slicer_active.py
@@ -23,6 +23,8 @@ class SlicerActive(AbsoluteObjectState, UpdateStateMixin):
 
         self._previously_touching_sliceable = False
         self.value = True
+        self.delay_counter = 0
+        self.delay_steps = 7
 
     def _get_value(self):
         return self.value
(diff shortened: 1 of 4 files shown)


omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,85 @@
+import numpy as np
+from omnigibson.macros import create_module_macros
+from omnigibson.object_states.object_state_base import AbsoluteObjectState
+from omnigibson.object_states.aabb import AABB
```
cgokmen, line 4: Remove

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+# Create settings for this module
+m = create_module_macros(module_path=__file__)
+
+class SlicerActive(AbsoluteObjectState, UpdateStateMixin):
+    @classmethod
+    def get_dependencies(cls):
+        deps = super().get_dependencies()
+        deps.add(AABB)
```
cgokmen, line 17: Remove

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,85 @@
...
+    
+    def _currently_touching_sliceable(self):
+        contact_links = self.obj.states[ContactBodies].get_value()
+        contact_link_prim_paths = {contact_link.prim_path for contact_link in contact_links}
+        for prim_path in contact_link_prim_paths:
+            obj_prim_path, _ = prim_path.rsplit("/", 1)
+            candidate_obj = og.sim.scene.object_registry("prim_path", obj_prim_path, None)
+            if candidate_obj is None:
+                continue
+            if hasattr(candidate_obj, '_abilities') and "sliceable" in candidate_obj._abilities:
```
cgokmen, line 44: Don't access the _ member. Check `if isinstance(candidate_obj, StatefulObject)` and then just read `candidate_obj.abilities` without the underscore

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,85 @@
...
+            self.set_value(False)
+            self.delay_counter = 0  # Reset the counter when we stop touching a sliceable object
+
+        # Are we currently touching any sliceables?
+        currently_touching_sliceable = self._currently_touching_sliceable()
+
+        # If our value is False, we need to consider reverting back.
+        if not self.value:
+            # If we are not touching any sliceable objects, we can revert to True.
+            if not currently_touching_sliceable:
```
cgokmen, line 60: The `else` case of this happens in the next step's line 52, but maybe let's still add it here - if we are touching sliceable, just reset the timer.

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,85 @@
...
+        deps.add(ContactBodies)
+        return deps
+
+    def __init__(self, obj):
+        super(SlicerActive, self).__init__(obj)
+
+        self._previously_touching_sliceable = False
+        self.value = True
+        self.delay_counter = 0
+        self.delay_steps = 7
```
cgokmen, line 27: Don't make this an instance member, make it a macro/constant. Right after line 11 add m.REACTIVATION_DELAY = 7 etc.

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Are we currently touching any sliceables?
+        currently_touching_sliceable = self._currently_touching_sliceable()
+
+        # If our value is False, we need to consider reverting back.
+        if not self.value:
+            # If we are not touching any sliceable objects, we can revert to True.
+            if not currently_touching_sliceable:
+                self.delay_counter += 1  # Increment the counter for each step we're not touching a sliceable object
+                if self.delay_counter >= self.delay_steps:  # Check if the counter has reached the delay
```
cgokmen, line 62: Note that delay_counter gets incremented once every rendering timestep which is a parameter. By default (at 1/60sec) it will happen at 0.1s but this parameter can be much higher. Perhaps it's best to specify the macro as a number of seconds (e.g. 0.1), then compute it by `min(1, m.REACTIVATION_DELAY / og.sim.rendering_dt)` to decide how many steps to wait.

cgokmen: Also let's just reintroduce that cooldown to be like 0.5 sec for now imo. @ChengshuLi FYI

cgokmen (changes requested):

omnigibson/object_states/slicer_active.py
hang-yin: Hi @cgokmen, just making sure that I understand this correctly: 
For example, if we set `REACTIVATION_DELAY` to be 0.1 and `og.sim.get_rendering_dt()` is now 0.016, that would give us about 6 steps/frames to wait. If we then take `min(1, m.REACTIVATION_DELAY / og.sim.get_rendering_dt())`, wouldn't that always give us 1? I guess for the min cap, did you mean something like 10 instead?

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,88 @@
...
+                steps_to_wait = max(1, m.REACTIVATION_DELAY / og.sim.get_rendering_dt())
+                if self.delay_counter >= steps_to_wait:  # Check if the counter has reached the delay
+                    self.set_value(True)
+            else:
+                # If we are touching a sliceable object, reset the counter.
+                self.delay_counter = 0
+
+        # Record if we were touching anything previously
+        self._previously_touching_sliceable = currently_touching_sliceable
+
```
cgokmen, line 70: nit: remove extra line

cgokmen (approved):
fix nit and merge

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,87 @@
...
+    def _load_state(self, state):
+        self.value = state["value"]
+        self._previously_touching_sliceable = state["previously_touching_sliceable"]
+
+    def _serialize(self, state):
+        return np.array([state["value"], state["previously_touching_sliceable"]], dtype=float)
+
+    def _deserialize(self, state):
+        return dict(value=state[0], previously_touching_sliceable=state[1]), 2
```
ChengshuLi, line 87: self.delay_counter should also be part of the state?

omnigibson/object_states/slicer_active.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Are we currently touching any sliceables?
+        currently_touching_sliceable = self._currently_touching_sliceable()
+
+        # If our value is False, we need to consider reverting back.
+        if not self.value:
+            # If we are not touching any sliceable objects, we can revert to True.
+            if not currently_touching_sliceable:
+                self.delay_counter += 1  # Increment the counter for each step we're not touching a sliceable object
+                steps_to_wait = max(1, m.REACTIVATION_DELAY / og.sim.get_rendering_dt())
```
ChengshuLi, line 61: maybe `steps_to_wait = max(1, m.REACTIVATION_DELAY / og.sim.get_rendering_dt())` should be computed once and stored?

ChengshuLi (approved):
LGTM, with two minor comments. Feel free to merge after addressing them.

PR #548, opened 2024-01-17
Fix broken functions in ArticulationView

cgokmen:
This copies 3 more functions verbatim from Isaac's ArticulationView and fixes their implementations. See my issue on NVIDIA developer forums for more details. https://forums.developer.nvidia.com/t/articulationview-bugs-in-joint-position-velocity-effort-setting/279186

### DIFF
--- a/omnigibson/robots/manipulation_robot.py
+++ b/omnigibson/robots/manipulation_robot.py
@@ -916,6 +916,7 @@ class ManipulationRobot(BaseRobot):
         for joint_name, j_val in self._ag_freeze_joint_pos[arm].items():
             joint = self._joints[joint_name]
             joint.set_pos(pos=j_val)
+            joint.set_vel(vel=0.0)
 
     @property
     def robot_arm_descriptor_yamls(self):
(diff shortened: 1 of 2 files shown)


omnigibson/robots/manipulation_robot.py
```diff
@@ -916,6 +916,7 @@ def _freeze_gripper(self, arm="default"):
         for joint_name, j_val in self._ag_freeze_joint_pos[arm].items():
             joint = self._joints[joint_name]
             joint.set_pos(pos=j_val)
+            joint.set_vel(vel=0.0)
```
cremebrule, line 919: why is this needed?

cremebrule (approved):
lgtm, just one comment! feel free to merge afterwards

omnigibson/robots/manipulation_robot.py
cgokmen: This used to be here but was removed recently by Wensi attempting to fix this issue. This solves it so we can put it back. It stops the gripper from moving out.

Issue #121, opened 2023-03-08
Improve ergonomics of importing omnigibson


mjlbach:
One issue with OG is that importing omnigibson immediately creates a simulation context, this is required because importing/subclassing any OG modules usually pulls in omniverse/isaac libraries which are not loaded by the python interpeter until the isaac simulation context is created. 

It would be good to improve the ergonomics of this by either:
1. Making all `omni.*` python imports lazy, this defers actually invoking the import until runtime and is probably the easiest solution
* https://github.com/telekom/lazy-imports
2. modifying pythonpath directly so that it can reference the omniverse runtime modules (this may not work as some are bindings to c)

This is critical as we need to be able to set certain options dynamically at runtime, like `active_gpu`, `physics_gpu` etc. to support parallel training.

PR #551, opened 2024-01-18
Lazy imports

hang-yin:
#121 
This PR implements `lazy import` for all omni.* imports, using this [repo](https://github.com/telekom/lazy-imports). This means that importing `omnigibson` no longer immediately launches the simulator.


cgokmen (approved):
LGTM

PR #562, opened 2024-01-23
Fix/prune logging

cremebrule:


omnigibson/__init__.py
```diff
@@ -52,8 +53,8 @@ def cleanup(*args, **kwargs):
         shutil.rmtree(tempdir)
     except PermissionError:
         log.info("Permission error when removing temp files. Ignoring")
-    from omnigibson.utils.ui_utils import suppress_omni_log
-    log.info(f"{'-' * 10} Shutting Down OmniGibson {'-' * 10}")
+    from omnigibson.simulator import logo_small
+    log.info(f"{'-' * 10} Shutting Down {logo_small()} {'-' * 10}")
```
cgokmen, line 57: I actually changed this explicitly because of an issue with circular imports - also feels weird for the simulator to need to be imported here now. (also the "as launch" part is unused). Could you put the logos in a separate utils file somewhere actually?

omnigibson/simulator.py
```diff
@@ -99,7 +99,22 @@ def _launch_app():
...
         config_kwargs["physics_gpu"] = gpu_id
+
+    # Omni's logging is super annoying and overly verbose, so suppress it by modifying the logging levels
+    if not gm.DEBUG:
+        import sys
+        from numba.core.errors import NumbaPerformanceWarning
+        import warnings
+        sys.argv.append("--/log/level=warning")
+        sys.argv.append("--/log/fileLogLevel=warning")
+        sys.argv.append("--/log/outputStreamLevel=error")
```
cgokmen, line 110: This seems really brittle, appending arguments to our own call after the fact like this :( if these are arguments that are supposed to be passed into the kit app, can we specify them as part of the config_kwargs passed into SimulationApp in 116? Note that this will be a huge surprise / pain in the ass for any script that still read `sys.argv` _after_ loading the simulator.

omnigibson/simulator.py
```diff
@@ -99,7 +99,22 @@ def _launch_app():
...
+        warnings.simplefilter("ignore", category=NumbaPerformanceWarning)
+
+    import omni.log
+    omni_log = omni.log.get_log()
+    omni_log.enabled = False
```
cgokmen, line 115: What does this do, does it disable the whole launch feed? Do we want to do that? Also should this also follow the debug mode stuff?

omnigibson/__init__.py
cremebrule: done

omnigibson/simulator.py
cremebrule: It only prunes the annoying initial logging at the beginning of launch, not the whole thing

ChengshuLi (approved):

PR #566, opened 2024-01-24
New QA script for batch examination

hang-yin:


ChengshuLi:
This code should be ported to ig-pipeline.

PR #578, opened 2024-01-30
Added caching for Kinematic-only object pose

hang-yin:


omnigibson/prims/rigid_prim.py
```diff
@@ -61,6 +61,11 @@ def __init__(
         self._visual_only = None
         self._collision_meshes = None
         self._visual_meshes = None
+        
+        # Caches for kinematic-only objects
+        # This exsists because RigidPrimView uses USD pose read, which is very slow
```
cgokmen, line 66: exsists

omnigibson/prims/rigid_prim.py
```diff
@@ -294,13 +303,25 @@ def set_position_orientation(self, position=None, orientation=None):
...
+        
         pos, ori = self._rigid_prim_view.get_world_poses()
 
         assert np.isclose(np.linalg.norm(ori), 1, atol=1e-3), \
             f"{self.prim_path} orientation {ori} is not a unit quaternion."
-        return pos[0], ori[0][[1, 2, 3, 0]]
+        
+        pos = pos[0]
+        ori = ori[0][[1, 2, 3, 0]]
+        self._kinematic_world_pose_cache = (pos, ori)
```
cgokmen, line 317: Maybe let's not set this when not kinematic to avoid confusion. Or I'm not sure - wdyt @cremebrule

cgokmen (approved):

omnigibson/prims/rigid_prim.py
cremebrule: agreed -- let's nest all of this logic within a `if self.kinematic_only` block if possible

cremebrule (commented):
one comment, otherwise LGTM!

hang-yin:
profile

omnigibson/prims/rigid_prim.py
hang-yin: I can add a `self.kinematic_only:` check before setting the cache but it would be kind of hard to wrap the all of these logic into one if check though. If we were to do that, we will need to duplicate the `get_world_poses()` call or create a separate function just for this.

PR #582, opened 2024-01-31
USDRt for XformPrim get_position_orientation

hang-yin:
**Original problem**: `get_position_orientation` returns local pose for vision sensor when flatcache is on.
**Cause**: USD not updated with flatcache on.
**Fix**: use `USDRt` version of xformprim `get_world_pose` to bypass the USD version `ComputeLocalToWorldTransform`.

**New problem**: `set_position_orientation` doesn't always set prim to the input pose.
**Cause**: 

1. parent-world transform still uses USD version `ComputeLocalToWorldTransform` when flatcache is on.
2. parent scale not taken into account for metalink transform.

**Fix**: 
 1. use `USDRt` version of `get_world_poses` to get parent-world transform.
 2. apply parent scale for parent-world transform.


omnigibson/prims/xform_prim.py
```diff
@@ -151,31 +152,55 @@ def set_position_orientation(self, position=None, orientation=None):
...
         mat.SetTranslation(lazy.pxr.Gf.Vec3d(*position))
 
-        # mat.SetScale(lazy.pxr.Gf.Vec3d(*(self.get_world_scale() / self.scale)))
-        # TODO (eric): understand why this (mat.setScale) works - this works empirically but it's unclear why.
-        mat.SetScale(lazy.pxr.Gf.Vec3d(*(self.scale.astype(np.float64))))
         my_world_transform = np.transpose(mat.GetMatrix())
 
-        parent_world_tf = lazy.pxr.UsdGeom.Xformable(lazy.omni.isaac.core.utils.prims.get_prim_parent(self._prim)).ComputeLocalToWorldTransform(lazy.pxr.Usd.TimeCode.Default())
-        parent_world_transform = np.transpose(parent_world_tf)
+        # When flatcache is on, we need to use USDRt to get the parent's world transform
```
cremebrule, line 170: and when physics is active (i.e.: when sim is playing)

omnigibson/prims/xform_prim.py
```diff
@@ -151,31 +152,55 @@ def set_position_orientation(self, position=None, orientation=None):
...
+            parent_pos, parent_ori = parent_view.get_world_poses()
+            parent_pose = (parent_pos[0], parent_ori[0][[1, 2, 3, 0]])
+
+            # Get the parent scale transform
+            parent_scale = np.array(np.ones(3) if parent_prim.GetAttribute("xformOp:scale").Get() is None else parent_prim.GetAttribute("xformOp:scale").Get())
+            parent_scale_tf = np.eye(4)
+            parent_scale_tf[:3, :3] = np.diag(parent_scale)
 
-        local_transform = np.matmul(np.linalg.inv(parent_world_transform), my_world_transform)
+            parent_world_transform = T.pose2mat(parent_pose) @ parent_scale_tf
```
cremebrule, line 190: can we add a comment here specifying why we multiply by the scale? (ie: clearly document the linear algebra formula we came up with yesterday, and then show the derived solution)

cremebrule (commented):
awesome work!!! a couple small comments, feel free to merge once addressed

omnigibson/prims/xform_prim.py
cremebrule: Hmmm I don't think is entirely true -- T_parent should also be multiplied by the scaling matrix too, right? Maybe it'd be helpful to show the sequence of operations, ie: translation --> then rotation --> then scale, for each nested prim

omnigibson/prims/xform_prim.py
```diff
@@ -187,12 +187,22 @@ def set_position_orientation(self, position=None, orientation=None):
...
+            # |          0            1        |
             parent_world_transform = T.pose2mat(parent_pose) @ parent_scale_tf
         # When flatcache is off, we can simply use USD to get the parent's world transform
         else:
             parent_world_tf = lazy.pxr.UsdGeom.Xformable(lazy.omni.isaac.core.utils.prims.get_prim_parent(self._prim)).ComputeLocalToWorldTransform(lazy.pxr.Usd.TimeCode.Default())
             parent_world_transform = np.transpose(parent_world_tf)
 
+        # Calculate the local transform from parent to current prim
+        # Inverse of parent world transform multiplied by the current world transform
+        # Note: only the parent's scale matters for the local transform
```
cremebrule, line 202: This is great! maybe explicitly specify here that Pose(X) = Translation(X) * Rotation(X) * Scale(X) so it's clear how it's decomposed

cremebrule (commented):
a couple minor comments!

hang-yin:
Ping for review.

omnigibson/prims/xform_prim.py
```diff
@@ -280,6 +272,9 @@ def set_local_pose(self, position=None, orientation=None):
             orientation (None or 4-array): if specified, (x,y,z,w) quaternion orientation in the local frame of the prim
                 (with respect to its parent prim). Default is None, which means left unchanged.
         """
+        if gm.ENABLE_FLATCACHE:
+            xformable_prim = lazy.usdrt.Rt.Xformable(lazy.omni.isaac.core.utils.prims.get_prim_at_path(self.prim_path, fabric=True))
+        # assert xformable_prim.HasWorldXform() == False
```
cgokmen, line 277: Why do we remove this assertion? Reenable if possible, remove if not.

cgokmen (approved):
See comments

omnigibson/prims/entity_prim.py
```diff
@@ -621,6 +621,7 @@ def set_joint_positions(self, positions, indices=None, normalized=False, drive=F
             self._articulation_view.set_joint_position_targets(positions, joint_indices=indices)
         else:
             self._articulation_view.set_joint_positions(positions, joint_indices=indices)
+            PoseAPI.invalidate()
```
ChengshuLi, line 624: nit, let's also call it `clear` to be consistent?

tests/test_robot_states.py
```diff
@@ -0,0 +1,99 @@
...
+    assert np.allclose(sensor_world_ori, sensor_world_ori_gt)
+    
+    # Now, we want to move the robot and check if the sensor pose is updated
+    old_camera_local_pose = vision_sensor.get_local_pose()
+    old_camera_world_pose = vision_sensor.get_position_orientation()
+    robot.set_position_orientation(position=[100, 100, 100])
+    new_camera_local_pose = vision_sensor.get_local_pose()
+    new_camera_world_pose = vision_sensor.get_position_orientation()
+    assert np.allclose(old_camera_local_pose[0], new_camera_local_pose[0])
+    assert not np.allclose(old_camera_world_pose[0], new_camera_world_pose[0])
```
ChengshuLi, line 71: for all the `not np.all_close`, since we know the new pose we set, can we analytically compute the expected value and change them to `np.all_close(new_value)` instead?

omnigibson/utils/usd_utils.py
```diff
@@ -697,10 +697,46 @@ def reset(cls):
         cls.MODIFIED_PRIMS = set()
 
 
+class PoseAPI:
+    DIRTY = True
```
ChengshuLi, line 701: let's call this `VALID` instead, and the following two functions `validate` and `invalidate`

omnigibson/utils/usd_utils.py
```diff
@@ -697,10 +697,46 @@ def reset(cls):
...
+        cls.DIRTY = False
+        
+    @classmethod
+    def _refresh(cls):
+        if og.sim is not None and cls.DIRTY:
+            if og.sim._physx_fabric_interface:
+                og.sim._physx_fabric_interface.update(og.sim.get_physics_dt(), og.sim.current_time)
+            else:
+                og.sim.psi.fetch_results()
+            cls.mark_valid()
```
ChengshuLi, line 718: 1. when will `_physx_fabric_interface` be None?
2. will `og.sim._physx_fabric_interface.update(og.sim.get_physics_dt(), og.sim.current_time)` cause one physics dt to elapse?

omnigibson/simulator.py
```diff
@@ -767,6 +770,7 @@ def step_physics(self):
             """
             Step the physics a single step.
             """
+            PoseAPI.invalidate()
```
ChengshuLi, line 773: on a high-level note, for all these `invalidate` call, why not just call `refresh` directly? 

also, nit, this line should be after `self._physics_context._step(current_time=self.current_time)`?

omnigibson/prims/xform_prim.py
```diff
@@ -300,7 +286,10 @@ def set_local_pose(self, position=None, orientation=None):
             else:
                 rotq = lazy.pxr.Gf.Quatd(*orientation)
             xform_op.Set(rotq)
+        PoseAPI.invalidate()
         BoundingBoxAPI.clear()
+        if gm.ENABLE_FLATCACHE:
+            xformable_prim.SetLocalXformFromUsd()
```
ChengshuLi, line 292: can you add more comments to explain what is going on?

omnigibson/utils/usd_utils.py
```diff
@@ -697,10 +697,46 @@ def reset(cls):
         cls.MODIFIED_PRIMS = set()
 
 
+class PoseAPI:
```
ChengshuLi, line 700: Add docstring to this class, explaining what is this class trying to do

tests/test_robot_states.py
hang-yin: set parent scale for stress test

omnigibson/prims/xform_prim.py
```diff
@@ -153,32 +154,20 @@ def set_position_orientation(self, position=None, orientation=None):
...
-        # TODO (eric): understand why this (mat.setScale) works - this works empirically but it's unclear why.
-        mat.SetScale(lazy.pxr.Gf.Vec3d(*(self.scale.astype(np.float64))))
-        my_world_transform = np.transpose(mat.GetMatrix())
+        my_world_transform = T.pose2mat((position, orientation))
 
-        parent_world_tf = lazy.pxr.UsdGeom.Xformable(lazy.omni.isaac.core.utils.prims.get_prim_parent(self._prim)).ComputeLocalToWorldTransform(lazy.pxr.Usd.TimeCode.Default())
-        parent_world_transform = np.transpose(parent_world_tf)
+        parent_prim = lazy.omni.isaac.core.utils.prims.get_prim_parent(self._prim)
+        parent_path = str(parent_prim.GetPath())
+        parent_world_transform = PoseAPI.get_world_pose_with_scale(parent_path).T
```
cremebrule, line 167: I assume this has been tested?

omnigibson/prims/xform_prim.py
```diff
@@ -300,7 +284,18 @@ def set_local_pose(self, position=None, orientation=None):
...
         BoundingBoxAPI.clear()
+        if gm.ENABLE_FLATCACHE:
+            # If flatcache is on, make sure the USD local pose is synced to the fabric local pose.
+            # Ideally we should call usdrt's set local pose directly, but there is no such API.
+            # The only available API is SetLocalXformFromUsd, so we update USD first, and then sync to fabric.
+            xformable_prim = lazy.usdrt.Rt.Xformable(lazy.omni.isaac.core.utils.prims.get_prim_at_path(self.prim_path, fabric=True))
+            # We also need to clear the fabric's world pose (potentially set by physx) to force fabric to use
+            # the local pose we just set, because world pose takes precendence over local pose.
+            # if xformable_prim.HasWorldXform():
+            assert not xformable_prim.HasWorldXform(), "Fabric's world pose is set for a non-rigid prim which is unexpected. Please report this."
```
cremebrule, line 297: should be explicitly say "xformprim" instead of "non-rigid prim"? Or are there other cases where this applies?

omnigibson/simulator.py
```diff
@@ -684,6 +684,9 @@ def play(self):
                 # Take a render step -- this is needed so that certain (unknown, maybe omni internal state?) is populated
                 # correctly.
                 self.render()
+                
+                # During rendering, the Fabric API is updated, so we can mark it as clean
+                PoseAPI.mark_valid()
```
cremebrule, line 689: if rendering automatically validates PoseAPI, should we bake this directly into the render() call?

cremebrule (commented):
overall lgtm! left some clarifying questions

omnigibson/prims/xform_prim.py
cgokmen: So the gist of it is that for anything that has a physx counterpart (articulation, rigidbody) there will be a worldxform but that's OK because this code will never get called for those objects. It's only the non-physx stuff that this gets called for, and for those there should never be a worldxform or the rest of this code wont work

omnigibson/simulator.py
cgokmen: @hang-yin lets do this by overriding render. I think there might be another place where we do this post-render validation too.

omnigibson/prims/xform_prim.py
```diff
@@ -300,6 +284,16 @@ def set_local_pose(self, position=None, orientation=None):
...
                 rotq = lazy.pxr.Gf.Quatd(*orientation)
             xform_op.Set(rotq)
+        PoseAPI.invalidate()
+        if gm.ENABLE_FLATCACHE:
+            # If flatcache is on, make sure the USD local pose is synced to the fabric local pose.
+            # Ideally we should call usdrt's set local pose directly, but there is no such API.
+            # The only available API is SetLocalXformFromUsd, so we update USD first, and then sync to fabric.
+            xformable_prim = lazy.usdrt.Rt.Xformable(lazy.omni.isaac.core.utils.prims.get_prim_at_path(self.prim_path, fabric=True))
+            # We also need to clear the fabric's world pose (potentially set by physx) to force fabric to use
+            # the local pose we just set, because world pose takes precendence over local pose.
```
ChengshuLi, line 294: Fix the comment.

omnigibson/utils/usd_utils.py
```diff
@@ -549,10 +549,56 @@ def reset(cls):
...
+                og.sim.psi.fetch_results()
+            cls.mark_valid()
+        
+    @classmethod
+    def get_world_pose(cls, prim_path):
+        cls._refresh()
+        position, orientation = lazy.omni.isaac.core.utils.xforms.get_world_pose(prim_path)
+        return np.array(position), np.array(orientation)[[1, 2, 3, 0]]
+    @classmethod
+    def get_world_pose_with_scale(cls, prim_path):
```
ChengshuLi, line 588: missing a new line here

ChengshuLi (approved):
LGTM, with some minor comments!

PR #587, opened 2024-02-02
Add VX300S Robot

cremebrule:


omnigibson/prims/rigid_prim.py
```diff
@@ -201,7 +201,7 @@ def update_meshes(self):
                     # If we're not a valid volume, use bounding box approximation for the underlying collision approximation
                     if not is_volume:
                         log.warning(f"Got invalid (non-volume) collision mesh: {mesh.name}")
-                        mesh.set_collision_approximation("boundingCube")
+                        # mesh.set_collision_approximation("boundingCube")
```
ChengshuLi, line 204: is this no longer needed?

omnigibson/robots/vx300s.py
```diff
@@ -0,0 +1,244 @@
...
+        # Not supported for this robot
+        raise NotImplementedError()
+
+    def _create_discrete_action_space(self):
+        # Fetch does not support discrete actions
+        raise ValueError("VX300S does not support discrete actions!")
+
+    def update_controller_mode(self):
+        super().update_controller_mode()
+        # overwrite joint params (e.g. damping, stiffness, max_effort) here
```
ChengshuLi, line 138: Maybe remove this function is there is no overwriting?

ChengshuLi (approved):
LGTM, left two minor comments, feel free to merge!

omnigibson/prims/rigid_prim.py
cremebrule: Yes, per discussion with @cgokmen . I think our conclusion was that we should never do this anymore, or rather that (a) all of our assets should never have this problem, and (b) any edge cases we encounter we should explicitly address

omnigibson/robots/vx300s.py
cremebrule: done!

PR #588, opened 2024-02-05
[WIP] Ensure contact happens within slicer volume before slicing

arman-aydin:


omnigibson/examples/object_states/slicing_demo.py
```diff
@@ -89,6 +89,7 @@ def main(random_selection=False, headless=False, short_exec=False):
     knife.set_position_orientation(
         position=apple.get_position() + np.array([-0.15, 0.0, 0.2]),
         orientation=T.euler2quat([-np.pi / 2, 0, 0]),
+        # orientation=T.euler2quat([np.pi / 2, 0, 0])   # The back of the knife cuts the apple as well
```
cgokmen, line 92: In tests/test_transition_rules.py create two unit tests, one that has the object falling on the sharp side and cutting, one that has the object falling on the blunt side and not cutting. Essentially you need to create a function called `test_{test_name_here}` and run it using `pytest tests/test_transition_rules.py` and it should pass. Ask for in-person help from Eric if you can't figure out how to run pytest. Once the test is ready, remove this change.

omnigibson/transition_rules.py
```diff
@@ -411,6 +412,49 @@ def modifies_filter_names(self):
         return {self._filter_1_name}
 
 
+class TouchingVolumeCheckCondition(TouchingAnyCondition):
```
cgokmen, line 415: Let's call this TouchingSlicerVolumeCondition.

omnigibson/transition_rules.py
```diff
@@ -411,6 +412,49 @@ def modifies_filter_names(self):
...
+    def __init__(self, filter_1_name, filter_2_name):
+        '''
+        Checks if filter_2_bodies have a contact point within the collision volume generated from each obj in filter_1
+        '''
+        # super().__init__(filter_2_name, filter_1_name)
```
cgokmen, line 420: remove

omnigibson/transition_rules.py
```diff
@@ -411,6 +412,49 @@ def modifies_filter_names(self):
         return {self._filter_1_name}
 
 
+class TouchingVolumeCheckCondition(TouchingAnyCondition):
+    def __init__(self, filter_1_name, filter_2_name):
+        '''
+        Checks if filter_2_bodies have a contact point within the collision volume generated from each obj in filter_1
```
cgokmen, line 418: Let's explicitly say that we are using the slicer meta link as the volume here.

omnigibson/transition_rules.py
```diff
@@ -411,6 +412,49 @@ def modifies_filter_names(self):
...
+        objs = []
+
+        # Manually check contact
+        filter_2_bodies = set.union(*(self._filter_2_bodies[obj] for obj in object_candidates[self._filter_2_name]))
+        for obj in object_candidates[self._filter_1_name]:
+            slicer_volume = None
+            for name, link in self.obj.links.items():
+                if 'slicer' in name:
+                    slicer_volume = link
+                    break
```
cgokmen, line 434: Ask @cremebrule if this is the best way to do this.

omnigibson/transition_rules.py
```diff
@@ -411,6 +412,49 @@ def modifies_filter_names(self):
...
+            
+            if slicer_volume is None:
+                continue
+
+            check_in_volume, _ = generate_points_in_volume_checker_function(
+                obj=obj,
+                volume_link=slicer_volume  
+            )
```
cgokmen, line 442: Ask @cremebrule if repeatedly creating this function incurs a big performance penalty. We could do something clever here where the volume is known and the checker function pointer is held by the SlicerActive state and here we can just query it.

At the very least, add before this line:
```
# TODO: Consider moving this to the SlicerActive state for performance.
```

omnigibson/transition_rules.py
```diff
@@ -723,7 +767,10 @@ def candidate_filters(cls):
     @classmethod
     def _generate_conditions(cls):
         # sliceables should be touching any slicer
-        return [TouchingAnyCondition(filter_1_name="sliceable", filter_2_name="slicer"),
+        # return [TouchingAnyCondition(filter_1_name="sliceable", filter_2_name="slicer"),
+        #         StateCondition(filter_name="slicer", state=SlicerActive, val=True, op=operator.eq)]
+
```
cgokmen, line 772: Remove

cgokmen (changes requested):

cgokmen:
Issue here is that contact points dont show up where you expect them to. We need to figure out what to do

cgokmen:
@cgokmen to look into this later (low prio)

PR #599, opened 2024-02-14
Refactor system clear and MaterialPrim

ChengshuLi:


omnigibson/simulator.py
```diff
@@ -477,46 +478,50 @@ def import_object(self, obj, register=True):
...
+                # we first move the object to a safe location, then remove it.
+                pos = list(m.OBJECT_GRAVEYARD_POS)
+                for ob in objs:
+                    ob.set_position_orientation(pos, [0, 0, 0, 1])
+                    pos[0] += max(ob.aabb_extent)
 
-            # One timestep will elapse
-            self.app.update()
+                # One timestep will elapse
+                self.app.update()
```
cgokmen, line 503: Why this way instead of calling `self.step_physics` or if we dont want to clear APIs `self._physics_context._step(current_time=self.current_time)` like in `step_physics`? Just curious.

omnigibson/prims/xform_prim.py
```diff
@@ -78,7 +80,8 @@ def _post_load(self):
     def remove(self):
         # Remove the material prim if one exists
         if self._material is not None:
-            self._material.remove()
+            self._material.remove_user(self)
+            self._material.remove_if_unused()
```
cgokmen, line 84: If these are always called together, can remove_user just call self.remove_if_unused?

omnigibson/scenes/scene_base.py
```diff
@@ -460,15 +461,17 @@ def add_object(self, obj, register=True, _is_call_from_simulator=False):
 
         return prim
 
-    def remove_object(self, obj):
+    def remove_object(self, obj, has_registered=True):
```
cgokmen, line 464: Shouldn't this be called is_registered (across the board)?

omnigibson/simulator.py
```diff
@@ -477,46 +478,50 @@ def import_object(self, obj, register=True):
...
-            # Update all handles that are now broken because objects have changed
-            self.update_handles()
+            if self.is_playing():
+                # Update all handles that are now broken because objects have changed
+                self.update_handles()
 
-            # Load the state back
-            self.load_state(state)
+                # Load the state back
+                self.load_state(state)
```
cgokmen, line 513: After some recent work with Hang I'm under the impression that it's actually possible with physx to update the simulation without taking a time step (e.g. update collisions etc. but don't apply any impulse or displacement). That way we can avoid the step-and-load. Tracking now at #604

omnigibson/systems/macro_particle_system.py
```diff
@@ -30,6 +30,8 @@ class MacroParticleSystem(BaseSystem):
     # Note that this object is NOT part of the actual particle system itself!
     _particle_object = None
 
+    _particle_template = None
```
cgokmen, line 33: Do we need to track these separately? Is _particle_object not always _particle_template.visual_meshes[0]? If so, can we replace it with a property or something so that the duplication is removed?

omnigibson/prims/material_prim.py
```diff
@@ -28,6 +28,32 @@ class MaterialPrim(BasePrim):
             mtl_name (None or str): If specified, should be the name of the mtl preset to load.
                 None results in default, "OmniPBR"
     """
+    # Persistent dictionary of materials, mapped from prim_path to MaterialPrim
+    MATERIALS = dict()
```
cgokmen, line 32: I think at some point we still need to get a global prim registry instead of doing this ad-hoc this way for a particular prim type. Should we do that now? Or maybe you can open an issue.

omnigibson/systems/micro_particle_system.py
```diff
@@ -433,6 +440,17 @@ def initialize(cls):
         # Potentially modify the material
         cls._customize_particle_material()
 
+    @classmethod
+    def _clear(cls):
+        cls._material.remove_user(cls)
+        cls._material.remove_if_unused()
```
cgokmen, line 446: Same here

cgokmen (approved):
Left some comments, do take a look but LGTM otherwise

omnigibson/simulator.py
ChengshuLi: Sounds good. I have always been trying to do this for the last years and didn't manage to find a way :(

omnigibson/prims/material_prim.py
ChengshuLi: Makes sense. Let's do this in the next PR. I opened an issue.

omnigibson/simulator.py
ChengshuLi: It was `self.app.update()` before so I didn't think too much. But yeah, I think `self.step_physics()` makes more sense. Fixed.

omnigibson/prims/material_prim.py
```diff
@@ -37,6 +63,9 @@ def __init__(
         # Other values that will be filled in at runtime
         self._shader = None
 
+        # Users of this material: should be a list of BaseObject and BaseSystem
+        self._users = []
```
cremebrule, line 67: should this be a set instead? slightly more efficient, plus avoids redundancies

omnigibson/scenes/scene_base.py
```diff
@@ -460,15 +461,17 @@ def add_object(self, obj, register=True, _is_call_from_simulator=False):
 
         return prim
 
-    def remove_object(self, obj):
+    def remove_object(self, obj, is_registered=True):
```
cremebrule, line 464: is this flag necessary? maybe we can just automatically determine if the object is registered or not internally?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1406,6 +1436,7 @@ def _create_particle_prototypes(cls):
         # Load the particle template
         particle_template = cls._create_particle_template()
         og.sim.import_object(obj=particle_template, register=False)
+        cls._particle_template = particle_template
```
cremebrule, line 1439: should _create_particle_template() automatically store the particle template?

ChengshuLi: hmm I am not sure. I feel like it's cleaner if _create_particle_template just call the constructor of a template object, without worrying about importing or registering

cremebrule (approved):
lgtm!

PR #608, opened 2024-02-16
Optimize Controller Logic

cremebrule:


omnigibson/objects/object_base.py
```diff
@@ -165,6 +165,12 @@ def _post_load(self):
                     body1=f"{self._prim_path}/{self._root_link_name}",
                 )
 
+            # Delete n_fixed_joints cached property if it exists since the number of fixed joints has now changed
+            # See https://stackoverflow.com/questions/59899732/python-cached-property-how-to-delete and
+            # https://docs.python.org/3/library/functools.html#functools.cached_property
+            if "n_fixed_joints" in self.__dict__:
+                del self.n_fixed_joints
```
ChengshuLi, line 172: why do we need to do this? wouldn't n_fixed_joints automatically be updated?

ChengshuLi (approved):

omnigibson/objects/object_base.py
cremebrule: computing n_fixed_joints is actually non-trivial because we recursively search through the USD prim tree.

So n_fixed_joints is now a cached_property, which will get computed exactly once. However, this value changes when we are adding a fixed base root joint, so I delete it here so that the property will get recomputed (and re-cached) the next time it's called

PR #625, opened 2024-02-24
Fix missing import in exception type (we can be a lot more conservative anyway)

cgokmen:


omnigibson/prims/rigid_prim.py
```diff
@@ -626,7 +626,7 @@ def _compute_points_on_convex_hull(self, visual):
         try:
             hull = ConvexHull(points)
             return points[hull.vertices, :]
-        except scipy.spatial.qhull.QhullError:
+        except:
```
ChengshuLi, line 629: hmm do we know what other types of error there is?

cgokmen: I'm not sure if there are any tbh, there was an importerror which I fixed, but then I thought regardless of what goes wrong in that block we can safely use all the points so being more conservative and just using all the points in case of any error doesnt really hurt

hang-yin (approved):

PR #626, opened 2024-02-27
Accommodate non-uniformly scaled meshes

ChengshuLi:


omnigibson/utils/usd_utils.py
```diff
@@ -689,6 +691,63 @@ def mesh_prim_to_trimesh_mesh(mesh_prim, include_normals=True, include_texcoord=
...
+        trimesh_mesh = trimesh.creation.cone(radius=radius, height=height)
+        # Trimesh cones are centered at the base. We'll move them down by half the height.
+        transform = trimesh.transformations.translation_matrix([0, 0, -height / 2])
+        trimesh_mesh.apply_transform(transform)
+    elif mesh_type == "Cylinder":
+        radius = mesh_prim.GetAttribute("radius").Get()
+        height = mesh_prim.GetAttribute("height").Get()
+        trimesh_mesh = trimesh.creation.cylinder(radius=radius, height=height)
+    else:
+        raise ValueError(f"Expected mesh prim to have type Sphere, Cube, Cone or Cylinder, got {mesh_type}")
```
cgokmen, line 723: It's really clean to centralize this logic, but it's a significant slowdown to compute these throwaway meshes compared to the previous approach which analytically computed the volumes etc. - I guess we're OK with that?

omnigibson/utils/geometry_utils.py
```diff
@@ -238,7 +238,7 @@ def _generate_convex_hull_volume_checker_functions(convex_hull_mesh):
                 USD mesh
     """
     # For efficiency, we pre-compute the mesh using trimesh and find its corresponding faces and normals
-    trimesh_mesh = mesh_prim_to_trimesh_mesh(convex_hull_mesh, include_normals=False, include_texcoord=False).convex_hull
+    trimesh_mesh = mesh_prim_mesh_to_trimesh_mesh(convex_hull_mesh, include_normals=False, include_texcoord=False).convex_hull
```
cgokmen, line 241: I'm getting a feeling that a lot of stuff in this function (the pos-quat-scale stuff especially) can possibly have issues with the skew etc right? Should we try to unify everything to use the 4x4 scaled transform matrices instead of these adhoc partial transforms?

omnigibson/systems/micro_particle_system.py
```diff
@@ -1616,6 +1617,15 @@ def clothify_mesh_prim(cls, mesh_prim, remesh=True, particle_distance=None):
             mesh_prim.GetAttribute("normals").Set(lazy.pxr.Vt.Vec3fArray.FromNumpy(new_normals))
             mesh_prim.GetAttribute("primvars:st").Set(lazy.pxr.Vt.Vec2fArray.FromNumpy(new_texcoord))
 
+            # Load the new mesh into trimesh and apply the inverse of the world transform
```
cgokmen, line 1620: Right now this code does pymeshlab -> omni -> trimesh -> omni. Can we change that to pymeshlab -> trimesh -> omni? Essentially instead of the stuff in 1607-1612 we would first create the trimesh and then get the data from trimesh.

omnigibson/prims/rigid_prim.py
```diff
@@ -195,10 +196,10 @@ def update_meshes(self):
                     mesh.set_rest_offset(m.DEFAULT_REST_OFFSET)
                     self._collision_meshes[mesh_name] = mesh
 
-                    is_volume, volume, com = get_mesh_volume_and_com(mesh_prim)
-                    vols.append(volume)
-                    # We need to translate the center of mass from the mesh's local frame to the link's local frame
+                    volume, com = get_mesh_volume_and_com(mesh_prim)
```
cgokmen, line 199: I think we should do this accumulation in the world frame too and then at the very end for the CoM we should apply the inverse scaled world transform for this object. Otherwise the volumes are not affected by the scales of any parents etc.

cgokmen (commented):
Some quick things. Thanks!

omnigibson/utils/usd_utils.py
ChengshuLi: I think this should be okay. Our code only compute the volume once when initializing the objects.

omnigibson/utils/geometry_utils.py
ChengshuLi: I don't think this has issues with the skew. The checker here is to transform points from the link frame to the mesh frame, so the math is accurate (pos, quat, scale should be enough). I agree it's possible to unify everything using the scaled transform matrices, but this part of the code is kind of complicated and well-tested and I don't want to break it. We can unify later I think.

omnigibson/prims/rigid_prim.py
ChengshuLi: Actually what you suggested won't work. Here we need the meshes CoM and volume in the *link* frame, not in the *world* frame. These computations are only for computing and setting the centerOfMass in the link frame, so I think my math is correct. But please sanity check me once again on this :D

omnigibson/systems/micro_particle_system.py
```diff
@@ -1610,10 +1611,27 @@ def clothify_mesh_prim(cls, mesh_prim, remesh=True, particle_distance=None):
...
 
-            mesh_prim.GetAttribute("faceVertexCounts").Set(np.ones(n_faces, dtype=int) * 3)
-            mesh_prim.GetAttribute("points").Set(lazy.pxr.Vt.Vec3fArray.FromNumpy(new_vertices))
+            face_vertex_counts = np.ones(n_faces, dtype=int) * 3
+            faces = []
+            i = 0
+            for count in face_vertex_counts:
+                for j in range(count - 2):
+                    faces.append([new_face_vertex_ids[i], new_face_vertex_ids[i + j + 1], new_face_vertex_ids[i + j + 2]])
+                i += count
```
cgokmen, line 1620: You can just not call .flatten() on 1608 and you'll get the array you're building here. You can flatten it later when passing it into 1631. Or you can still flatten it but pass it into trimesh as .reshape(-1, 3)

omnigibson/prims/rigid_prim.py
cgokmen: OK I see now that we don't return the volume and just use it for CoM computation. That should be fine. Thanks!

cgokmen (approved):
Left a quick nit but LGTM otherwise. Thanks!

omnigibson/systems/micro_particle_system.py
ChengshuLi: done!

PR #628, opened 2024-02-29
Mesh scaling v2

ChengshuLi:


cgokmen:
Wow thats a lot of great finds. Thanks!

cgokmen (approved):

omnigibson/prims/cloth_prim.py
```diff
@@ -72,6 +74,9 @@ def __init__(
         )
 
     def _post_load(self):
+        # Cache the original scale of the visual mesh
```
cremebrule, line 77: maybe add additional comment why we want to cache the original scale in the first place?

omnigibson/prims/xform_prim.py
```diff
@@ -75,10 +75,6 @@ def _post_load(self):
             material.add_user(self)
             self._material = material
 
-        # Optionally set the scale and visibility
```
cremebrule, line 78: why is this no longer done? is it because we allocate the setting in entity prim / we don't want to override the native scale?

tests/test_transition_rules.py
```diff
@@ -79,8 +80,10 @@ def test_washer_rule():
     og.sim.step()
 
     # Place the two objects inside the washer
-    remover_dishtowel.set_position_orientation([0.0, 0.0, 0.03], [0, 0, 0, 1])
-    bowl.set_position_orientation([0.10, 0.0, 0.06], [0, 0, 0, 1])
+    # (Hacky) use baking_sheet as a stepping stone to elevate the objects so that they are inside the container volume.
+    baking_sheet.set_position_orientation([0.0, 0.0, 0.04], R.from_euler("xyz", [np.pi, 0, 0]).as_quat())
```
cremebrule, line 84: shouldn't we use T.euler2quat?

cremebrule (approved):
lgtm! a couple minor comments

omnigibson/prims/xform_prim.py
ChengshuLi: Post-discussion outcome: we undo this change, still keep the optional of setting the scale in load_config of XFormPrim. However, we don't pass in scale in the load_config when we are instantiating RigidPrim/ClothPrim for each link in EntityPrim.

PR #629, opened 2024-03-01
make GranularSystems not self-collidable to improve physics stability

cremebrule:


omnigibson/systems/micro_particle_system.py
```diff
@@ -1414,6 +1411,11 @@ class GranularSystem(MicroPhysicalParticleSystem):
 
     _particle_template = None
 
+    @classproperty
+    def self_collision(cls):
+        # Don't self-collide to improve physics stability
```
ChengshuLi, line 1416: maybe add more explanation here? I just want to make sure when we look back at this a couple months later, we know this hack is solving what issue exactly.

ChengshuLi (approved):

omnigibson/systems/micro_particle_system.py
cremebrule: done!

PR #594, opened 2024-02-12
SyntheticData -> Replicator

hang-yin:
A continuation of @cremebrule 's branch [here](https://github.com/StanfordVL/OmniGibson/tree/refactor-vision-sensor). 

This PR should be a fix to: 
#458 
#462 
#66 
#301 

Some follow-ups to this PR: 
- [x] Investigate how `particle systems` and `cloth` show up in semantic segmentation.
- [x] Unit test for remapped ID (waiting for this PR #582 to get merged so we can add this test to `test_robot_states.py`).

PR #631, opened 2024-03-01
lazy access of gm values

hang-yin:
#603 
This PR removes eager uses of things like gm.DATASET_PATH e.g. in loading systems so that these values can be changed at runtime before initializing env. 

Note: this PR needs to wait until PR #594 gets merged because we need some refactoring from there.

### DIFF
--- a/omnigibson/__init__.py
+++ b/omnigibson/__init__.py
@@ -31,8 +31,6 @@ nest_asyncio.apply()
 
 __version__ = "0.2.1"
 
-log.setLevel(logging.DEBUG if gm.DEBUG else logging.INFO)
-
 root_path = os.path.dirname(os.path.realpath(__file__))
 
 # Store paths to example configs
--- a/omnigibson/envs/env_base.py
+++ b/omnigibson/envs/env_base.py
@@ -3,7 +3,6 @@ import numpy as np
 from copy import deepcopy
 
 import omnigibson as og
-from omnigibson.objects import REGISTERED_OBJECTS
 from omnigibson.robots import REGISTERED_ROBOTS
 from omnigibson.scene_graphs.graph_builder import SceneGraphBuilder
 from omnigibson.simulator import launch_simulator
@@ -253,6 +252,7 @@ class Environment(gym.Env, GymObservable, Recreatable):
                 obj_config["name"] = f"obj{i}"
             # Pop the desired position and orientation
             position, orientation = obj_config.pop("position", None), obj_config.pop("orientation", None)
+            from omnigibson.objects import REGISTERED_OBJECTS
             # Make sure robot exists, grab its corresponding kwargs, and create / import the robot
             obj = create_class_from_registry_and_config(
                 cls_name=obj_config["type"],
--- a/omnigibson/scenes/scene_base.py
+++ b/omnigibson/scenes/scene_base.py
@@ -24,9 +24,6 @@ log = create_module_logger(module_name=__name__)
 # Create settings for this module
 m = create_module_macros(module_path=__file__)
 
-# Default texture to use for skybox
-m.DEFAULT_SKYBOX_TEXTURE = f"{gm.ASSET_PATH}/models/background/sky.jpg"
-
 # Global dicts that will contain mappings
 REGISTERED_SCENES = dict()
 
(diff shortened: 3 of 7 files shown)


ChengshuLi (approved):
LGTM, although I don't entirely understand the motivation of this PR. @cgokmen maybe you can give the final approval and merge this?

omnigibson/envs/env_base.py
```diff
@@ -253,6 +252,7 @@ def _load_objects(self):
                 obj_config["name"] = f"obj{i}"
             # Pop the desired position and orientation
             position, orientation = obj_config.pop("position", None), obj_config.pop("orientation", None)
+            from omnigibson.objects import REGISTERED_OBJECTS
```
cgokmen, line 255: Is this actually needed?

omnigibson/scenes/scene_base.py
```diff
@@ -199,7 +196,9 @@ def _load(self):
             )
             og.sim.import_object(self._skybox, register=False)
             self._skybox.color = (1.07, 0.85, 0.61)
-            self._skybox.texture_file_path = m.DEFAULT_SKYBOX_TEXTURE
+            # Default texture to use for skybox
+            default_skybox_texture = f"{gm.ASSET_PATH}/models/background/sky.jpg"
+            self._skybox.texture_file_path = default_skybox_texture
```
cgokmen, line 201: nit: make this into one line. no need for default_skybox_texture

omnigibson/simulator.py
```diff
@@ -149,6 +149,9 @@ def _launch_app():
     return app
 
 
+
+    log.setLevel(logging.DEBUG if gm.DEBUG else logging.INFO)
+
```
cgokmen, line 154: What is this? This doesnt even execute. Where does it actually need to go? Check with @cremebrule

omnigibson/simulator.py
```diff
@@ -1370,6 +1373,10 @@ def _deserialize(self, state):
             return self._scene.deserialize(state=state), self._scene.state_size
 
     if not og.sim:
+
```
cgokmen, line 1376: Supernit: remove empty line after if

omnigibson/simulator.py
```diff
@@ -1370,6 +1373,10 @@ def _deserialize(self, state):
             return self._scene.deserialize(state=state), self._scene.state_size
 
     if not og.sim:
+        
+        from omnigibson.systems.system_base import import_og_systems
+        # Import all OG systems from dataset
+        import_og_systems()
```
cgokmen, line 1379: This is kind of annoying in that it opaquely makes system load dependent on simulator launch but we dont say that anywhere. Let's reconsider this approach - we can discuss tomorrow.

cgokmen (changes requested):
Let's delay this PR to post-freeze.

cgokmen (approved):

cgokmen:
@cremebrule please review this too

hang-yin:
This PR also addresses #655

omnigibson/__init__.py
```diff
@@ -30,8 +30,6 @@
 
 __version__ = "1.0.0"
 
-log.setLevel(logging.DEBUG if gm.DEBUG else logging.INFO)
```
cremebrule, line 33: why was this removed again?

cremebrule: because gm is being accessed too early right?

omnigibson/macros.py
```diff
@@ -10,33 +10,57 @@
...
+class MacroDict(Dict):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self["_read"] = set()
+
+    def __setattr__(self, name, value):
+        if name in self.get("_read", set()):
+            raise AttributeError(f"Cannot set attribute {name} in MacroDict, it has already been used.")
+        # Use the super's setattr for setting attributes, but handle _read directly to avoid recursion.
+        if name == "_read":
```
cremebrule, line 23: maybe we can have a lock / unlock feature? Only certain macros shouldn't be touched after they're read, but a lot of them can be modified later. What do you think @cgokmen ? We could take inspiration from https://github.com/ARISE-Initiative/robomimic/blob/master/robomimic/config/config.py#L37

omnigibson/scenes/scene_base.py
```diff
@@ -208,7 +205,8 @@ def _load(self):
             )
             og.sim.import_object(self._skybox, register=False)
             self._skybox.color = (1.07, 0.85, 0.61)
-            self._skybox.texture_file_path = m.DEFAULT_SKYBOX_TEXTURE
+            # Default texture to use for skybox
+            self._skybox.texture_file_path = f"{gm.ASSET_PATH}/models/background/sky.jpg"
```
cremebrule, line 209: why is this no longer a macro?

cremebrule (commented):
overall LGTM! Just a few minor comments

omnigibson/__init__.py
hang-yin: Yes! This is moved to the start of `_launch_app`. Please let me know if that make sense.

omnigibson/scenes/scene_base.py
hang-yin: `m.DEFAULT_SKYBOX_TEXTURE` would access gm before users have a chance to change the asset path. Also, this is only used once here, so I just swapped it out with the actual definition.

PR #636, opened 2024-03-05
Require positive scales and validate local transforms

cgokmen:


ChengshuLi (approved):
LGTM, but are these two lines sufficient/complete? Do we need to add more assertions in other places of our code, or is that it?

PR #643, opened 2024-03-12
Fixes for all example scripts

hang-yin:
With dataset `3a773e40047e7e104a819fe0a9d03622`
And with dataset `1.0.0`


omnigibson/examples/robots/all_robots_visualizer.py
```diff
@@ -18,6 +18,7 @@ def main(random_selection=False, headless=False, short_exec=False):
     }
 
     env = og.Environment(configs=cfg)
+    og.sim.stop()
```
cgokmen, line 21: Why do we do this? It makes many things (e.g. robot.keep_still()) meaningless

omnigibson/examples/robots/advanced/ik_example.py
```diff
@@ -38,7 +38,12 @@ def main(random_selection=False, headless=False, short_exec=False):
         programmatic_pos = True
 
     # Import scene and robot (Fetch)
+    scene_cfg = {"type": "Scene"}
+    cfg = dict(scene=scene_cfg)
+    env = og.Environment(configs=cfg)
     scene = Scene()
+    if og.sim.is_playing():
+        og.sim.stop()
```
cgokmen, line 46: All this is quite weird. The env already has a scene so why import a new one, also why stop it and import a robot afterwards? Lets put the scene and the robot in the config, and don't make any additional stop/import/etc. calls.

cgokmen (approved):
A couple quick comments. Also I'm not sure we want this IK example. Why would someone control the robot that way?

PR #645, opened 2024-03-12
Don't mess up toggle button offsets

cgokmen:
This is one of those things where I don't know what the code is supposed to be doing but it's doing something wrong currently (objects like charcoal_grill-crdaeu end up with their toggleable meshes moved to the wrong spot)


cremebrule (approved):
dunno why i wrote that line. feel free to remove it for now, and we can revert if something breaks

PR #651, opened 2024-03-14
Attached sampling

ChengshuLi:


omnigibson/utils/bddl_utils.py
```diff
@@ -191,7 +191,7 @@ def process_single_condition(condition):
     "insource": ObjectStateInsourcePredicate,
 }
 
-KINEMATIC_STATES_BDDL = frozenset([state.__name__.lower() for state in _KINEMATIC_STATE_SET])
+KINEMATIC_STATES_BDDL = frozenset([state.__name__.lower() for state in _KINEMATIC_STATE_SET] + ["attached"])
```
cremebrule, line 194: why is AttachedTo not a kinematic state?

omnigibson/utils/bddl_utils.py
```diff
@@ -846,6 +897,10 @@ def _filter_object_scope(self, input_object_scope, conditions, condition_type):
                             # Reset if we're sampling a kinematic state
                             if condition.STATE_NAME in {"inside", "ontop", "under"}:
                                 kwargs["reset_before_sampling"] = True
+                            elif condition.STATE_NAME in {"attached"}:
+                                kwargs["bypass_alignment_checking"] = True
+                                kwargs["check_physics_stability"] = True
+                                kwargs["can_joint_break"] = False
```
cremebrule, line 903: add TODO here to note that we need to revert this when possible in the future?

omnigibson/utils/bddl_utils.py
```diff
@@ -965,53 +1077,53 @@ def _import_sampleable_objects(self):
...
+                model_choices = set()
+                for category in categories:
+                    # Get all available models that support all of its synset abilities
+                    model_choices = set(get_all_object_category_models_with_abilities(
                         category=category,
-                        model=model,
-                        prim_type=PrimType.CLOTH if "cloth" in OBJECT_TAXONOMY.get_abilities(obj_synset) else PrimType.RIGID,
-                    )
-                    num_new_obj += 1
+                        abilities=OBJECT_TAXONOMY.get_abilities(OBJECT_TAXONOMY.get_synset_from_category(category)),
```
cremebrule, line 1095: This code seems to change a lot from the previous iteration. Are we sure this doesn't break sampling?

cremebrule (approved):
overall LGTM, some clarifying comments

omnigibson/utils/bddl_utils.py
ChengshuLi: Not sure, but I don't want to change the current behavior. We can make it inherit KinematicStateMixin later.

omnigibson/utils/bddl_utils.py
ChengshuLi: why do we need to revert this?

omnigibson/utils/bddl_utils.py
ChengshuLi: most of them are indentation changes. Yeah I tested some other task and I think it doesn't break.

PR #667, opened 2024-03-19
Replicator hot fix

hang-yin:
Two hot fixes for:
1. Mismatched segmentation & id_to_label
2. When some bounding box semantic IDs are not yet present in the semantic segmentation


hang-yin:
In terms of performance: 
profile

cgokmen (approved):

PR #672, opened 2024-03-20
Black formatter

hang-yin:


cgokmen (approved):
Rubber stamping - hope nothing breaks :)

PR #680, opened 2024-03-21
Check Isaac Sim version

hang-yin:


omnigibson/utils/sim_utils.py
```diff
@@ -363,5 +363,5 @@ def land_object(obj, pos, quat=None, z_offset=None):
     obj.keep_still()
 
 
-def meets_minimum_isaac_version(minimum_version):
-    return python_utils.meets_minimum_version(lazy.omni.isaac.version.get_version()[0], minimum_version)
```
cgokmen, line 367: Dont do this. This function works well in other contexts. Here you convert it into a meaningless alias for the python_utils function. Instead, in our case above in simulator.py just call the python_utils function.

cgokmen (changes requested):

PR #686, opened 2024-03-25
Implement centralized control getter/setter for fast access in parallel envs

cgokmen:
Currently, adding more robots into the simulation significantly increases the amount of time spent in the controller step. This aims to alleviate that by unifying articulationview reads/writes.


cgokmen:
Will be included in #699

PR #698, opened 2024-04-04
Fix/general improvements

cremebrule:


omnigibson/scenes/interactive_traversable_scene.py
```diff
@@ -138,10 +138,10 @@ def filter_rooms_and_object_categories(
                 load_room_instances = [load_room_instances]
             load_room_instances_filtered = []
             for room_instance in load_room_instances:
-                if room_instance in self._seg_map.room_ins_name_to_ins_id:
-                    load_room_instances_filtered.append(room_instance)
-                else:
+                load_room_instances_filtered.append(room_instance)
+                if room_instance not in self._seg_map.room_ins_name_to_ins_id:
                     log.warning("room_instance [{}] does not exist.".format(room_instance))
+                load_room_instances_filtered.append(room_instance)
```
cgokmen, line 144: You're appending twice

omnigibson/robots/behavior_robot.py
```diff
@@ -139,11 +140,11 @@ def usd_path(self):
     def model_name(self):
         return "BehaviorRobot"
 
-    @property
+    @classproperty
     def n_arms(self):
```
cgokmen, line 144: Should this still be `self` and not `cls` if its a class property? not sure.

omnigibson/simulator.py
```diff
@@ -230,8 +230,8 @@ class Simulator(lazy.omni.isaac.core.simulation_context.SimulationContext, Seria
         def __init__(
             self,
             gravity=9.81,
-            physics_dt=1.0 / 120.0,
-            rendering_dt=1.0 / 30.0,
+            physics_dt=1.0 / gm.DEFAULT_PHYSICS_FREQ,
+            rendering_dt=1.0 / gm.DEFAULT_RENDERING_FREQ,
```
cgokmen, line 234: This is bad - it gets processed at import time so changing the macro does nothing. Can we default this into None and then do `if physics_dt is None: physics_dt = 1.0 / gm. ...` down the line

omnigibson/utils/ui_utils.py
```diff
@@ -485,7 +485,7 @@ def get_interpolated_positions(step):
                 pan_angle = np.arctan2(-xy_direction[0], xy_direction[1])
                 tilt_angle = np.arcsin(z)
                 # Infer global quat orientation from these angles
-                quat = T.euler2quat([np.pi / 2 - tilt_angle, 0.0, pan_angle])
+                quat = T.euler2quat([np.pi / 2 + tilt_angle, 0.0, pan_angle])
```
cgokmen, line 488: Assuming you have tested this

omnigibson/robots/behavior_robot.py
cremebrule: fixed!

omnigibson/simulator.py
cremebrule: fixed

ChengshuLi (approved):

PR #708, opened 2024-04-19
Remove simplified mesh from primitives

sujaygarlanka:


cgokmen:
@cgokmen to see if we did this already / merge if not

PR #734, opened 2024-05-23
Fix: Importing an alternative robot into presampled BehaviorTask doesn't work

hang-yin:
Original question on discord: 
```
Hi - I'm looking to try a behavior task with Tiago. However, many of the task configurations specify the Fetch robot in the json. 

What is the recommended workflow?

I'd like to be able to test an LLM agent against the Behavior 1K tasks but would start with a couple individual tasks. 

Currently I've chosen `putting_dirty_dishes_in_sink` however it explicitly has Fetch in the json instead of Tiago (which I'm using for the fully orientable camera head). 

Should I duplicate the task json and remove reference to Fetch?
```


cgokmen:
I believe we have already fixed this, so I'm closing it.

PR #735, opened 2024-05-26
fix render_product syncing when updating camera width / height

cremebrule:


hang-yin (approved):
LGTM! nit: the debugging statements are still in there

PR #781, opened 2024-06-27
Revive transition rules

ChengshuLi:


omnigibson/transition_rules.py
```diff
@@ -63,19 +62,23 @@
...
 class TransitionRuleAPI:
     """
-    Monolithic class containing methods to check and execute arbitrary discrete state transitions within the simulator
+    Containing methods to check and execute arbitrary discrete state transitions within the simulator
     """
 
-    # Set of active rules
-    ACTIVE_RULES = set()
+    def __init__(self, scene):
+        self.scene = scene
```
hang-yin, line 69: Similar to objects, do we want to break the circular dependency between scene and TransitionRuleAPI here by keeping a scene index here instead?

hang-yin (commented):
Do we also want to add a multiple-envs transition rule test somewhere?

omnigibson/transition_rules.py
cgokmen: The scene idx thing feels a bit brittle. Did we actually do that for objects yet or did we create a ticket for it? Lets add this comment to that ticket so that we can discuss it later. I'm not sure I agree with it.

cgokmen (approved):
I wasnt sure at first if I agreed with the design choice of putting one API per scene (it does mean _less_ tensorization possible which is not good) but it makes this PR much easier and also the future task of recording transition history also becomes better scoped (e.g. it can be saved/loaded with the scene). So I think it's good for now. Thanks for the cleanup!

omnigibson/scenes/scene_base.py
```diff
@@ -679,8 +683,25 @@ def get_system(self, system_name, force_active=True):
...
+    def active_physical_particle_systems(self):
+        return {
+            system.name: system
+            for system in self.systems
+            if system.initialized and self.is_physical_particle_system(system.name)
+        }
+
+    @property
+    def active_systems(self):
+        return dict(**self.active_visual_particle_systems, **self.active_physical_particle_systems)
```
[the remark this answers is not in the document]
ChengshuLi: Removed!

omnigibson/transition_rules.py
ChengshuLi: We did create a ticket for it. Yeah we can discuss further.

PR #803, opened 2024-07-16
Improve ControllableObjectViewAPI to work with different robot types

cgokmen:
Here we create a separate view for each type of robot. We also remove the separate dummy view and let the dummies also be handled by the main view. See the docstring on the new ControllableObjectViewAPI for implementation details.

Currently untested. Please review while I work on testing it.


hang-yin (approved):

omnigibson/utils/usd_utils.py
```diff
@@ -1032,8 +1018,188 @@ def get_relative_jacobian(self, prim_path):
...
+        from omnigibson.objects.controllable_object import ControllableObject
+
+        controllable_objects = [
+            obj for scene in og.sim.scenes for obj in scene.objects if isinstance(obj, ControllableObject)
+        ]
+
+        # Get their corresponding prim paths
+        expected_regular_prim_paths = {obj.articulation_root_path for obj in controllable_objects}
+        expected_dummy_prim_paths = {
+            obj._dummy.articulation_root_path
```
cremebrule, line 1062: wish there was a more elegant way to infer this but for now this is probably ok

omnigibson/utils/usd_utils.py
```diff
@@ -1032,8 +1018,188 @@ def get_relative_jacobian(self, prim_path):
...
+    @classmethod
+    def _get_pattern_from_prim_path(cls, prim_path):
+        """
+        Returns which of the regexes this prim path should be found in.
+
+        Note that since the prim path will be an articulation root path, this works for both fixed base and
+        non-fixed base robots. Fixed and non-fixed versions of the same robot will be mapped to different
+        patterns since they have different articulation root paths (object prim vs base link prim).
+        """
+        scene_id, robot_name = prim_path.split("/")[2:4]
```
cremebrule, line 1101: this seems quite brittle. will this always be true?

cremebrule: maybe we can infer this directly from the robot objects instead of their prim paths?

omnigibson/utils/usd_utils.py
```diff
@@ -1032,8 +1018,188 @@ def get_relative_jacobian(self, prim_path):
...
+        assert scene_id.startswith("scene_"), f"Prim path 2nd component {prim_path} does not start with scene_"
+        components = robot_name.split("__")
+        assert len(components) == 3, f"Robot name {robot_name} does not match expected format"
```
cremebrule, line 1104: Can this be explicit -- what exactly do we expect the 3 components to be?

cremebrule (approved):
a few small comments but overall LGTM!

omnigibson/utils/usd_utils.py
cgokmen: Done.

omnigibson/utils/usd_utils.py
cgokmen: Currently our setup involves only calling this API with the articulation root prim path and I'd be hesitant to replace that at least in this small PR. For now I am adding an assertion into controllable object that this is always true.

omnigibson/utils/usd_utils.py
cgokmen: Yeah like a "get_controllable_prim_paths" or something but it'd be very specific. I think we can live with this for now.

PR #811, opened 2024-07-19
Data Collection / Playback Environments

cremebrule:
This introduces environments for data collection / playback, as well as necessary infra changes required to support this feature. Moreover, miscellaneous infra bugs were fixed along the way as I ran into them.


tests/test_dump_load_states.py
```diff
@@ -5,43 +5,76 @@
...
+#             system.generate_particles(positions=[[0, 0, 1]])
+#         assert system.n_particles > 0
+#
+#     state = og.sim.dump_state(serialized=True)
+#     og.sim.load_state(state, serialized=True)
+#
+#     for system_name, system_class in SYSTEM_EXAMPLES.items():
+#         system = env.scene.get_system(system_name)
+#         system.clear()
+
```
hang-yin, line 48: Would be nice if we have a dump-load test for multiple environments, but it's not super relevant to data collection I guess

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,498 @@
...
+        for k, dat in data.items():
+            if "obs" in k:
+                obs_grp = traj_grp.create_group(k)
+                for mod, traj_mod_data in dat.items():
+                    obs_grp.create_dataset(mod, data=np.stack(traj_mod_data, axis=0))
+            else:
+                try:
+                    traj_grp.create_dataset(k, data=np.stack(dat, axis=0))
+                except ValueError:
+                    breakpoint()
```
hang-yin, line 158: Is this try-except-breakpoint here intentional?

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,498 @@
...
+        # Make sure obs space is flattened for recording
+        config["env"]["flatten_obs_space"] = True
+
+        # Set scene file and disable online object sampling if BehaviorTask is being used
+        config["scene"]["scene_file"] = json.loads(f["data"].attrs["scene_file"])
+        if config["task"]["type"] == "BehaviorTask":
+            config["task"]["online_object_sampling"] = False
+
+        # for i, robot_cfg in enumerate(config["robots"]):
+        #     robot_cfg["name"] = f"robot_new{i}"
```
hang-yin, line 372: Is this still relevant?

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,498 @@
...
+            only_successes (bool): Whether to only save successful episodes
+
+        Returns:
+            DataPlaybackWrapper: Generated playback environment
+        """
+        # Read from the HDF5 file
+        f = h5py.File(input_path, "r")
+        config = json.loads(f["data"].attrs["config"])
+
+        # Hot swap in additional info for playing back data
```
hang-yin, line 357: Is this still relevant?

hang-yin (commented):
overall LGTM! Left several minor comments

omnigibson/object_states/temperature.py
```diff
@@ -38,8 +38,12 @@ def update_temperature_from_heatsource_or_sink(cls, objs, temperature, rate):
             rate (float): Heating rate of the source / sink
         """
         # Get idxs for objs
-        idxs = [cls.OBJ_IDXS[obj] for obj in objs]
-        cls.VALUES[idxs] += (temperature - cls.VALUES[idxs]) * rate * og.sim.get_rendering_dt()
+        idxs = np.zeros(len(objs), dtype=int)
+        for i, obj in enumerate(objs):
+            # No need to explicitly add objects to updated objects set here, since @self.global_update() automatically
+            # handles this which occurs after all heatsources are updated
+            idxs[i] = cls.OBJ_IDXS[obj]
```
ChengshuLi, line 45: are these not equivalent to the original code?

omnigibson/prims/entity_prim.py
```diff
@@ -983,16 +1019,16 @@ def set_position_orientation(self, position=None, orientation=None):
             self._articulation_view.set_world_poses(position, orientation)
             PoseAPI.invalidate()
 
-    def get_position_orientation(self):
+    def get_position_orientation(self, clone=True):
```
ChengshuLi, line 1022: what is clone?

omnigibson/prims/entity_prim.py
```diff
@@ -1531,18 +1567,28 @@ def create_attachment_point_link(self):
...
-        joint_state = dict()
-        for prim_name, prim in self._joints.items():
-            joint_state[prim_name] = prim._dump_state()
-        state["joints"] = joint_state
+        if self.n_joints > 0:
+            state["joint_pos"] = self.get_joint_positions()
+            state["joint_vel"] = self.get_joint_velocities()
+            state["joint_eff"] = self.get_joint_efforts()
+            # state["joint_pos_target"] = self.get_joint_position_targets()
+            # state["joint_vel_target"] = self.get_joint_velocity_targets()
```
ChengshuLi, line 1575: still needed?

omnigibson/prims/entity_prim.py
```diff
@@ -1531,18 +1567,28 @@ def create_attachment_point_link(self):
...
+
+        # TODO: Remove this backwards-compatible part once we re-sample scenes
+        if "joints" in state:
+            for joint_name, joint_state in state["joints"].items():
+                self._joints[joint_name]._load_state(state=joint_state)
+
+        elif self.n_joints > 0:
+            self.set_joint_positions(state["joint_pos"])
+            self.set_joint_velocities(state["joint_vel"])
+            self.set_joint_efforts(state["joint_eff"])
```
ChengshuLi, line 1591: Is there any benefit of saving/loading joint states at the entity prim level rather than the joint level?

omnigibson/prims/rigid_prim.py
```diff
@@ -320,16 +326,29 @@ def set_position_orientation(self, position=None, orientation=None):
...
             return self._kinematic_world_pose_cache
 
-        pos, ori = self._rigid_prim_view.get_world_poses()
+        pos, ori = self._rigid_prim_view.get_world_poses(clone=clone)
 
-        assert np.isclose(
-            np.linalg.norm(ori), 1, atol=1e-3
-        ), f"{self.prim_path} orientation {ori} is not a unit quaternion."
+        if gm.DEBUG:
+            # Only run this assert when debugging since this actually results in some nontrivial slowdown
```
ChengshuLi, line 348: hmm slightly concerned about silencing this assertion. we can discuss!

omnigibson/robots/robot_base.py
```diff
@@ -53,7 +53,6 @@ def __init__(
         # Shared kwargs in hierarchy
         name,
         relative_prim_path=None,
-        uuid=None,
```
ChengshuLi, line 56: why remove uuid across all the objects?

omnigibson/scenes/scene_base.py
```diff
@@ -668,18 +743,41 @@ def is_fluid_system(self, system_name):
...
+    def remove_system(self, system_name):
+        """
+        Clear the system @system_name and remove it from our set of active systems
+
+        Args:
+            system_name (str): Name of the system to remove
+        """
+        system = self.system_registry("name", system_name)
+        if system is not None:
+            system.clear()
```
ChengshuLi, line 776: should we also remove it from the system registry?

agreed with Hang that we should also update the doc.

omnigibson/scenes/scene_base.py
```diff
@@ -259,13 +296,17 @@ def _load(self):
     def _load_systems(self):
         system_dir = os.path.join(gm.DATASET_PATH, "systems")
 
+        available_systems = dict()
         if os.path.exists(system_dir):
             system_names = os.listdir(system_dir)
             for system_name in system_names:
```
ChengshuLi, line 302: maybe you intend to use `get_all_system_names` here?

omnigibson/simulator.py
```diff
@@ -569,6 +591,48 @@ def add_skybox(self):
...
+            rendering_dt = self.get_rendering_dt()
+            physics_dt = self.get_physics_dt()
+            render_physics_ratio = rendering_dt / physics_dt
+            sim_render_ratio = self._sim_step_dt / rendering_dt
+            assert np.isclose(render_physics_ratio, np.round(render_physics_ratio)), \
+                f"Rendering dt ({rendering_dt}) must be a multiple of physics dt ({physics_dt})"
+            assert np.isclose(sim_render_ratio, np.round(sim_render_ratio)), \
+                f"Simulation step dt ({self._sim_step_dt}) must be a multiple of rendering dt ({rendering_dt})"
+            assert self._sim_step_dt > rendering_dt, \
+                f"Simulation step dt ({self._sim_step_dt}) cannot be smaller than rendering dt ({rendering_dt})"
```
ChengshuLi, line 634: should we put these sanity check in a helper function and reuse?

omnigibson/simulator.py
```diff
@@ -1322,43 +1407,82 @@ def get_metadata(self, key):
...
+
+                    # Add any extra objects that do not currently exist in the scene stage
+                    for obj_to_add in objs_to_add:
+                        obj = create_object_from_init_info(scene_info["objects_info"]["init_info"][obj_to_add])
+                        scene.add_object(obj)
+
+            # Start the simulation and restore the dynamic state of the scene and then pause again
+            self.play()
+            for i, state in enumerate(states):
+                self.scenes[i].load_state(state, serialized=False)
```
ChengshuLi, line 1485: wow very clean! Where is the `pause` tho?

omnigibson/systems/system_base.py
```diff
@@ -208,6 +208,9 @@ def _clear(self):
         if og.sim.is_playing() and gm.ENABLE_TRANSITION_RULES:
             self.scene.transition_rule_api.prune_active_rules()
 
+        # Clear from scene active system registry
+        self.scene.system_registry.remove(self)
+
```
ChengshuLi, line 213: we should probably move all these logic into `remove_system` of scene?

omnigibson/utils/deprecated_utils.py
```diff
@@ -286,6 +290,88 @@ def get_joint_limits(
             values = self._backend_utils.convert(values, dtype="float32", device=self._device, indexed=True)
             return values
 
+    def get_joint_position_targets(
```
ChengshuLi, line 293: can you add comments to the lines of code that you modify from the original source code?

omnigibson/utils/python_utils.py
```diff
@@ -277,18 +278,21 @@ def create_class_from_registry_and_config(cls_name, cls_registry, cfg, cls_type_
     return cls(**cls_kwargs)
 
 
-def get_uuid(name, n_digits=8):
+def get_uuid(name, n_digits=8, deterministic=True):
```
ChengshuLi, line 281: maybe we should just allow `deterministic=True`? I can't really think of a case where we want `deterministic=False`.

omnigibson/utils/registry_utils.py
```diff
@@ -351,28 +434,47 @@ def _load_state(self, state):
...
     def serialize(self, state):
         # Iterate over the entire dict and flatten
-        return (
-            np.concatenate([obj.serialize(state[obj.name]) for obj in self.objects])
-            if len(self.objects) > 0
+        # We keep track of how many objects are being saved, as well as the unique identifier for each object so that
+        # the saved flattened array is agnostic to object ordering
+        n_objs = len(state)
+        state_flat = (np.concatenate([np.insert(self("name", name).serialize(state[name]), 0, getattr(self("name", name), self.hash_key)) for name in state])
+            if len(state) > 0
```
ChengshuLi, line 449: hmm not sure I fully understand, maybe we can discuss?

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+            dict: Observations from the environment
+        """
+        return self.env.observation_spec()
+
+    def process_traj_to_hdf5(self, traj_data, traj_grp_name):
+        data_grp = self.hdf5_file.require_group("data")
+        traj_grp = data_grp.create_group(traj_grp_name)
+        traj_grp.attrs["num_samples"] = len(traj_data)
+
+        data = {key: defaultdict(list) if "obs" in key else [] for key in traj_data[0]}
```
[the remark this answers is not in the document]
ChengshuLi: maybe pass in "obs" key externally and definitely check "obs" == key, not "obs" in key.

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+
+        return step_data
+
+    def process_traj_to_hdf5(self, traj_data, traj_grp_name):
+        # First pad all state values to be the same max (uniform) size
+        for step_data in traj_data:
+            state = step_data["state"]
+            padded_state = np.zeros(self.max_state_size, dtype=np.float32)
+            padded_state[: len(state)] = state
+            step_data["state"] = padded_state
```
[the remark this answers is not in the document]
ChengshuLi: maybe try np.save?

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+    """
+
+    @classmethod
+    def create_from_hdf5(
+        cls,
+        input_path,
+        output_path,
+        robot_obs_modalities,
+        external_obs_modalities,
+        n_render_iterations=5,
```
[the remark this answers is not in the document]
ChengshuLi: we should add a comment to explain why it's needed.

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+            self.current_obs, _, _, _, _ = self.env.step(action=a, n_render_iterations=self.n_render_iterations)
+
+            # If recording, record data
+            if record:
+                step_data = dict()
+                step_data["obs"] = self.current_obs
+                step_data["action"] = a
+                step_data["reward"] = r
+                step_data["terminated"] = te
+                step_data["truncated"] = tr
```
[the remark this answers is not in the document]
ChengshuLi: just call `_parse_step_data` here

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+                self.current_traj_history.append(step_data)
+
+            self.step_count += 1
+
+        if record:
+            self.flush_current_traj()
```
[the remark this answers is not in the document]
ChengshuLi: no, a new one

omnigibson/object_states/inside.py
```diff
@@ -42,10 +43,10 @@ def _get_value(self, other):
...
             raise ValueError("Cannot detect if an object is inside a cloth object.")
 
-        # First check that the inner object's position is inside the outer's AABB.
-        # Since we usually check for a small set of outer objects, this is cheap
-        aabb_lower, aabb_upper = self.obj.states[AABB].get_value()
-        inner_object_pos = (aabb_lower + aabb_upper) / 2.0
+        # For performance sake, we just check the position within the external AABB
+        # Note that this will deprecated shortly, as we'll have explicit "Inside volume"
+        # volume that we can explicitly check against in the future
+        inner_object_pos = self.obj.states[Pose].get_value()[0]
```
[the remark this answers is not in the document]
ChengshuLi: revert.

omnigibson/prims/entity_prim.py
ChengshuLi: Maybe add a comment to explain why we explicitly DO NOT save/load state for joint targets

omnigibson/scenes/scene_base.py
ChengshuLi: think about unifying import_object/remove_object/get_system/remove_system

omnigibson/utils/registry_utils.py
ChengshuLi: can we make this more readable.

tests/test_dump_load_states.py
```diff
@@ -5,43 +5,76 @@
...
+#         else:
+#             system.generate_particles(positions=[[0, 0, 1]])
+#         assert system.n_particles > 0
+#
+#     state = og.sim.dump_state(serialized=True)
+#     og.sim.load_state(state, serialized=True)
+#
+#     for system_name, system_class in SYSTEM_EXAMPLES.items():
+#         system = env.scene.get_system(system_name)
+#         system.clear()
```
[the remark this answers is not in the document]
cremebrule: whoops! not intentional. uncommented, thanks!

tests/test_dump_load_states.py
cremebrule: that's a great point -- I assume once we have the multi-env setup fully fleshed out, we'll have this, right @cgokmen ?

omnigibson/envs/data_wrapper.py
cremebrule: nope -- great catch, removed!

omnigibson/envs/data_wrapper.py
cremebrule: yes!

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+    def observation_spec(self):
+        """
+        Grab the normal environment observation_spec
+
+        Returns:
+            dict: Observations from the environment
+        """
+        return self.env.observation_spec()
+
+    def process_traj_to_hdf5(self, traj_data, traj_grp_name):
```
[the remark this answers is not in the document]
cremebrule: added docstring!

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+        """
+        Args:
+            env (Environment): The environment to wrap
+            output_path (str): path to store hdf5 data file
+            viewport_camera_path (str): prim path to the camera to use when rendering the main viewport during
+                data collection
+            only_successes (bool): Whether to only save successful episodes
+        """
+        # Store additional variables needed for optimized data collection
+        self.max_state_size = 0
```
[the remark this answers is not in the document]
cremebrule: removed from parent class

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+        self.current_transitions = dict()
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+
+        # Use asynchronous rendering for faster performance
+        lazy.carb.settings.get_settings().set_bool("/app/asyncRendering", True)
+        lazy.carb.settings.get_settings().set_bool("/app/asyncRenderingLowLatency", True)
+
+        # Disable mouse grabbing since we're only using the UI passively
+        lazy.carb.settings.get_settings().set_bool("/physics/mouseInteractionEnabled", False)
+        lazy.carb.settings.get_settings().set_bool("/physics/mouseGrab", False)
+        lazy.carb.settings.get_settings().set_bool("/physics/forceGrab", False)
+        lazy.carb.settings.get_settings().set_bool("/physics/suppressReadback", True)
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+
+        # Set the dump filter for better performance
+        env.scene.object_registry.set_dump_filter(dump_filter=lambda obj: obj.is_active)
+
+        # TODO: load filter mismatch with actual state being loaded?
```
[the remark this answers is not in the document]
cremebrule: not relevant

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: no need for now

omnigibson/envs/env_base.py
```diff
@@ -592,6 +604,7 @@ def step(self, action):
             action (gym.spaces.Dict or dict or np.array): robot actions. If a dict is specified, each entry should
                 map robot name to corresponding action. If a np.array, it should be the flattened, concatenated set
                 of actions
+            n_render_iterations (int): Number of rendering iterations to use before returning observations
```
cgokmen, line 607: Why is this an argument here and not using the render frequency? What is the purpose of having an additional n_render_iterations here? I really think this doesnt belong in this API, it's way too low-level, plus we're now changing how many times a render will happen per step?

omnigibson/envs/env_base.py
```diff
@@ -601,8 +614,17 @@ def step(self, action):
...
         """
+        # Pre-processing before stepping simulation
         self._pre_step(action)
+
+        # Step simulation
         og.sim.step()
+
+        # Render any additional times requested
+        for _ in range(n_render_iterations - 1):
+            og.sim.render()
```
cgokmen, line 625: Same question as above. Doesn't step already take care of the additional rendering?

omnigibson/object_states/inside.py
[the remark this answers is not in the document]
cgokmen: agreed re: revert. let's focus on optimizing aabb in a follow up instead.

omnigibson/object_states/temperature.py
cgokmen: +1

omnigibson/objects/object_base.py
```diff
@@ -82,8 +79,7 @@ def __init__(
         relative_prim_path = f"/{name}" if relative_prim_path is None else relative_prim_path
 
         # Store values
-        self.uuid = get_uuid(name) if uuid is None else uuid
-        assert len(str(self.uuid)) <= 8, f"UUID for this object must be at max 8-digits, got: {self.uuid}"
+        self._uuid = get_uuid(name, deterministic=True)
```
cgokmen, line 82: What is the purpose of this UUID? It's annoying to me because these UUIDs are actually *not* unique in the sense that they are duplicated across scenes just like the object name - why do we need a separate identifier that serves exactly the same purpose as the name?

omnigibson/prims/rigid_prim.py
```diff
@@ -320,16 +326,29 @@ def set_position_orientation(self, position=None, orientation=None):
...
+            assert np.isclose(
+                np.linalg.norm(ori), 1, atol=1e-3
+            ), f"{self.prim_path} orientation {ori} is not a unit quaternion."
```
cgokmen, line 351: I'm guessing it's actually the sqrt that's slow in the norm computation. Can we replace it with `np.sum(ori * ori)` instead? Should still equal 1 although the atol might need to be decreased. I'd really like to keep this assertion because it catches physx crashes right after they happen.

omnigibson/prims/rigid_prim.py
```diff
@@ -283,12 +283,15 @@ def set_linear_velocity(self, velocity):
...
+    def get_linear_velocity(self, clone=True):
         """
+        Args:
+            clone (bool): Whether to clone the internal buffer or not when grabbing data
+
         Returns:
             np.ndarray: current linear velocity of the the rigid prim. Shape (3,).
         """
-        return self._rigid_prim_view.get_linear_velocities()[0]
+        return self._rigid_prim_view.get_linear_velocities(clone=clone)[0]
```
cgokmen, line 294: Here and below, is this cloning really a major source of slowdowns? What happens when we don't clone, is the underlying tensor returned in a mutable way or something?

What I would do here is take these clones out & leave it as it was prior, even if it's slow. I will follow up your PR with one that saves and dumps everything from the physx state using a global rigid prim view and articulation view and whatnot.

omnigibson/prims/xform_prim.py
```diff
@@ -419,15 +425,15 @@ def _dump_state(self):
         # If we are in a scene, compute the scene-local transform (and save this as the world transform
         # for legacy compatibility)
         if self.scene is not None:
-            pos, ori = T.relative_pose_transform(pos, ori, *self.scene.prim.get_position_orientation())
+            pos, ori = T.mat2pose(self.scene.pose_inv @ T.pose2mat((pos, ori)))
```
cgokmen, line 428: Let's undo all of this - Frank's new PR will centralize this behavior into the "scene" mode of the pose getters.

omnigibson/robots/behavior_robot.py
```diff
@@ -363,8 +361,8 @@ def base_footprint_link(self):
         """
         return self._links[self.base_footprint_link_name]
 
-    def get_position_orientation(self):
-        return self.base_footprint_link.get_position_orientation()
+    def get_position_orientation(self, clone=True):
```
cgokmen, line 364: Lets undo clones here too. It's a tricky interface imo.

omnigibson/robots/tiago.py
```diff
@@ -679,9 +675,9 @@ def arm_workspace_range(self):
             "right": [np.deg2rad(-75), np.deg2rad(-15)],
         }
 
-    def get_position_orientation(self):
+    def get_position_orientation(self, clone=True):
         # TODO: Investigate the need for this custom behavior.
-        return self.base_footprint_link.get_position_orientation()
+        return self.base_footprint_link.get_position_orientation(clone=clone)
```
cgokmen, line 680: Remove clone

cgokmen (changes requested):
Part 1 of review, will continue after lunch. Thanks!

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+
+
+class DataPlaybackWrapper(DataWrapper):
+    """
+    An OmniGibson environment wrapper for playing back data and collecting observations.
+
+    NOTE: This assumes a DataCollectionWrapper environment has been used to collect data!
+
+    # TODO: Add params for setting camera width / height
+    """
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: done

omnigibson/scenes/scene_base.py
```diff
@@ -469,6 +526,7 @@ def _create_registry(self):
                 name="system_registry",
                 class_types=BaseSystem,
                 default_key="name",
+                hash_key="uuid",
```
cgokmen, line 529: Doesn't "name" work just as well for both of these?

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cgokmen: this hasnt been pushed

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+        transitions = json.loads(traj_grp.attrs["transitions"])
+        action = traj_grp["action"]
+        state = traj_grp["state"]
+        state_size = traj_grp["state_size"]
+        reward = traj_grp["reward"]
+        terminated = traj_grp["terminated"]
+        truncated = traj_grp["truncated"]
+
+        # Reset environment
+        og.sim.restore(scene_files=[self.scene_file])
```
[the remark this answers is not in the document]
cremebrule: solution: record init state during data collection, and then load the state at the beginning during playback so we have a consistent s0 / obs0 recorded

omnigibson/object_states/inside.py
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cgokmen: This last line is supposed to be enabling the GPU pipeline no? If you enable it with GPU dynamics doesnt it wreak havoc on all numpy inputs/outputs to tensor views?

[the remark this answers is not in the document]
cgokmen: also change is not pushed here, unresolving

omnigibson/prims/entity_prim.py
cremebrule: done

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+        lazy.carb.settings.get_settings().set_bool("/app/asyncRenderingLowLatency", True)
+
+        # Disable mouse grabbing since we're only using the UI passively
+        lazy.carb.settings.get_settings().set_bool("/physics/mouseInteractionEnabled", False)
+        lazy.carb.settings.get_settings().set_bool("/physics/mouseGrab", False)
+        lazy.carb.settings.get_settings().set_bool("/physics/forceGrab", False)
+        lazy.carb.settings.get_settings().set_bool("/physics/suppressReadback", True)
+
+        # Set the dump filter for better performance
+        env.scene.object_registry.set_dump_filter(dump_filter=lambda obj: obj.is_active)
```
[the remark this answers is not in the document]
cgokmen: This whole is_active mechanism feels very very brittle and like it can bite us down the road, especially due to its reliance on the sleep mechanism. Let's at least leave a TODO here and open a ticket to follow up on this after we have full physics state saving/loading via custom views alongside the usual dump/load state mechanism.

cgokmen (changes requested):

omnigibson/scenes/scene_base.py
cremebrule: done

omnigibson/scenes/scene_base.py
cremebrule: done

omnigibson/simulator.py
cremebrule: done

omnigibson/systems/system_base.py
cremebrule: done

omnigibson/simulator.py
cremebrule: removed outdated comment

omnigibson/utils/registry_utils.py
cremebrule: updated

tests/test_data_collection.py
```diff
@@ -0,0 +1,134 @@
...
+        water = env.scene.get_system("water")
+        pos = np.random.random((10, 3)) * 10.0
+        water.generate_particles(positions=pos)
+
+        # Take a few more steps
+        for _ in range(5):
+            env.step(env.robots[0].action_space.sample())
+
+        # Clear the system
+        water.clear()
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/data_wrapper.py
```diff
@@ -0,0 +1,525 @@
...
+        # Store additional variables
+        self.n_render_iterations = n_render_iterations
+
+        # Run super
+        super().__init__(env=env, output_path=output_path, only_successes=only_successes)
+
+    def _parse_step_data(self, action, obs, reward, terminated, truncated, info):
+        # Store action, obs, reward, terminated, truncated, info
+        step_data = dict()
+        step_data["obs"] = self.current_obs
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/env_base.py
```diff
@@ -68,24 +68,34 @@ def __init__(self, configs, in_vec_env=False):
...
+        sim_step_dt = 1.0 / self.action_frequency
         viewer_width = self.render_config["viewer_width"]
         viewer_height = self.render_config["viewer_height"]
+
+        # Make sure action frequency is divisible by rendering frequency
+        assert self.rendering_frequency >= self.action_frequency, \
+            f"Action frequency ({self.action_frequency} Hz) cannot be greater than rendering frequency ({self.rendering_frequency} Hz)!"
+
+        assert self.rendering_frequency % self.action_frequency == 0.0, \
+            f"Rendering frequency ({self.rendering_frequency} Hz) must be divisible by action frequency ({self.action_frequency} Hz)!"
```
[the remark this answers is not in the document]
cremebrule: done

omnigibson/envs/env_base.py
cremebrule: cf. eric's comment above. It's a two-fold issue here:

1. Omni's real-time raytracing, for whatever reason, is both laggy and jittery. By laggy I mean that the actual rendered frame is always a few frames BEHIND what the actual physical state of the world is. So if you teleport the robot you need to render multiple times before you actually observe the change in the GUI / observation.
2. We need the rendered frame to be accurate _before_ we grab observations. This means that we can't simply do additional renders after env.step() gets called, because at the point the observations have already been recorded and they're not exactly accurate with respect to the most up-to-date physical state of the world (assuming we've only rendered once).

omnigibson/envs/env_base.py
cremebrule: No, it's separate issues:

- We want multiple rendering steps per action step merely for the sake of human comfort. As you mentioned, if headless mode is enabled, there's no reason for rendering_dt =/= sim_step_dt
- We want our observations to be accurate wrt to the most up-to-date representation of the world. This requires multiple renders to occur after **all** physics steps have been taken, but **before** observations are aggregated.

omnigibson/objects/object_base.py
cremebrule: it's for the sake of serializing the state, which, by construction, records the state as a flattened 1D numpy array. So we can't directly use strings.

omnigibson/prims/rigid_prim.py
cremebrule: Just timed it. Changing to squared norm doesn't actually impact it that much. Looks like the most expensive part is actually the `np.isclose()` check.

cremebrule: Do we have an idea about when the quat is invalid? i.e: are all values nans, or all 0s, or something? maybe that's a faster check

cremebrule: e.g: `any(quat)` --> this is 2 orders of magnitude faster than `np.isclose()`

omnigibson/prims/rigid_prim.py
cremebrule: Yeah...not sure why. I'm guessing it's because we keep cloning different tensors many many times.

For now, the behavior is pretty much consistent all across the repo (`clone=True` by default), except in the context of dumping state. Can we keep this for now for the sake of demo collection, and simply remove the flag once you have your PR ready? I'm guessing the follow-up PR will be nontrivial, so it'd be ideal to not block demo collection with this.

omnigibson/utils/python_utils.py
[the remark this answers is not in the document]
cremebrule: see comment above

omnigibson/utils/registry_utils.py
```diff
@@ -351,28 +434,47 @@ def _load_state(self, state):
         # the state might contain additional information about objects that are NOT in the scene. For both cases, state
         # loading will be skipped.
         for obj in self.objects:
-            if obj.name not in state:
-                log.warning(f"Object '{obj.name}' is not in the state dict to load from. Skip loading its state.")
-                continue
-            obj.load_state(state[obj.name], serialized=False)
+            if self._load_filter(obj):
+                if obj.name not in state:
+                    log.warning(f"Object '{obj.name}' is not in the state dict to load from. Skip loading its state.")
```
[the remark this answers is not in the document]
cremebrule: Yep, not a problem, done

omnigibson/robots/behavior_robot.py
cremebrule: resolved in person -- keep for now, revert later

omnigibson/robots/tiago.py
cremebrule: resolved in person -- keep for now, revert later

omnigibson/prims/rigid_prim.py
cremebrule: resolved in person -- keep for now, revert later

omnigibson/envs/env_base.py
cremebrule: resolved in person

omnigibson/envs/env_base.py
cremebrule: resolved in person

omnigibson/simulator.py
```diff
@@ -359,6 +380,9 @@ def __init__(
             CollisionAPI.add_group_filter(col_group="structures", filter_group="fixed_base_nonroot_links")
             CollisionAPI.add_group_filter(col_group="structures", filter_group="fixed_base_root_links")
 
+            # Store stage ID
+            self._stage_id = lazy.pxr.UsdUtils.StageCache.Get().GetId(self.stage).ToLongInt()
+
```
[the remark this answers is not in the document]
cremebrule: as it turns out, manually calling stage_id over and over is actually nontrivial!

omnigibson/simulator.py
```diff
@@ -1149,6 +1213,18 @@ def callback() --> None
...
+            Adds a function @callback, referenced by @name, to be executed at the beginning of every sim.step() call
+
+            Args:
+                name (str): Name of the callback
+                callback (function): Callback function. Function signature is expected to be:
+
+                    def callback() --> None
+            """
+            self._callbacks_on_pre_sim_step[name] = callback
+
```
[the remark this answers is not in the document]
cremebrule: great point! done

omnigibson/simulator.py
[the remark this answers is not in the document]
cgokmen: ah this is the stage ID we pass into the sim functions I see. sorry, resolved

omnigibson/objects/object_base.py
cgokmen: I'm opening a follow-up issue for renaming UUIDs into name hashes. I shouldnt hold up this PR on that. See #831.

omnigibson/prims/rigid_prim.py
cgokmen: Can we just change this to np.sum(ori  * ori) - 1 < 1e-3 ?

omnigibson/tasks/task_base.py
```diff
@@ -118,12 +119,40 @@ def load(self, env):
         self._load(env=env)
 
         # Load the obs space dim
+        og.sim.play()
         obs = self.get_obs(env=env, flatten_low_dim=True)
         self._low_dim_obs_dim = len(obs["low_dim"]) if "low_dim" in obs else 0
+        og.sim.stop()
```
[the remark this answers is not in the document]
cgokmen: We shouldn't need to get the observations here though, we just need the dimensions. Perhaps a clever trick is feasible. This code was previously working in vector env and this change acutely has the effect of changing load time from 10-15 mins to 2-3 hours on 50+ envs.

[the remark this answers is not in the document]
cgokmen: So my point is you shouldn't have to play in the first place.

omnigibson/prims/rigid_prim.py
cremebrule: yep, this works!

omnigibson/prims/xform_prim.py
cremebrule: what behavior will be centralized? the caching mechanism or the computation? This change was done to nontrivially improve the state saving, since every object computes their pose in this manner (so naively we have N calls to the scene.prim.get_position_orientation). 

The refactor is definitely good, but I would prefer to leave this change here so that the caching mechanism is not forgotten about during Frank's follow up PR (not to mention avoiding the inverse transformation in this call too)

omnigibson/tasks/task_base.py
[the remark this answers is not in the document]
cremebrule: Ahhh I see. Yes, just checked the code and there's a nice workaround! Will fix this.

omnigibson/scenes/scene_base.py
```diff
@@ -219,12 +248,20 @@ def prebuild(self):
             str: Path to the prebuilt USD file
         """
         # Prebuild and cache the scene USD using the objects
-        if self.scene_file not in PREBUILT_USDS:
+        if isinstance(self.scene_file, str):
+            scene_file_path = self.scene_file
+        else:
+            # The scene file is a dict, so write it to disk directly
+            decrypted_fd, scene_file_path = tempfile.mkstemp("scene_file.json", dir=og.tempdir)
```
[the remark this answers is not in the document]
cremebrule: Can we just use the current (unique) timestamp or is that not sufficient?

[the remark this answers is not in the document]
cremebrule: oh nvm i see -- you want the dict --> file name mapping to be deterministic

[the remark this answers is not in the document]
cremebrule: ok: I'll do the following:

```python
dic_str = json.dumps(dict)
uuid = get_uuid(dic_str, deterministic=True)
```

omnigibson/scenes/scene_base.py
```diff
@@ -668,18 +705,41 @@ def is_fluid_system(self, system_name):
...
-        # If system_name is not in REGISTERED_SYSTEMS, create from metadata
-        system = self.system_registry("name", system_name)
-        assert system is not None, f"System {system_name} not in system registry."
+        assert system_name in self._available_systems, f"System {system_name} is not a valid system name"
+        # If system is not initialized, initialize and add it to our registry
+        system = self._available_systems[system_name]
         if not system.initialized and force_init:
             system.initialize(scene=self)
+            self.system_registry.add(system)
         return system
```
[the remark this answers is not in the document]
cremebrule: updated!

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: pushed

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: pushed. tested with GPU dynamics and seems to work

omnigibson/envs/data_wrapper.py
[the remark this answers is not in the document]
cremebrule: done!

omnigibson/prims/rigid_prim.py
cremebrule: actually this doesn't work, need to also check for negative side too. added that as well.

omnigibson/prims/xform_prim.py
cgokmen: OK sounds good lets leave it in and merge. We'll unify in his PR

Bailey-24:
hello, thanks  for your great work.
I can ran successfully in tests/test_data_colloection.py, but I can't use it with the teleopeation system.
I change the code in the following, but it stuck, I can't control the robot by keyboard or quest pro.
How to solve, Thank you.

```python
ROBOTS = {
    "FrankaPanda": "Franka Emika Panda (default)",
    "Fetch": "Mobile robot with one arm",
    # "Tiago": "Mobile robot with two arms",
}
TELEOP_METHOD = {
    "keyboard": "Keyboard (default)",
    "oculus": "Oculus Quest",
}


def test_data_collect_and_playback():
    from telemoma.configs.base_config import teleop_config
    from telemoma.utils.camera_utils import RealSenseCamera

    from omnigibson.utils.teleop_utils import TeleopSystem
    robot_name = choose_from_options(options=ROBOTS, name="robot")
    arm_teleop_method = choose_from_options(options=TELEOP_METHOD, name="robot arm teleop method")
    if robot_name != "FrankaPanda":
        base_teleop_method = choose_from_options(options=TELEOP_METHOD, name="robot base teleop method")
    else:
        base_teleop_method = "keyboard"  # Dummy value since FrankaPanda does not have a base
    # Generate teleop config
    teleop_config.arm_left_controller = arm_teleop_method
    teleop_config.arm_right_controller = arm_teleop_method
    teleop_config.base_controller = base_teleop_method
    teleop_config.interface_kwargs["keyboard"] = {"arm_speed_scaledown": 0.04}
    teleop_config.interface_kwargs["spacemouse"] = {"arm_speed_scaledown": 0.04}
    if arm_teleop_method == "vision" or base_teleop_method == "vision":
        teleop_config.interface_kwargs["vision"] = {"camera": RealSenseCamera()}

    cfg = {
        "env": {
            "external_sensors": [],
        },
        "scene": {
            "type": "InteractiveTraversableScene",
            "scene_model": "Rs_int",
            "load_object_categories": ["floors", "breakfast_table"],
        },
        "robots": [
            {
                "type": "Fetch",
                "obs_modalities": [],
            }
        ],
        # Task kwargs
        "task": {
            "type": "BehaviorTask",
            # BehaviorTask-specific
            "activity_name": "assembling_gift_baskets",
            "online_object_sampling": True,
        },
    }

    if og.sim is None:
        # Make sure GPU dynamics are enabled (GPU dynamics needed for cloth)
        gm.ENABLE_OBJECT_STATES = True
        gm.USE_GPU_DYNAMICS = True
        gm.ENABLE_TRANSITION_RULES = False
    else:
        # Make sure sim is stopped
        og.sim.stop()

    vr_data = "/home/pi/Documents/OmniGibson/vr_data"
    collect_hdf5_path = os.path.join(vr_data, "test_data_collection.hdf5")
    playback_hdf5_path = os.path.join(vr_data, "test_data_playback.hdf5")
    # Create the environment (wrapped as a DataCollection env)
    env = og.Environment(configs=cfg)
    robot = env.robots[0]
    env = DataCollectionWrapper(
        env=env,
        output_path=collect_hdf5_path,
        only_successes=False,
    )
    for _ in range(2):
        env.reset()

    # Initialize teleoperation system
    teleop_sys = TeleopSystem(config=teleop_config, robot=robot, show_control_marker=True)
    teleop_sys.start()
    
    # Record 2 episodes
    for i in range(2000):
        action = teleop_sys.get_action(teleop_sys.get_obs())
        env.step(action)

    # Save this data
    env.save_data()
    print("Data collection done")
    # Clear the sim
    og.clear(
        physics_dt=0.001,
        rendering_dt=0.001,
        sim_step_dt=0.001,
    )
```

PR #817, opened 2024-07-26
Fix two controller issues

cgokmen:
Multifinger has issues in smooth mode, also the grasp checking makes a strong assumption.

### DIFF
--- a/omnigibson/controllers/multi_finger_gripper_controller.py
+++ b/omnigibson/controllers/multi_finger_gripper_controller.py
@@ -176,11 +176,8 @@ class MultiFingerGripperController(GripperController):
                     else self._closed_qpos
                 )
         else:
-            # Use continuous signal
-            if len(target) == 1:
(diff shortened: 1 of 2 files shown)


omnigibson/controllers/multi_finger_gripper_controller.py
```diff
@@ -177,7 +177,10 @@ def compute_control(self, goal_dict, control_dict):
                 )
         else:
             # Use continuous signal
-            u = target
+            if len(target) == 1:
```
cremebrule, line 180: nit, make one liner

cremebrule (approved):
small comment, after resolving feel free to merge!

omnigibson/controllers/multi_finger_gripper_controller.py
cgokmen: Done

PR #819, opened 2024-07-29
Add robot load and drive test

hang-yin:


tests/test_robot_states_flatcache.py
```diff
@@ -114,3 +115,86 @@ def camera_pose_test(flatcache):
...
+        # Set viewer in front facing robot
+        og.sim.viewer_camera.set_position_orientation(
+            position=np.array([2.69918369, -3.63686664, 4.57894564]),
+            orientation=np.array([0.39592411, 0.1348514, 0.29286304, 0.85982]),
+        )
+
+        # If this is a locomotion robot, we want to test driving
+        if isinstance(robot, LocomotionRobot):
+            initial_base_pos = robot.get_position()
+            robot.move_forward(0.1)
```
ChengshuLi, line 163: I felt like very people will use `move_forward`. Maybe we should call `env.step(action)` with `base_action_idx` instead?

ChengshuLi (approved):

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1414,24 +1418,17 @@ def _empty_action(self):
...
-                control_idx = self.robot.controller_action_idx["arm_" + self.arm]
-                action[control_idx[3:]] = current_ori
+            no_op_goal = controller.compute_no_op_goal(self.robot.get_control_dict())
+
+            if self.robot._controller_config[name]["name"] == "InverseKinematicsController":
+                # assert (
+                #     self.robot._controller_config["arm_" + self.arm]["mode"] == "pose_absolute_ori"
+                # ), "Controller must be in pose_absolute_ori mode"
+                # convert quaternion to axis-angle representation for control input
+                no_op_goal["target_quat"] = T.quat2axisangle(no_op_goal["target_quat"])
```
ChengshuLi, line 1429: why is it needed to convert quaternion to axis-angle representation for control input?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1642,11 +1646,38 @@ def _navigate_to_pose_direct(self, pose_2d, low_precision=False):
...
+                    elif base_action_size == 2:
+                        # [left, right]
+                        direction_vec = body_target_pose[0][:2]
+                        distance = np.linalg.norm(direction_vec)
+                        # Move forward
+                        linear_x = min(distance, m.KP_LIN_VEL[type(self.robot)])
+                        wheel_velocities = self._calculate_wheel_velocities(linear_x, 0.0, 2)
+                        action[self.robot.controller_action_idx["base"]] = wheel_velocities
+                    else:
+                        raise ValueError("Invalid base action size")
```
ChengshuLi, line 1677: for these three different actions sizes (2, 3, 4), do they necessarily correspond to the comments? e.g. `[x,y,theta]`? Seems like a strong assumption.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1661,6 +1692,26 @@ def _navigate_to_pose_direct(self, pose_2d, low_precision=False):
         # Rotate in place to final orientation once at location
         yield from self._rotate_in_place(end_pose, angle_threshold=angle_threshold)
 
+    def _calculate_wheel_velocities(self, linear_x, angular_z, wheel_count=2):
+        # Calculate individual wheel velocities based on desired motion
```
ChengshuLi, line 1696: diff drive already have this piece of code implemented? can we reuse?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1682,9 +1733,27 @@ def _rotate_in_place(self, end_pose, angle_threshold=m.DEFAULT_ANGLE_THRESHOLD):
...
+            elif base_action.size == 4:
+                wheel_velocities = self._calculate_wheel_velocities(0.0, ang_vel, 4)
+                base_action[:] = wheel_velocities
+            elif base_action.size == 3:
+                base_action[2] = ang_vel
+            elif base_action.size == 2:
+                wheel_velocities = self._calculate_wheel_velocities(0.0, ang_vel, 2)
+                base_action[:] = wheel_velocities
+            else:
+                raise ValueError("Invalid base action size")
```
ChengshuLi, line 1755: should probably add some comments for each of these cases as well.

omnigibson/objects/object_base.py
```diff
@@ -192,6 +192,9 @@ def _post_load(self):
             self.root_prim.RemoveAPI(lazy.pxr.UsdPhysics.ArticulationRootAPI)
             self.root_prim.RemoveAPI(lazy.pxr.PhysxSchema.PhysxArticulationAPI)
 
+        if og.sim.is_playing:
+            log.warning("Cannot set articulation root API while simulation is playing!")
```
ChengshuLi, line 196: hmm interesting.. are you sure? Previously I was able to import new object while sim is playing. Are you suggesting the articulation root API won't work?

tests/test_robot_states_flatcache.py
```diff
@@ -114,3 +116,88 @@ def camera_pose_test(flatcache):
...
+        robot = robot_cls(
+            name=robot_name,
+            obs_modalities=[],
+        )
+        og.sim.stop()
+        env.scene.add_object(robot)
+        og.sim.play()
+
+        # At least one step is always needed while sim is playing for any imported object to be fully initialized
+        og.sim.play()
```
ChengshuLi, line 150: why play twice?

tests/test_robot_states_flatcache.py
```diff
@@ -114,3 +116,88 @@ def camera_pose_test(flatcache):
...
+        # Make sure sim is stopped
+        og.sim.stop()
+
+    config = {
+        "scene": {
+            "type": "Scene",
+        },
+    }
+
+    env = og.Environment(configs=config)
```
ChengshuLi, line 137: I think you should add `og.sim.stop()` here, and then remove the `og.sim.stop()` in line 145.

omnigibson/objects/object_base.py
hang-yin: This doesn't work all the time. It depends on how the articulation root is annotated originally. Essentially the reason why we added this warning is we can only load some robots while sim is playing but not the others.

hang-yin: This doesn't work all the time. It depends on how the articulation root is annotated originally. Essentially the reason why we added this warning is we can only load some robots while sim is playing but not the others.

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: I suppose this is a thing with IK controller `pose_absolute_ori` mode. It seems to me that it takes axis angles as action. This might help:
```
"pose_absolute_ori": 6,  # 6DOF (dx,dy,dz,ax,ay,az) control over pose, where the orientation is given in absolute axis-angle coordinates
```

ChengshuLi: move the logic into `InverseKinematicsController`

ChengshuLi (approved):

omnigibson/objects/object_base.py
cgokmen: @ChengshuLi this causes a very annoying problem: when you import a USD while the simulation is playing, the object gets loaded into PhysX before post_load can run, meaning it gets loaded with the articulation root API still on w/e prim it was on in the USD regardless of whether the USD is fixed or not, and you can't fix it without stopping (e.g. even if you move the articulation root API on the USD, physx remains unaware). We can hack this by removing the articulation root API from all of our USDs, then at load time we can pre-load it using the Usd library instead of loading it directly into Isaac, add the articulation root, and load that one into Isaac instead - I'll do it with the next asset release.

omnigibson/action_primitives/starter_semantic_action_primitives.py
cremebrule: both OSC and IK take in axis angle, so their action spaces were unified in this way.

It's a bit more efficient than using quaternions, since it's 3-dim vs. 4-dim

cgokmen: @cremebrule why axis angle instead of euler angles? iGibson uses to do euler I think and so does the joint controlled base on the tiago etc. Maybe not the best place to discuss this but this is the first time I'm hearing we use axis angle representation anywhere so just curious

PR #821, opened 2024-07-30
Bbox test

hang-yin:


omnigibson/sensors/vision_sensor.py
```diff
@@ -447,18 +459,22 @@ def _register_instance(self, instance_name, id=False):
...
+            id_to_labels (dict): Dictionary of semantic IDs to class labels
         Returns:
             list of dict: Remapped list of bounding boxes
+            dict: Remapped id_to_labels dictionary
         """
+        replicator_mapping = self._preprocess_semantic_labels(id_to_labels)
         for bbox in bboxes:
-            bbox["semanticId"] = VisionSensor.SEMANTIC_REMAPPER.remap_bbox(bbox["semanticId"], self._scene)
-        return bboxes
+            bbox["semanticId"] = semantic_class_name_to_id()[replicator_mapping[bbox["semanticId"]]]
```
ChengshuLi, line 475: seems wasteful to call `semantic_class_name_to_id()` every time? Why not use another remapper instead?

hang-yin: Potentially we could! Here are some counterarguments though:
1. We don't really have many bounding boxes per observation
2. This `semantic_class_name_to_id()` is actually a cached function that returns a dictionary, which means each look up would be constant time without function call overhead. Compared to mapping 128x128/624x624/1024x1024 images, this is probably nothing.

ChengshuLi: sounds good

ChengshuLi (approved):

PR #823, opened 2024-07-30
Add benchmark metrics to simulator

yyf20001230:
initial commit to benchmark metrics


omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+            return 0.0, {}
+
+        work_metric = 0.0
+        for linkname, posrot in new_state_cache.items():
+            work_metric += self.calculate_displacement(posrot, self.state_cache[linkname]) * self.link_masses[linkname]
+
+        if self.measure_work:
+            self._reward = 0
+        if not self.measure_work:
+            self.state_cache = new_state_cache
```
cgokmen, line 51: add comment here - we update the "state cache" only when we are in the energy metric (e.g, "distance" mode). For "displacement" mode we always want to be comparing against the initial state.

omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+
+            for obj in env.scene.objects:
+                for link_name, link in obj._links.items():
+                    self.link_masses[link_name] = link.mass
+            return 0.0, {}
+
+        work_metric = 0.0
+        for linkname, posrot in new_state_cache.items():
+            work_metric += self.calculate_displacement(posrot, self.state_cache[linkname]) * self.link_masses[linkname]
```
cgokmen, line 46: Leave TODO comment here reminding us to handle object addition and removal

omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Args:
+        measure_work: If true, measure beginning and end delta rather than step by step delta
+    """
+
+    def __init__(self, measure_work=False):
+        # Run super
+        super().__init__()
+        self._reward = 0
+        self.initialized = False
```
cgokmen, line 20: Remove this. Everywhere this variable is used, check if state_cache == None instead.

omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+        self.state_cache = {}
```
cgokmen, line 21: Start this as None

omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+                for link_name, link in obj._links.items():
+                    self.link_masses[link_name] = link.mass
+            return 0.0, {}
+
+        work_metric = 0.0
+        for linkname, posrot in new_state_cache.items():
+            work_metric += self.calculate_displacement(posrot, self.state_cache[linkname]) * self.link_masses[linkname]
+
+        if self.measure_work:
+            self._reward = 0
```
cgokmen, line 49: Remove these two lines

omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+        if not self.measure_work:
+            self.state_cache = new_state_cache
+
+        self._reward += work_metric
```
cgokmen, line 53: Here do `self._reward = work_metric if self.measure_work else (self._reward + work_metric)`

omnigibson/reward_functions/energy_metric.py
```diff
@@ -0,0 +1,59 @@
...
+        return self._reward, {}
+
+    def reset(self, task, env):
```
cgokmen, line 56: Here reset reward to 0

omnigibson/tasks/task_base.py
```diff
@@ -193,6 +212,9 @@ def reset(self, env):
             termination_condition.reset(self, env)
         for reward_function in self._reward_functions.values():
             reward_function.reset(self, env)
+        for metric_function in self._metric_functions.values():
+            metric_function.reset(self, env)
+            metric_function._reward = 0.0
```
cgokmen, line 217: Instead of putting this here, make sure *every* metric function sets reward=0 when reset. Then remove this

cgokmen (changes requested):

omnigibson/reward_functions/energy_metric.py
yyf20001230: shouldn't this be *not self.mesaure_work" here, as we only get the current work_metric if measuring energy

omnigibson/envs/env_base.py
```diff
@@ -533,6 +531,11 @@ def _populate_info(self, info):
 
     def _pre_step(self, action):
         """Apply the pre-sim-step part of an environment step, i.e. apply the robot actions."""
+
+        # record the start time
```
ChengshuLi, line 535: duplicate

omnigibson/envs/env_base.py
```diff
@@ -533,6 +531,11 @@ def _populate_info(self, info):
...
+        # record the start time of the simulation step in the beginning of the step
+        self._cur_sim_start_ts = time.perf_counter()
```
ChengshuLi, line 537: should we move these logic into the metrics themselves? e.g. the env class just call a bunch of pre_step for each of the metrics, something like this:
```
for metric in self.task.metrics:
    metric.pre_step()
```

omnigibson/metrics/energy_metric.py
```diff
@@ -0,0 +1,64 @@
...
+
+class EnergyMetric(BaseMetric):
+    """
+    Energy Metric
+
+    Measures displacement * mass for every link
+
+    Args:
+        measure_work: If true, measure beginning and end delta rather than step by step delta
+    """
```
ChengshuLi, line 14: I think we should just create a new WorkMetric instead of combining both into the same class with a boolean flag

omnigibson/metrics/energy_metric.py
```diff
@@ -0,0 +1,64 @@
...
+                    self.link_masses[link_name] = link.mass
+
+            return 0.0
+
+        # calculate the energy spent from the previous state to the current state
+        work_metric = 0.0
+        for linkname, posrot in new_state_cache.items():
+
+            # TODO: this computation is very slow, consider using a more efficient method
+            # TODO: this method needs to be updated to account for object addition and removal
```
ChengshuLi, line 50: can we actually finish these two TODOs?

omnigibson/metrics/energy_metric.py
```diff
@@ -0,0 +1,64 @@
...
+            posrot2 = self.state_cache[linkname]
+            work_metric += np.linalg.norm(posrot[0] - posrot2[0]) * self.link_masses[linkname]
```
ChengshuLi, line 52: Maybe we should consider rotation changes as well, e.g. a weighted sum of translation changes and rotation changes, where the weights are hyperparams that can be configured.

omnigibson/metrics/energy_metric.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # if measuring energy, update the state cache
+        if not self.measure_work:
+            self.state_cache = new_state_cache
+
+        # update the metric accordingly, either set the work done (measuring work) or add it to previous energy spent (measuring energy)
+        self._metric = work_metric if self.measure_work else (self._metric + work_metric)
+        return self._metric
```
ChengshuLi, line 60: If measure_work, we should only compute the metrics once, at the very end of the episode. One way to implement this is to add a `on_episode_end` function to the metrics base class, and then call that in the task or env class.

omnigibson/metrics/metrics_base.py
```diff
@@ -0,0 +1,59 @@
...
+# from omnigibson.utils.python_utils import Registerable, classproperty
+class BaseMetric:
+    """
+    Base Metric class
+    Metric-specific reset and step methods are implemented in subclasses
+    """
+
+    def __init__(self):
+        # Store internal vars that will be filled in at runtime
+        self._metric = 0
```
ChengshuLi, line 14: is metric necessarily a single number?

omnigibson/termination_conditions/predicate_goal.py
```diff
@@ -44,3 +44,12 @@ def goal_status(self):
...
         """
         return self._goal_status
+
+    @property
+    def partial_success(self):
+        """
+        Returns:
+            bool: returns true if partial success is supported, false otherwise
+        """
+        assert self._done is not None, "At least one step() must occur before partial_success can be calculated!"
```
ChengshuLi, line 54: You are not computing partial success here though, only whether this termination condition supports partial success or not. Do you still need to require `done is not None`?

omnigibson/tasks/task_base.py
```diff
@@ -261,6 +284,29 @@ def _step_reward(self, env, action, info=None):
...
+
+        Args:
+            env (Environment): Environment instance
+            action (n-array): 1D flattened array of actions executed by all agents in the environment
+
+        Returns:
+            - the break down of the metric scores and the metric info
+        """
+
+        # We'll also store individual reward split
```
ChengshuLi, line 299: wrong comment

omnigibson/tasks/task_base.py
```diff
@@ -261,6 +284,29 @@ def _step_reward(self, env, action, info=None):
...
+        breakdown_dict = dict()
+
+        for metric_name, metric_function in self._metric_functions.items():
+            metric = metric_function.step(self, env, action)
+            breakdown_dict[metric_name] = metric
+
+        # TODO: metric score is currently summed, work on the global coefficients
```
ChengshuLi, line 306: finish this TODO as well

omnigibson/tasks/task_base.py
```diff
@@ -144,6 +147,23 @@ def _create_reward_functions(self):
...
+        """
+        metrics = dict()
+
+        metrics["steps"] = StepMetric()
+        metrics["task_success"] = TaskSuccessMetric()
+        metrics["wall_time"] = WallTimeMetric()
+        metrics["energy"] = EnergyMetric()
+        metrics["work"] = EnergyMetric(measure_work=True)
+
+        return metrics
```
ChengshuLi, line 165: these should be configured per task, not in the task base class.

omnigibson/objects/controllable_object.py
```diff
@@ -576,7 +576,7 @@ def get_control_dict(self):
         # TODO: Move gravity force computation dummy to this class instead of BaseRobot
         fcns["gravity_force"] = lambda: (
             ControllableObjectViewAPI.get_generalized_gravity_forces(self.articulation_root_path)
-            if self.fixed_base
+            if self.fixed_base or self._dummy is None
```
ChengshuLi, line 579: why is this line needed?

omnigibson/tasks/behavior_task.py
```diff
@@ -380,7 +377,9 @@ def _get_obs(self, env):
         low_dim_obs = dict()
 
         # Batch rpy calculations for much better efficiency
-        objs_exist = {obj: obj.exists for obj in self.object_scope.values() if not obj.is_system}
+        objs_exist = {
+            obj: obj.exists for obj in self.object_scope.values() if not obj.is_system and obj.states[Pose]._initialized
+        }
```
ChengshuLi, line 382: why is this needed?

tests/test_metric_functions.py
```diff
@@ -0,0 +1,95 @@
...
+    action = env.action_space.sample()
+    state, reward, terminated, truncated, info = env.step(action)
+    metrics = info["metrics"]
+
+    assert isinstance(metrics, dict)
+
+    env.reset()
+
+    # perform a step with no action
+    action = OrderedDict([("robot0", [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])])
```
ChengshuLi, line 47: do `action = np.zeros_like(env.action_space.sample())` instead

tests/test_metric_functions.py
```diff
@@ -0,0 +1,95 @@
...
+    metrics = info["metrics"]
+
+    assert metrics["steps"] == 0, "Step metric was not reset"
+    assert metrics["task_success"] == 0, "Task success metric was not reset"
+    assert metrics["wall_time"] == 0, "Wall time metric was not reset"
+    assert metrics["energy"] == 0, "Energy metric was not reset"
+    assert metrics["work"] == 0, "Work metric was not reset"
+
+
+def test_behavior_task_work_metric():
```
ChengshuLi, line 59: can we add tests for all the metrics, e.g. wall time, energy, steps, etc.?

yyf20001230: env.step() 
number of time step * action_time_step

omnigibson/objects/controllable_object.py
yyf20001230: I believe that self._dummy is None should be there to account for the fact that dummy is not loaded at this stage of the simulator initialization. However, dummy will be loaded later, after the gravity force computation. 
I also resolved this TODO, moving the dummy initialization from the BaseRobot to ControllableObject.

omnigibson/metrics/energy_metric.py
yyf20001230: The work done by rotation should be W = torque * rotation. Without knowing torque, I added a rough calculation of the work/energy metric. I am happy to discuss this calculation more specifically with you tomorrow.
```
position, orientation = T.pose_transform(posrot[0], posrot[1], init_posrot[0], init_posrot[1])
work_metric += np.linalg.norm(orientation) * self.link_masses[linkname] * self.metric_config["rotation"]
```

PR #829, opened 2024-08-02
isaacsim new version support and windows installation fix

yyf20001230:
4.1.0 support and windows installation fix


cgokmen (changes requested):
Can you also change both setup.sh and setup.bat's string matching to match isaac* instead of isaac-sim? Ping me for re-review. Thanks!

ChengshuLi (approved):

PR #836, opened 2024-08-15
Support multiple agent in scene graph

hang-yin:
Fix this: https://github.com/StanfordVL/OmniGibson/issues/800

### DIFF
--- a/omnigibson/scene_graphs/graph_builder.py
+++ b/omnigibson/scene_graphs/graph_builder.py
@@ -9,6 +9,7 @@ from PIL import Image
 from omnigibson import object_states
 from omnigibson.object_states.factory import get_state_name
 from omnigibson.object_states.object_state_base import AbsoluteObjectState, BooleanStateMixin, RelativeObjectState
+from omnigibson.robots import BaseRobot
 from omnigibson.sensors import VisionSensor
 from omnigibson.utils import transform_utils as T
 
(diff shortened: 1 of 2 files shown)


omnigibson/scene_graphs/graph_builder.py
```diff
@@ -296,10 +313,9 @@ def _draw_graph():
...
     if show_window:
-        import cv2
-
-        cv_img = cv2.cvtColor(img, cv2.COLOR_RGB2BGR)
-        cv2.imshow("SceneGraph", cv_img)
-        cv2.waitKey(0)
+        plt.imshow(img)
+        plt.title("SceneGraph")
+        plt.axis("off")
+        plt.show()
```
cgokmen, line 319: There is a difference here: cv2 will update repeatedly while plt will hang. I don't think we want it to hang here. Also the above comment is outdated.

hang-yin: Note: calling cv2.imshow gives us - 
```
cv2.error: OpenCV(4.9.0) /io/opencv/modules/highgui/src/window.cpp:1272: error: (-2:Unspecified error) The function is not implemented. Rebuild the library with Windows, GTK+ 2.x or Cocoa support. If you are on Ubuntu or Debian, install libgtk2.0-dev and pkg-config, then re-run cmake or configure script in function 'cvShowImage'
```
We could potentially use our own version of opencv but we would need to uninstall Omniverse's opencv.

PR #853, opened 2024-08-21
[WIP] Always enable GPU dynamics & Flatcache + support for both CPU & GPU pipeline

hang-yin:
- [ ] Enable full GPU dynamics & GPU pipeline
- [ ] Cloth view implementation
- [ ] Always enable GPU dynamics and Flatcache
- [ ] Minimize gpu-cpu sync for speed
  - [ ] Control
  - [ ] Object states
  - [ ] Transition rules
- [ ] General bug fixes
- [ ] Run all tests with both cpu & gpu pipeline modes
- [ ] Robustify `test_toggle_on` with ik primitive
- [ ] Fix property setters (e.g. link.mass = xxx doesn't work now because it expects a torch tensor instead of a list)
- [ ] Fix transform utils quat_distance
- [ ] Fix set_linear_velocities and set_angular_velocities with gpu pipeline


cgokmen:
@hang-yin follow up on forum & wait on PhysX fix, OR go with the beyond hacky fix of reading raw memory

PR #855, opened 2024-08-23
Disable gravity compensation

cgokmen:
Here we disable gravity compensation but leave it as an option for fixed-base robots. It is fully removed for floating base robots.


omnigibson/controllers/joint_controller.py
```diff
@@ -42,6 +46,8 @@ def __init__(
         kp=None,
         damping_ratio=None,
         use_impedances=False,
+        use_gravity_compensation=False,
+        use_cc_compensation=False,
```
cremebrule, line 50: i think we still need CC compensation (And this value is accurate IIRC) -- so this should be True by default

omnigibson/controllers/joint_controller.py
```diff
@@ -74,6 +80,10 @@ def __init__(
                 damping ratio applied to the joint controller. If None, a default value will be used.
             use_impedances (bool): If True, will use impedances via the mass matrix to modify the desired efforts
                 applied
+            use_gravity_compensation (bool): If True, will add gravity compensation to the computed efforts. This is
+                an experimental feature that only works on fixed base robots. We do not recommend enabling this.
+            use_cc_compensation (bool): If True, will add Coriolis / centrifugal compensation to the computed efforts
+                This is an experimental feature. We do not recommend enabling this.
```
cremebrule, line 86: fix -- i think we should default to including this

omnigibson/controllers/joint_controller.py
```diff
@@ -99,6 +109,20 @@ def __init__(
...
+        # Warn the user about gravity compensation and Coriolis / centrifugal compensation being experimental.
+        if self._use_gravity_compensation:
+            log.warning(
+                "JointController is using gravity compensation. This is an experimental feature that only works on "
+                "fixed base robots. We do not recommend enabling this."
+            )
+        if self._use_cc_compensation:
+            log.warning(
+                "JointController is using Coriolis / centrifugal compensation. This is an experimental feature. We do "
+                "not recommend enabling this."
```
cremebrule, line 124: fix

omnigibson/controllers/osc_controller.py
```diff
@@ -69,6 +69,8 @@ def __init__(
         mode="pose_delta_ori",
         decouple_pos_ori=False,
         workspace_pose_limiter=None,
+        use_gravity_compensation=False,
+        use_cc_compensation=False,
```
cremebrule, line 73: same here

omnigibson/objects/controllable_object.py
```diff
@@ -189,6 +188,18 @@ def load(self, scene):
...
+    def _post_load(self):
+        # Call super first
+        super()._post_load()
+
+        # For controllable objects, we disable gravity of all links that are not the base link. This
+        # is because we cannot accurately apply gravity compensation in the absence of a working
+        # generalized gravity force computation. This may have some side effects on the measured
+        # torque on each of these links, but it provides a greatly improved joint control behavior.
+        for link in self._links.values():
+            if link is not self.root_link:
```
cremebrule, line 200: this is quite elegant, but may be problematic depending on the robot. For example, isn't Tiago's root link technically x_base_footprint? And a robot's root link may not actually be the highest mass link

A very hacky, but perhaps more accurate heuristic could be to only select the largest mass link to be gravity enabled.

omnigibson/robots/robot_base.py
```diff
@@ -129,7 +129,6 @@ def __init__(
 
         # Initialize internal attributes that will be loaded later
         self._sensors = None  # e.g.: scan sensor, vision sensor
-        self._dummy = None  # Dummy version of the robot w/ fixed base for computing generalized gravity forces
```
cremebrule, line 132: BLESS

omnigibson/utils/ui_utils.py
```diff
@@ -590,7 +590,7 @@ def __init__(self, robot):
                 "command_dim": controller.command_dim,
             }
             idx += controller.command_dim
-            for i in controller.dof_idx:
+            for i in controller.dof_idx.tolist():
```
cremebrule, line 593: is this because it could be shape (1, N) instead of (N,)?

cremebrule (commented):
overall LGTM, a couple of comments

omnigibson/utils/ui_utils.py
cgokmen: It's just an error that happened in the torch conversion. `i` needs to be an int and not a tensor.

omnigibson/robots/robot_base.py
cgokmen: :))))))))

omnigibson/objects/controllable_object.py
cgokmen: I have implemented our discussesd solution.

omnigibson/objects/controllable_object.py
```diff
@@ -189,6 +189,34 @@ def load(self, scene):
...
+        # Note that we do NOT disable gravity for links that are fixed to the base link, as these links
+        # are typically where most of the downward force on the robot is applied. Disabling gravity
+        # for these links would result in the robot floating in the air easily. Also note that here
+        # we use the base link footprint which takes into account the presence of virtual joints.
+
+        # Find all the links that are accessible from the base link footprint via a chain of fixed
+        # joints. We will disable gravity for all links that are not in this set.
+        articulation_tree = self.articulation_tree
+        base_footprint = self.base_footprint_link_name
+        is_edge_fixed = lambda f, t: articulation_tree[f][t]["joint_type"] == "FixedJoint"
```
cremebrule, line 209: use JointType class from utils.constants

cremebrule (approved):

PR #881, opened 2024-09-13
update dependency versions, remove torchvision dep

cremebrule:


omnigibson/scene_graphs/graph_builder.py
```diff
@@ -1,10 +1,10 @@
 import itertools
 
 import networkx as nx
+import numpy as np
```
hang-yin, line 4: Somehow the numpy pre-commit checker isn't working. Will need to look into it.

cgokmen (approved):

omnigibson/scene_graphs/graph_builder.py
cremebrule: manually fixed this for now

PR #886, opened 2024-09-17
Minor update to isaac sim setup file

yyf20001230:
Bug fixes on the setup.sh file not correctly finding the isaac sim installation path and the isaac sim .sh file


ChengshuLi (approved):
LGTM

PR #905, opened 2024-09-24
Eric's release fixes

ChengshuLi:


omnigibson/object_states/kinematics_mixin.py
```diff
@@ -29,7 +29,7 @@ def cache_info(self, get_value_args):
         info[self.obj] = {"q": self.obj.states[Joint].get_value(), "p": self.obj.states[Pose].get_value()}
         for arg in get_value_args:
             if isinstance(arg, StatefulObject):
-                info[arg] = {"q": self.obj.states[Joint].get_value(), "p": arg.states[Pose].get_value()}
+                info[arg] = {"q": arg.states[Joint].get_value(), "p": arg.states[Pose].get_value()}
```
hang-yin, line 32: @yyf20001230 could this be related to your joint state error?

omnigibson/utils/usd_utils.py
```diff
@@ -1743,3 +1752,39 @@ def deep_copy_prim(source_root_prim, dest_stage, dest_root_path):
...
+        return False
+
+    # If the prim is not ancestral, we can delete it.
+    if not lazy.omni.isaac.core.utils.prims.is_prim_ancestral(prim_path):
+        lazy.omni.usd.commands.DeletePrimsCommand([prim_path], destructive=True).do()
+
+    # Otherwise, we can only deactivate it, which essentially serves the same purpose.
+    # All objects that are originally in the scene are ancestral because we add the pre-build scene to the stage.
+    else:
+        lazy.omni.usd.commands.DeletePrimsCommand([prim_path], destructive=False).do()
```
hang-yin, line 1788: Does this mean we won't see this prim on the current stage and we will be able to add a new prim at this path?

omnigibson/object_states/kinematics_mixin.py
ChengshuLi: Yes.

omnigibson/object_states/attached_to.py
```diff
@@ -316,7 +316,7 @@ def _attach(self, other, child_link, parent_link, joint_type=None, can_joint_bre
             _, parent_local_quat = T.relative_pose_transform([0, 0, 0], child_quat, [0, 0, 0], parent_quat)
 
         # Disable collision between the parent and child objects
-        self._disable_collision_between_child_and_parent(child=self.obj, parent=other)
+        # self._disable_collision_between_child_and_parent(child=self.obj, parent=other)
```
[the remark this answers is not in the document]
ChengshuLi: I am not entirely sure. I think this might be needed for other attachment assets (not needed for the shelves in our demo). In any case, I will leave the code here and do the refactoring after release.

omnigibson/utils/usd_utils.py
ChengshuLi: We will still see the prim on the current stage, but it will be grayed out. This is the best we can do when removing an object that is originally in the scene json because we pre-baked all the objects into a monolithic USD before loading it.

And unfortunately we cannot add a new prim at this path afterwards, which can introduce some unexpected behavior (e.g. if we save the scene with og.sim.save, and then remove a scene object, og.sim.restore won't be able to incrementally import back the original object). I have created a ticket to track this and I will handle this post-release.

hang-yin (approved):

PR #908, opened 2024-09-26
Trigger volumes

hang-yin:
Implements https://github.com/StanfordVL/OmniGibson/issues/882

- [ ] Consider trigger report events: `og.sim.psi.subscribe_physics_trigger_report_events`
- [x] Verify with flatcache on


cgokmen:
@hang-yin figure out if this works w/ flatcache. if not, how do we use the callback?

PR #913, opened 2024-09-27
Check for missing test xml

hang-yin:


cgokmen (approved):
This doesnt fully fix the issue, but is a good start

PR #919, opened 2024-10-01
PyPI release infra, version bump

cgokmen:
Also add some instructions for future release-makers and bump the docker image base version.


cremebrule (approved):
lgtm!

PR #927, opened 2024-10-02
Fix missing rename behavior in old glibc versions

cgokmen:
Also fix small issue in dataset install parameter name.

Also unfold the dataset install and isaac path parameters from the + button into just always-visible text.


wensi-ai (approved):
ltgm

cremebrule (approved):
lgtm!

PR #980, opened 2024-10-15
Address Contact Issues

hang-yin:
**Motivation**
1. Small/light objects may fall asleep during motion after teleporting
2. RigidContactAPI does not give valid impulse/contact data even when two objects are clearly touching and awake

**Fix**
1. Lower sleep threshold from 0.001 to 0.00005
2. Lower contact threshold to 0.0; Use contact sensor instead for transition rules TouchingAnyCondition

**Implications**
1. We have three ways of checking contact in OG:
 - RigidContactAPI: 
   - This works when contacting objects are both awake, but not when they are asleep
   - We will use this when objects are guaranteed to be awake when we do the contact check
     - Assisted Grasping: robot fingers need to be moving and touching the object
     - ToggleOn: robot fingers need to be moving and touching the toggle button
     - SlicerActive: slicer need to be moving and touching the sliceable
 - Simulator _on_contact callback:
   - This triggers whenever there's contact; currently only used for attachment
 - Contact sensor: 
   - This works regardless if an object is awake or asleep, presumably because of some internal contact caching mechanism
   - We will use this when motion/objects being awake is not guaranteed:
     - Transition Rules TouchingAnyCondition
     - ContactBodies
     - Touching
2. In general, contact sensor query is slower than batched RigidContactAPI query. Going forward, we will try to use RigidContactAPI whenever possible, and use contact sensor as a fallback.


omnigibson/prims/entity_prim.py
```diff
@@ -21,7 +21,8 @@
 m = create_module_macros(module_path=__file__)
 
 # Default sleep threshold for all objects -- see https://docs.omniverse.nvidia.com/extensions/latest/ext_physics/simulation-control/physics-settings.html?highlight=sleep#sleeping
-m.DEFAULT_SLEEP_THRESHOLD = 0.001
+# Mass-normalized kinetic energy threshold below which an actor may go to sleep
+m.DEFAULT_SLEEP_THRESHOLD = 0.00005
```
cremebrule, line 25: with this threshold do objects still go to sleep? I'm curious what the performance implications are

omnigibson/prims/rigid_prim.py
```diff
@@ -159,7 +155,7 @@ def _initialize(self):
 
         # Get contact info first
         if self.contact_reporting_enabled:
-            self._cs.get_rigid_body_raw_data(self.prim_path)
+            og.sim.contact_sensor.get_rigid_body_raw_data(self.prim_path)
```
cremebrule, line 158: is this faster than the cs method?

cremebrule: o nvm this is equivalent!

omnigibson/simulator.py
```diff
@@ -1733,6 +1736,14 @@ def initial_rendering_dt(self):
             """
             return self._initial_rendering_dt
 
+        @property
+        def contact_sensor(self):
```
cremebrule, line 1740: should this live under usd_utils as a separate API singleton? or maybe part of RigidContactAPI?

cremebrule (commented):
just a few comments! thanks for looking into this!!

btw, forcing objects awake doesn't make RigidContactAPI work? And if so, why is  that?????

omnigibson/prims/entity_prim.py
hang-yin: They do! I found this from an isaac sim source file, seems to be the default. I tested both of these in `house_single_floor` with a Fetch and with object states & transition rules on, got similar fps out of them. I think it was something like 27 vs 27.3.

This was originally a problem because when I teleport a tablespoon on top of a bowl and drop it, it would freeze mid-air, falling asleep even when it seems like it should keep moving. This smaller numbers helps it fall normally and still gets it to sleep once it settles.

omnigibson/simulator.py
hang-yin: Good point! I did think about integrating it into RigidContactAPI, but went with this to get minimal changes in this PR. 

@cgokmen @ChengshuLi What are your thoughts on this?

hang-yin:
> comments! thanks for looking into this!!
> 
> btw, forcing objects awake doesn't make RigidContactAPI

I guess the point is that if objects need explicit waking before generic contact checking, this check would generally be slow. We could potentially benchmark contact sensor call against waking+RigidContactAPI

omnigibson/simulator.py
cgokmen: I think this is fine to keep here, it's just like the physx sim interface etc. that just provide a centralized interface and dont provide any additional functionality

omnigibson/transition_rules.py
```diff
@@ -347,78 +346,22 @@ def __init__(self, filter_1_name, filter_2_name):
         self._filter_1_name = filter_1_name
         self._filter_2_name = filter_2_name
 
-        # Will be filled in during self.initialize
```
cgokmen, line 350: Is the final decision to undo this?

cgokmen (changes requested):
LGTM, but are we undoing the transition rule thing?

omnigibson/transition_rules.py
hang-yin: @ChengshuLi eric can you confirm? Can revert this if you're ok with it

hang-yin: @cgokmen reverted back to RigidContactAPI, just confirmed with eric

cgokmen (approved):

PR #990, opened 2024-10-18
Feat/controller dependencies

cremebrule:


omnigibson/prims/geom_prim.py
```diff
@@ -117,7 +117,7 @@ def opacity(self, opacity):
         if self.has_material():
             self.material.opacity_constant = opacity
         else:
-            self.set_attribute("primvars:displayOpacity", th.tensor([opacity]))
+            self.set_attribute("primvars:displayOpacity", np.array([opacity]))
```
cgokmen, line 120: There is no `np` here and it's not allowed either. Does set_attribute not unwrap the tensor?

omnigibson/robots/articulated_trunk_robot.py
```diff
@@ -143,21 +47,110 @@ def trunk_control_idx(self):
...
+            "mode": "pose_delta_ori",
+            "workspace_pose_limiter": None,
+        }
+
+    @property
+    def _default_trunk_joint_controller_config(self):
+        """
+        Returns:
+            dict: Default base joint controller config to control this robot's base. Uses velocity
+                control by default.
```
cgokmen, line 114: Seems to use position control by default

omnigibson/robots/holonomic_base_robot.py
```diff
@@ -282,11 +282,18 @@ def _postprocess_control(self, control, control_type):
...
+
+        local_pose = th.zeros((2, 4, 4))
+        local_pose[:] = th.eye(4)
+        local_pose[:, :3, 3] = u_vec[self.base_idx].view(2, 3)
 
         # Rotate the linear and angular velocity to the desired frame
-        lin_vel_global, _ = T.pose_transform(th.zeros(3), cur_orn, u_vec[self.base_idx[:3]], th.tensor([0, 0, 0, 1]))
-        ang_vel_global, _ = T.pose_transform(th.zeros(3), cur_orn, u_vec[self.base_idx[3:]], th.tensor([0, 0, 0, 1]))
+        global_pose = cur_pose @ local_pose
+        lin_vel_global, ang_vel_global = global_pose[0, :3, 3], global_pose[1, :3, 3]
```
cgokmen, line 296: Can you explain what's going on here? What was the original thing doing, what's the new one doing? Very hard to follow!

omnigibson/robots/vx300s.py
```diff
@@ -115,7 +115,7 @@ def _create_discrete_action_space(self):
         raise ValueError("VX300S does not support discrete actions!")
 
     @property
-    def controller_order(self):
+    def _raw_controller_order(self):
```
cgokmen, line 118: General question: it seems to me that this could easily be implemented in the manipulation/articulatedtrunk/locomotion etc robot classes since it's always exactly the same thing for every robot? (e.g. each class calls super and appends its extra element).

cgokmen: Same for default_controllers.

omnigibson/utils/python_utils.py
```diff
@@ -590,20 +591,20 @@ def __getitem__(self, item):
...
-            *args (tuple): Positional arguments to pass into the function call
-            **kwargs (tuple): Keyword arguments to pass into the function call
 
         Returns:
             any: Output of the function referenced by @name
         """
-        return self._fcns[name](*args, **kwargs)
+        if name not in self._cache:
+            self._cache[name] = self._fcns[name]()
+        return self._cache[name]
```
cgokmen, line 607: I removed this on purpose: the caching currently already implicitly happens in ControllableObjectViewAPI thanks to its batching mechanism so this actually doesn't do anything except make cache misses just a bit slower.

omnigibson/utils/transform_utils.py
```diff
@@ -7,10 +7,11 @@
 import math
 from typing import List, Optional, Tuple
 
+import torch
```
cgokmen, line 10: Lets remove the `as th ` import then

omnigibson/configs/fetch_behavior.yaml
```diff
@@ -61,8 +59,11 @@ robots:
     controller_config:
       base:
         name: DifferentialDriveController
+      trunk:
+        name: JointController
```
cgokmen, line 63: I think we shouldn't need this right? Can we remove it? From all of these files?

omnigibson/configs/fetch_behavior.yaml
```diff
@@ -61,8 +59,11 @@ robots:
...
       arm_0:
         name: InverseKinematicsController
+        dependencies: [trunk]
```
cgokmen, line 66: Can we change this to `subsume_controllers:` to make it very very explicit what's going on? Also in the code that parses these.

cgokmen (changes requested):

omnigibson/prims/geom_prim.py
cremebrule: nope, apparently np is required.

omnigibson/robots/articulated_trunk_robot.py
cremebrule: fixed

omnigibson/robots/holonomic_base_robot.py
cremebrule: Oh it's doing the exact same transform. Basically transforming the desired local (robot frame) pose the equivalent global orientation. It's really inefficient here because we waste compute doing redundant orientation convention conversations back and forth, and multiple matrix multiplies. I optimized it by removing the redundancies and batching the matrix multiplications into a single call.

omnigibson/robots/vx300s.py
cremebrule: Yeah this is a great point. The only downside is that it then comes down to the arbitrary superclass definition order defined by each Robot class (e.g.: R1), so it might not be consistent across different robots

omnigibson/utils/python_utils.py
cremebrule: Oh interesting...that makes sense because I was almost positive I wouldn't have missed such a glaring bug lol

When I was timing I found that something is evidently not caching properly in ControllableObjectViewAPI, as the time cost is linear when called multiple times during the same controller loop, which doesn't seem correct.

omnigibson/utils/transform_utils.py
cremebrule: done

omnigibson/configs/fetch_behavior.yaml
cremebrule: done

omnigibson/configs/fetch_behavior.yaml
cremebrule: this is more just to make it clear that the Fetch robot has a trunk that may be controlled

PR #1017, opened 2024-10-26
Draped demo

hang-yin:
image


cgokmen:
@cgokmen will try this out and try to complete it

PR #1027, opened 2024-11-06
Integrate cuRobo into action primitives & primitives refactoring

hang-yin:
- [x] Replace OMPL planning context with cuRobo motion planning & collision checking
- [x] Implement cuRobo single IK solver for robot reachability test
- [x] Implement linear motion in cartesian space with cuRobo constrained planning
- [x] Generalize primitives to dual-arm settings
- [x] Investigate why cuRobo planning keeps failing if we do not do reset+gripper matching
- [ ] Accommodate differential drive base planning (implement a light-weight RRT)
- [x] Potentially support flatcache with curobo (get_obstacle implementation)

https://github.com/user-attachments/assets/bf386bb2-a986-4927-b223-ebc4b8691058


omnigibson/configs/tiago_primitives.yaml
```diff
@@ -47,33 +75,53 @@ robots:
...
         name: JointController
+        motor_type: position
+        command_input_limits: null
+        use_delta_commands: false
+        use_impedances: true
+        pos_kp: 500
       arm_left:
         name: JointController
-        subsume_controllers: [trunk]
+        # subsume_controllers: [trunk]
```
ChengshuLi, line 94: why comment out?

omnigibson/configs/tiago_primitives.yaml
```diff
@@ -47,33 +75,53 @@ robots:
     controller_config:
       base:
         name: JointController
+        motor_type: position
+        command_input_limits: null
+        use_delta_commands: false
+        use_impedances: true
+        pos_kp: 200
+        pos_ki: 0.5
+        max_integral_error: 10.0
```
ChengshuLi, line 84: what is this?

omnigibson/controllers/joint_controller.py
```diff
@@ -204,7 +215,11 @@ def compute_control(self, goal_dict, control_dict):
                 # Run impedance controller -- effort = pos_err * kp + vel_err * kd
                 position_error = target - base_value
                 vel_pos_error = -control_dict[f"joint_velocity"][self.dof_idx]
-                u = position_error * self.pos_kp + vel_pos_error * self.pos_kd
+
+                # Update integral error
+                self._integral_error += position_error * (1.0 / self.control_freq)  # Integrate error over time
+                self._integral_error = self._integral_error.clip(-self.max_integral_error, self.max_integral_error)
+                u = position_error * self.pos_kp + vel_pos_error * self.pos_kd + self._integral_error * self.pos_ki
```
ChengshuLi, line 222: @cremebrule @cgokmen should also review this. Any implication? Did all the robot tests still pass? Does robot_control_example still work?

omnigibson/examples/action_primitives/rs_int_example.py
```diff
@@ -48,11 +50,31 @@ def main():
...
+    env.scene.update_initial_state()
+    env.scene.reset()
+
+    og.sim.viewer_camera.set_position_orientation(
+        th.tensor([-0.6084, -3.3571, 1.2033]), th.tensor([0.6349, -0.1778, -0.2027, 0.7240])
+    )
+
+    # Let the object settle
+    for _ in range(30):
+        og.sim.step()
```
ChengshuLi, line 71: these can be removed after my PR.

omnigibson/examples/action_primitives/solve_simple_task.py
```diff
@@ -57,10 +58,26 @@ def main():
...
+
+    for _ in range(5):
+        og.sim.step()
+
+    env.scene.update_initial_state()
+    env.scene.reset()
+
+    # Let the object settle
+    for _ in range(30):
+        og.sim.step()
```
ChengshuLi, line 75: same

omnigibson/robots/r1.py
```diff
@@ -274,7 +302,33 @@ def disabled_collision_pairs(self):
...
+            ["left_arm_link6", "left_gripper_link1"],
+            ["left_arm_link6", "left_gripper_link2"],
+            ["torso_link4", "right_arm_link1"],
+            ["right_arm_link1", "right_arm_link2"],
+            ["right_arm_link2", "right_arm_link3"],
+            ["right_arm_link3", "right_arm_link4"],
+            ["right_arm_link4", "right_arm_link5"],
+            ["right_arm_link5", "right_arm_link6"],
+            ["right_arm_link6", "right_gripper_link1"],
+            ["right_arm_link6", "right_gripper_link2"],
```
ChengshuLi, line 333: Do we actually need these? I thought the collision between parent and child links are already disabled by default.

omnigibson/utils/grasping_planning_utils.py
```diff
@@ -28,15 +28,18 @@ def get_grasp_poses_for_object_sticky(target_obj):
...
+
+    bbox_center_world = (aabb_min_world + aabb_max_world) / 2
+    bbox_extent_world = aabb_max_world - aabb_min_world
+
+    grasp_center_pos = bbox_center_world + th.tensor([0, 0, bbox_extent_world[2] / 2])
+    towards_object_in_world_frame = bbox_center_world - grasp_center_pos
     towards_object_in_world_frame /= th.norm(towards_object_in_world_frame)
 
-    grasp_quat = T.euler2quat(th.tensor([0, math.pi / 2, 0], dtype=th.float32))
+    # TODO: why was this pi/2 in the y-axis?
```
ChengshuLi, line 41: is this comment still valid? Also, what does the grasp_quat mean? Is it the desired eef orientation? If so, will different robots have different values (which is ugly..)? Lastly, this only works for "sticky" grasping right? Maybe we should assert that or at least add some comments to indicate this.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -273,28 +129,35 @@ class StarterSemanticActionPrimitives(BaseActionPrimitiveSet):
...
             task_relevant_objects_only (bool): Whether to only consider objects relevant to the task
               when computing the action space. Defaults to False.
+            planning_batch_size (int): The batch size for curobo motion planning. Defaults to 3.
+            collision_check_batch_size (int): The batch size for curobo collision checking. Defaults to 5.
+            debug_visual_marker (PrimitiveObject): The object to use for debug visual markers. Defaults to None.
         """
         log.warning(
             "The StarterSemanticActionPrimitive is a work-in-progress and is only provided as an example. "
-            "It currently only works with Fetch and Tiago with their JointControllers set to delta mode."
+            "It currently only works with Tiago and R1 with their JointControllers set to absolute position mode with impedance."
```
ChengshuLi, line 157: maybe you should just assert these instead of printing out a warning.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -345,107 +201,28 @@ def __init__(
...
-    @property
-    def _base_controller_is_joint(self):
-        return isinstance(self.robot.controllers["base"], JointController)
+    @arm.setter
+    def arm(self, value):
+        if not isinstance(self.robot, ManipulationRobot):
+            raise ValueError("Cannot use arm for non-manipulation robot")
+        if value not in self.robot.arm_names:
+            raise ValueError(f"Invalid arm name: {value}. Must be one of {self.robot.arm_names}")
+        self._arm = value
```
ChengshuLi, line 219: What does setting the arm mean?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -733,40 +515,56 @@ def _grasp(self, obj):
         grasp_poses = get_grasp_poses_for_object_sticky(obj)
         grasp_pose, object_direction = random.choice(grasp_poses)
 
+        grasp_offset_in_z = self.robot.finger_lengths[self.arm] + m.GRASP_APPROACH_DISTANCE
+
+        # Adjust grasp pose with reset orientation and finger length offset
+        reset_orientation = self._get_reset_eef_pose("world")[self.arm][1]
+        grasp_pos = grasp_pose[0] - object_direction * grasp_offset_in_z
```
ChengshuLi, line 522: hmm does `grasp_pose` initially contain this offset and you are undoing it here? Not sure if this is the cleanest code, happy to discuss in case I misunderstand!

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -733,40 +515,56 @@ def _grasp(self, obj):
...
-
-        # Since the grasp pose is slightly off the object, we want to move towards the object, around 5cm.
-        # It's okay if we can't go all the way because we run into the object.
-        indented_print("Performing grasp approach")
-        yield from self._move_hand_linearly_cartesian(approach_pose, stop_on_contact=True)
-
-        # Step once to update
-        empty_action = self._empty_action()
-        yield self._postprocess_action(empty_action)
+        yield from self._move_hand(grasp_pose, motion_constraint=[0, 0, 0, 0, 1, 0])
```
ChengshuLi, line 535: add comment to explain what does `motion_constraint=[0, 0, 0, 0, 1, 0]` do here.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -733,40 +515,56 @@ def _grasp(self, obj):
...
+            yield from self._execute_grasp()
+            # Since the grasp pose is slightly off the object, we want to move towards the object, around 5cm.
+            # It's okay if we can't go all the way because we run into the object.
+            indented_print("Performing grasp approach")
+            # Use direct IK to move the hand to the approach pose.
+            yield from self._move_hand(approach_pose, avoid_collision=False)
+        elif self.robot.grasping_mode == "assisted":
+            indented_print("Performing grasp approach")
+            yield from self._move_hand(approach_pose, motion_constraint=[0, 0, 0, 0, 0, 1])
+            yield from self._execute_grasp()
```
ChengshuLi, line 549: hmm let's discuss these grasping logic in person.. I am not sure I fully understand all of these.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -733,40 +515,56 @@ def _grasp(self, obj):
...
+        obj_in_hand = self._get_obj_in_hand()
+        if obj_in_hand is None:
             raise ActionPrimitiveError(
                 ActionPrimitiveError.Reason.POST_CONDITION_ERROR,
                 "Grasp completed, but no object detected in hand after executing grasp",
                 {"target object": obj.name},
             )
 
+        # TODO: ag force seems to not be enough to keep the object in hand. need to investigate
+        obj_in_hand.root_link.density = 1.0
```
ChengshuLi, line 564: hmm should we investigate this right now?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -854,11 +658,39 @@ def _place_with_predicate(self, obj, predicate):
...
 
         # Sample location to place object
-        obj_pose = self._sample_pose_with_object_and_predicate(predicate, obj_in_hand, obj)
-        hand_pose = self._get_hand_pose_for_object_pose(obj_pose)
+        pose_candidates = []
+        directly_move_hand_pose = None
+        while len(pose_candidates) < m.MAX_ATTEMPTS_FOR_SAMPLING_PLACE_POSE:
+            obj_pose = self._sample_pose_with_object_and_predicate(predicate, obj_in_hand, obj)
+            # TODO: why do we sample this orientation?
+            obj_pose = (obj_pose[0], obj_in_hand.get_position_orientation()[1])
```
ChengshuLi, line 666: is this sampling the stable orientation?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -854,11 +658,39 @@ def _place_with_predicate(self, obj, predicate):
...
+                    yield from self._navigate_to_pose(valid_navigation_pose)
+                    yield from self._move_hand(candidate)
+                    break
+
+        if valid_navigation_pose is None:
+            raise ActionPrimitiveError(
+                ActionPrimitiveError.Reason.PLANNING_ERROR,
+                "Could not find a valid pose to place the object",
+                {"target object": obj.name},
+            )
```
ChengshuLi, line 692: is `valid_navigation_pose is None` the only failure mode? is it possible for `pose_candidates` to be empty?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -919,36 +751,27 @@ def _target_in_reach_of_robot_relative(self, relative_target_pose):
             target_pose (Iterable of array): Position and orientation arrays in an iterable for pose for the eef
 
         Returns:
-            bool: Whether eef can the reach target pose
+            dict: Whether each eef can reach the target pose
         """
-        return self._ik_solver_cartesian_to_joint_space(relative_target_pose) is not None
+        # TODO: output may need to be a dictionary of arms and whether they can reach the target pose
```
ChengshuLi, line 756: is this TODO finished?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -919,36 +751,27 @@ def _target_in_reach_of_robot_relative(self, relative_target_pose):
...
-            return self.robot.robot_arm_descriptor_yamls["left_fixed"]
-
-        # Otherwise just return the default arm control idx
-        return self.robot.robot_arm_descriptor_yamls[self.arm]
+        if m.TORSO_FIXED or arm == "right":
+            return self.robot.arm_control_idx[arm]
+        else:
+            # Only append trunk control idx if it is not fixed and the arm is left
+            if hasattr(self.robot, "trunk_control_idx"):
+                return th.cat([self.robot.trunk_control_idx, self.robot.arm_control_idx[arm]])
```
ChengshuLi, line 772: Why is this needed? Also, do we actually need a separate flag `m.TORSO_FIXED`? Should we instead infer from the robot controller config?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -960,22 +783,36 @@ def _ik_solver_cartesian_to_joint_space(self, relative_target_pose):
...
         """
+        if arm is None:
+            arm = self.arm
+        # TODO: make sure descriptor file match torso & arm setting
         ik_solver = IKSolver(
-            robot_description_path=self._manipulation_descriptor_path,
+            robot_description_path=self.robot.robot_arm_descriptor_yamls[arm],
             robot_urdf_path=self.robot.urdf_path,
-            reset_joint_pos=self.robot.reset_joint_pos[self._manipulation_control_idx],
+            reset_joint_pos=self.robot.reset_joint_pos[self._manipulation_control_idx(arm)],
```
ChengshuLi, line 792: oh `self._manipulation_control_idx` is used here - so if we get rid of lula completely, we don't need this any more?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -985,86 +822,168 @@ def _move_hand(self, target_pose, stop_if_stuck=False):
...
-            th.tensor or None: Action array for one step for the robot to move arm or None if its at the joint positions
-        """
-        with PlanningContext(self.env, self.robot, self.robot_copy, "original") as context:
-            plan = plan_arm_motion(
-                robot=self.robot,
-                end_conf=joint_pos,
-                context=context,
-                torso_fixed=m.TIAGO_TORSO_FIXED,
+        self._motion_generator.update_obstacles()
+        while not success and planning_attempts < m.MAX_PLANNING_ATTEMPTS:
```
ChengshuLi, line 892: hmm why don't we directly pass in `max_attempts = planning_attempts`?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1495,43 +1423,102 @@ def _empty_action(self, follow_arm_targets=True):
...
+        Returns:
+            dict of th.tensor: The reset eef pose for each robot arm
+        """
+        if isinstance(self.robot, Fetch):
+            pose = {
+                self.arm: (
+                    th.tensor([0.48688125, -0.12507881, 0.97888719]),
+                    th.tensor([0.61324748, 0.61305553, -0.35266518, 0.35173529]),
+                )
+            }
```
ChengshuLi, line 1480: remove fetch?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1495,43 +1423,102 @@ def _empty_action(self, follow_arm_targets=True):
...
+            pose = {
+                self.robot.arm_names[0]: (
+                    th.tensor([0.5021, 0.2458, 0.7648]),
+                    th.tensor([-0.5599, 0.5649, 0.4303, 0.4269]),
+                ),
+                self.robot.arm_names[1]: (
+                    th.tensor([0.4999, -0.2486, 0.7633]),
+                    th.tensor([-0.5592, -0.5646, 0.4311, -0.4274]),
+                ),
+            }
```
ChengshuLi, line 1513: is this the same as the reset joint pos?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1584,12 +1571,12 @@ def _get_reset_joint_pos(self):
                 4.50000000e-02,
             ]
         )
-        if self.robot_model == "Fetch":
+        if isinstance(self.robot, Fetch):
```
ChengshuLi, line 1574: remove fetch and add R1?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1601,31 +1588,20 @@ def _navigate_to_pose(self, pose_2d):
...
             th.tensor or None: Action array for one step for the robot to navigate or None if it is done navigating
         """
-        with PlanningContext(self.env, self.robot, self.robot_copy, "simplified") as context:
-            plan = plan_base_motion(
-                robot=self.robot,
-                end_conf=pose_2d,
-                context=context,
-            )
+        pose_3d = self._get_robot_pose_from_2d_pose(pose_2d)
+        # TODO: why was this 0.1?
```
ChengshuLi, line 1592: is this comment still valid?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -2042,12 +2049,13 @@ def _settle_robot(self):
         Returns:
             th.tensor or None: Action array for one step for the robot to do nothing
         """
+        # TODO: fix empty action
```
ChengshuLi, line 2052: is this comment still valid?

omnigibson/controllers/joint_controller.py
hang-yin: I'll leave default ki to be 0.0 for now - let's chat more about this in-person

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: I did it initially, but I found that our `test_robot_load_drive` depends on some primitives utilities such as `_move_hand_direct_ik` and `_navigate_to_pose_direct` that apply to all robots - not just Tiago and R1. I did talk with Cem about this - he wanted these utilities to stay within the primitives for now. Let me know what you think!

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: Some of the primitives take in an arm argument to specify which arm to move, e.g. _move_hand and _reset_hand. This arm setter sets the default arm if this arm argument isn't passed in. Some reachability tests also assume to use this self.arm. We can also make all of them take the arm argument! In that case, this would better be phrased as default_arm.

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: The initial grasp orientation is robot-agnostic, so the orientation it produces doesn't really mean anything. We could rename `get_grasp_poses_for_object_sticky` to something like `get_grasp_position_for_object_sticky` to make things clear. I added this reset orientation here so that if `[0,0,0,1]` doesn't align with the upright eef orientation, we can still transform that into the default (hopefully upright) eef orientation. I agree that this doesn't work all the time though - we should allow a robot to have non-upright reset eef pose. What do you think? Should we record some form of "upright" eef orientation for each manipulation robot for sticking grasping?

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: Essentially, 
for sticky grasping - we 1) close gripper 2) poke the object
for assisted grasping - we 1) get into position first with the gripper open 2) close the gripper

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: This is also related to how we handle gravity compensation. I'm not super sure what the best way to do this is.

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: I guess the arbitrary thing here is that: 

for bimanual manipulation, we should allow each arm to bond with the torso for maximum reachability. If we subsume torso into only one arm with the robot controller config, it seems to me that we will only attach torso to one, not two arms.  Potentially a TORSO_FIXED flag can help here, idk

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: These are the eef poses - so potentially we can run forward kinematics to get these from reset joint positions.

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: I guess with cuRobo arm embodiment selection, we automatically include all the torso indices. Here's our current implementation:
```
    def _manipulation_control_idx(self, arm=None):
        """The appropriate manipulation control idx for the current settings."""
        if arm is None:
            arm = self.arm

        return th.cat([self.robot.trunk_control_idx, self.robot.arm_control_idx[arm]])
```
Do you think we need a fixed-torso mode?

omnigibson/controllers/joint_controller.py
hang-yin: This is removed!

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: This is fixed!

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: This is fixed!

omnigibson/action_primitives/starter_semantic_action_primitives.py
hang-yin: For some reasons, R1 would drift given its current joint positions as action. I'll look into this more tomorrow!

omnigibson/robots/holonomic_base_robot.py
```diff
@@ -301,32 +344,59 @@ def get_angular_velocity(self) -> th.Tensor:
...
+        for name, controller in self.controllers.items():
+            assert (
+                isinstance(controller, JointController) and not controller.use_delta_commands
+            ), f"Controller [{name}] should be a JointController/HolonomicBaseJointController with use_delta_commands=False!"
+            command = q[controller.dof_idx]
+            if isinstance(controller, HolonomicBaseJointController):
+                # For a holonomic base joint controller, the command should be in the robot local frame
+                cur_rz_joint_pos = self.get_joint_positions()[self.base_idx][5]
+                delta_q = wrap_angle(command[2] - cur_rz_joint_pos)
+                local_pose = self._2d_world_pose_to_robot_pose(command)
```
ChengshuLi, line 374: can we change this to the following?

```
body_pose = self.get_position_orientation()
canonical_pos = th.tensor([command[0], command[1], body_pose[0][2]], dtype=th.float32)
local_pos = T.relative_pose_transform(canonical_pos, th.tensor([0.0, 0.0, 0.0, 1.0]), *body_pose)[0]
command = th.tensor([local_pos[0], local_pos[1], delta_q])
```

omnigibson/robots/holonomic_base_robot.py
```diff
@@ -301,32 +344,59 @@ def get_angular_velocity(self) -> th.Tensor:
...
+        action = th.cat(action, dim=0)
+        assert (
+            action.shape[0] == self.action_dim
+        ), f"Action should have dimension {self.action_dim}, got {action.shape[0]}"
+        return action
 
-        # Rotate the linear and angular velocity to the desired frame
-        global_pose = cur_pose @ local_pose
-        lin_vel_global, ang_vel_global = global_pose[0, :3, 3], global_pose[1, :3, 3]
+    def _2d_world_pose_to_robot_pose(self, pose_2d):
```
ChengshuLi, line 384: This function can be removed.

omnigibson/robots/r1.py
```diff
@@ -243,4 +243,5 @@ def disabled_collision_pairs(self):
             ["base_link", "wheel_link1"],
             ["base_link", "wheel_link2"],
             ["base_link", "wheel_link3"],
+            ["torso_link2", "torso_link4"],
```
ChengshuLi, line 246: is this what we want?

omnigibson/tasks/grasp_task.py
```diff
@@ -104,16 +100,16 @@ def _reset_agent(self, env):
...
-                    if not set_arm_and_detect_collision(context, initial_joint_pos):
-                        robot.set_joint_positions(joint_pos, joint_control_idx)
-                        og.sim.step()
-                        break
+            for _ in range(MAX_JOINT_RANDOMIZATION_ATTEMPTS):
+                joint_pos, joint_control_idx = self._get_random_joint_position(robot)
+                initial_joint_pos[control_idx_in_joint_pos] = joint_pos
+                collision_detected = self._primitive_controller._motion_generator.check_collisions(
+                    [initial_joint_pos], self_collision_check=False
+                ).cpu()[0]
```
ChengshuLi, line 108: why `self_collision_check=False`?

ChengshuLi (commented):
Will finish my review for `omnigibson/action_primitives/starter_semantic_action_primitives.py` in the next round!

omnigibson/configs/tiago_primitives.yaml
```diff
@@ -32,50 +32,91 @@ scene:
...
         name: JointController
         motor_type: position
-        command_input_limits: [-1, 1]
+        command_input_limits: null
         command_output_limits: null
         use_delta_commands: false
+        use_impedances: false
       camera:
         name: JointController
+        use_delta_commands: false
```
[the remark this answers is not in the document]
hang-yin: Details in `_overwrite_head_action` and `_get_head_goal_q`

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1784,62 +1729,75 @@ def _rotate_in_place(self, end_pose, angle_threshold=m.DEFAULT_ANGLE_THRESHOLD):
...
+        yaw_lo, yaw_hi = -math.pi, math.pi
+        avg_arm_workspace_range = th.mean(self.robot.arm_workspace_range[self.arm])
 
-                if not self._test_pose(pose_2d, context, pose_on_obj=pose_on_obj, **kwargs):
-                    continue
+        target_pose = (
+            (self._sample_position_on_aabb_side(obj), self._get_reset_eef_pose()[self.arm][1])
+            if pose_on_obj is None
+            else pose_on_obj
+        )
```
ChengshuLi, line 1761: Shouldn't you reuse the logic from `_grasp`?
e.g. calling `get_grasp_poses_for_object_sticky` stc?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1784,62 +1729,75 @@ def _rotate_in_place(self, end_pose, angle_threshold=m.DEFAULT_ANGLE_THRESHOLD):
         empty_action = self._empty_action()
         yield self._postprocess_action(empty_action)
 
-    def _sample_pose_near_object(self, obj, pose_on_obj=None, **kwargs):
+    def _sample_pose_near_object(
+        self,
+        obj,
+        pose_on_obj=None,
```
ChengshuLi, line 1735: In general, I think the term `pose_on_obj` is quite confusing. Is it a pregrasp pose? Is it a grasp pose? By the end of the day, it's just an arbitrary eef pose to reach. In some way, you should either pass in obj or `pose_on_obj`, but not both, right?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1784,62 +1729,75 @@ def _rotate_in_place(self, end_pose, angle_threshold=m.DEFAULT_ANGLE_THRESHOLD):
...
-                },
-            )
+        attempt = 0
+        if not skip_obstacle_update:
+            # Update obstacle once before sampling
+            self._motion_generator.update_obstacles()
+        while attempt < sampling_attempts:
+            candidate_poses = []
+            for _ in range(self._collision_check_batch_size):
+                while True:
```
ChengshuLi, line 1774: hmm what if this while True loop is stuck forever?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1940,32 +1898,62 @@ def _sample_pose_with_object_and_predicate(
...
-            indented_print("Candidate position failed collision test.")
-            return False
-        return True
+        # First check collisions for all candidate poses
+        candidate_joint_positions = []
+        current_joint_pos = self.robot.get_joint_positions()
+        for pose in candidate_poses:
+            joint_pos = current_joint_pos.clone()
+            joint_pos[self.robot.base_idx[:2]] = pose[:2]
+            joint_pos[self.robot.base_idx[3:]] = th.tensor([0.0, 0.0, pose[2]])
```
ChengshuLi, line 1925: maybe the following instead?
```
joint_pos[self.robot.base_control_idx] = pose
```

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1940,32 +1898,62 @@ def _sample_pose_with_object_and_predicate(
...
+                pose = self._get_robot_pose_from_2d_pose(candidate_poses[i])
+                relative_pose = T.relative_pose_transform(*pose_on_obj, *pose)
+                if not self._target_in_reach_of_robot_relative(
+                    relative_pose, skip_obstacle_update=skip_obstacle_update
+                )[arm]:
+                    invalid_results[i] = True
+
+        return ~invalid_results
+
+    def _get_robot_pose_from_2d_pose(self, pose_2d):
```
ChengshuLi, line 1956: Figure out exactly what this function should return?

omnigibson/configs/tiago_primitives.yaml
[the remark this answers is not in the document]
ChengshuLi: I thought by default `command_input_limits` is "default", which set the command input limit to be between [-1, 1]

omnigibson/robots/r1.py
ChengshuLi: maybe comment out if no problem occurs?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -847,12 +657,38 @@ def _place_with_predicate(self, obj, predicate):
...
-        # Sample location to place object
-        obj_pose = self._sample_pose_with_object_and_predicate(predicate, obj_in_hand, obj)
-        hand_pose = self._get_hand_pose_for_object_pose(obj_pose)
+        self._motion_generator.update_obstacles()
+        target_in_reach = False
+        valid_navigation_pose = None
+        for _ in range(m.MAX_ATTEMPTS_FOR_SAMPLING_PLACE_POSE):
+            # Sample one pose at a time
+            obj_pose = self._sample_pose_with_object_and_predicate(predicate, obj_in_hand, obj)
+            obj_pose = (obj_pose[0], obj_in_hand.get_position_orientation()[1])
```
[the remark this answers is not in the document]
ChengshuLi: think about whether it's required for the object base link z-axis to be upright when placing the object

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -883,182 +719,300 @@ def _convert_cartesian_to_joint_space(self, target_pose):
...
+            timeout=60.0,
+            ik_fail_return=m.MAX_IK_FAILURES_BEFORE_RETURN,
+            enable_finetune_trajopt=True,
+            finetune_attempts=1,
+            return_full_result=False,
+            attached_obj=attached_obj,
+            motion_constraint=motion_constraint,
+            skip_obstacle_update=True,
+            emb_sel=embodiment_selection,
+        )
```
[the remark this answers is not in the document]
ChengshuLi: Make it explicit even though the default is already 1 / batch_size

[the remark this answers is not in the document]
ChengshuLi: not need to explicitly batch, curobo already does so if len(target_pos) < batch_size

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1594,31 +1541,18 @@ def _navigate_to_pose(self, pose_2d):
...
-
-        # Follow the plan to navigate.
-        # self._draw_plan(plan)
-        indented_print(f"Navigation plan has {len(plan)} steps")
-        for i, pose_2d in enumerate(plan):
-            indented_print(f"Executing navigation plan step {i + 1}/{len(plan)}")
-            low_precision = True if i < len(plan) - 1 else False
-            yield from self._navigate_to_pose_direct(pose_2d, low_precision=low_precision)
+        q_traj = self._plan_joint_motion(target_pos, target_quat, CuRoboEmbodimentSelection.BASE)
+        yield from self._execute_motion_plan(q_traj, stop_on_contact=False)
```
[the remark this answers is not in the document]
ChengshuLi: maybe `stop_on_contact=True` - unexpected collision during navigation

omnigibson/action_primitives/starter_semantic_action_primitives.py
ChengshuLi: quat = R.from_euler("XYZ", [cur_joint_rx, cur_joint_ry, cur_joint_rz]).as_quat()

omnigibson/controllers/holonomic_base_joint_controller.py
```diff
@@ -0,0 +1,143 @@
...
+
+            # Transform command to canonical frame
+            command_in_canonical_frame = canonical_to_base_pose_rotation @ command_in_base_frame
+
+            # Extract x,y linear velocity from transformed command
+            linear_velocity = command_in_canonical_frame[:2, 3]
+
+            angular_velocity = command[2:3]
+
+            command = cb.as_float32(cb.cat([linear_velocity, angular_velocity]))
```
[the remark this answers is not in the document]
ChengshuLi: removed.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -266,32 +131,41 @@ class StarterSemanticActionPrimitiveSet(IntEnum):
     TOGGLE_OFF = auto(), "Toggle an object off"
 
 
+# (TODO) execution failure: overwrite_head_action might move the head that causes self-collision with curobo-generated motion plans (at the time of curobo planning, it assumes the head doesn't move).
+
```
hang-yin, line 135: Another TODO for the future: we probably need some form of constraint on R1's torso so that it doesn't do any "yoga dance".

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -693,6 +488,36 @@ def _move_hand_upward(self, steps=5, speed=0.1):
...
+        # Identity quaternion for top-down grasping (x-forward, y-right, z-down)
+        approach_dir = T.quat2mat(grasp_quat) @ th.tensor([0.0, 0.0, -1.0])
+
+        pregrasp_offset = self.robot.finger_lengths[self.arm] / 2.0 + m.GRASP_APPROACH_DISTANCE
+
+        pregrasp_pos = grasp_pos - approach_dir * pregrasp_offset
+
+        # The sampled grasp pose is robot-agnostic
+        # We need to multiply by the quaternion of the robot's eef frame of top-down grasping (x-forward, y-right, z-down)
+        grasp_quat = T.quat_multiply(grasp_quat, th.tensor([1.0, 0.0, 0.0, 0.0]))
```
hang-yin, line 514: Is this dependent on the fact that all robot eef frames have the same orientation?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -722,45 +560,52 @@ def _grasp(self, obj):
...
             raise ActionPrimitiveError(
                 ActionPrimitiveError.Reason.POST_CONDITION_ERROR,
                 "Grasp completed, but no object detected in hand after executing grasp",
                 {"target object": obj.name},
             )
 
-        indented_print("Moving hand back")
-        yield from self._reset_hand()
+        # indented_print("Moving hand back")
+        # yield from self._reset_hand()
```
hang-yin, line 608: Remove this?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -869,196 +752,299 @@ def _place_with_predicate(self, obj, predicate):
                 {"dropped object": obj_in_hand.name, "target object": obj.name},
             )
 
-        yield from self._move_hand_upward()
+        # indented_print("Moving hand back")
+        # yield from self._reset_hand()
```
hang-yin, line 757: Remove this?

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -869,196 +752,299 @@ def _place_with_predicate(self, obj, predicate):
...
-                robot=self.robot,
-                end_conf=joint_pos,
-                context=context,
-                torso_fixed=m.TIAGO_TORSO_FIXED,
-            )
+        # Aggregate target_pos and target_quat to match batch_size
+        target_pos = {k: th.stack([v for _ in range(self._motion_generator.batch_size)]) for k, v in target_pos.items()}
+        target_quat = {
+            k: th.stack([v for _ in range(self._motion_generator.batch_size)]) for k, v in target_quat.items()
+        }
```
hang-yin, line 933: Do we still need to do this? I vaguely remember there's something in the curobo code that takes care of this.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -869,196 +752,299 @@ def _place_with_predicate(self, obj, predicate):
...
-        Yields action for the robot to move arm to reach the specified eef positions using the planner
+        traj_path = traj_paths[success_idx[0]]
+        q_traj = self._motion_generator.path_to_joint_trajectory(
+            traj_path, get_full_js=True, emb_sel=embodiment_selection
+        ).cpu()
 
-        Args:
-            eef_pose (th.tensor): End Effector pose for the arm
+        # Smooth out the trajectory
+        q_traj = th.stack(self._add_linearly_interpolated_waypoints(plan=q_traj, max_inter_dist=0.01))
```
hang-yin, line 974: We can look into this a little more in the future but I'm curious why the robot motion become really jerky without this interpolation.

omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1488,137 +1496,90 @@ def _empty_action(self, follow_arm_targets=True):
...
+        # breakpoint()
+        q_traj = self._plan_joint_motion(
+            target_pos,
+            target_quat,
+            embodiment_selection=CuRoboEmbodimentSelection.BASE,
+            skip_obstacle_update=skip_obstacle_update,
+        )
+        yield from self._execute_motion_plan(q_traj)
 
     def _draw_plan(self, plan):
```
hang-yin, line 1580: Another future todo: could be nice to have something like this for curobo - 1) base motion plan on the trav map 2) 3D eef trajectory

setup.py
```diff
@@ -68,7 +68,7 @@
             "telemoma~=0.1.2",
         ],
         "primitives": [
-            "nvidia-curobo @ git+https://github.com/StanfordVL/curobo@dc52be668a8d0b426b8639de3c8d6443e58cc348",
+            "nvidia-curobo @ git+https://github.com/StanfordVL/curobo@cbaf7d32436160956dad190a9465360fad6aba73",
```
hang-yin, line 71: Is there a specific reason why we don't use the latest commit on main?

tests/test_symbolic_primitives.py
```diff
@@ -9,11 +9,11 @@
...
 )
-from omnigibson.macros import gm
 
-gm.USE_GPU_DYNAMICS = True
-gm.ENABLE_TRANSITION_RULES = True
-current_robot_type = "Fetch"
+# TODO: Using GPU dynamics causes cuda memory issues, need to investigate
+# gm.USE_GPU_DYNAMICS = True
+# gm.ENABLE_TRANSITION_RULES = True
+current_robot_type = "R1"
```
hang-yin, line 16: We should probably fix this in a follow-up. I have a working version for this test file but it isn't as efficient.

docker/prod.Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN wget --no-verbose -O /cuda-keyring.deb https://developer.download.nvidia.com
   DEBIAN_FRONTEND=noninteractive apt-get install -y cuda-toolkit-11-8 && \
   TORCH_CUDA_ARCH_LIST='7.5;8.0;8.6+PTX' PATH=/usr/local/cuda-11.8/bin:$PATH LD_LIBRARY_PATH=/usr/local/cuda-11.8/lib64:$LD_LIBRARY_PATH \
     micromamba run -n omnigibson pip install \
-    git+https://github.com/StanfordVL/curobo@6a4eb2ca8677829b0f57451ad107e0a3186525e9#egg=nvidia_curobo \
+    git+https://github.com/StanfordVL/curobo@cbaf7d32436160956dad190a9465360fad6aba73#egg=nvidia_curobo \
```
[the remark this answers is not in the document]
ChengshuLi: great point, i really don't know. Cem just suggested me to manually fix this every time we have a new commit on curobo repo.. I can discuss with Cem more later today.

omnigibson/action_primitives/starter_semantic_action_primitives.py
ChengshuLi: Correct. Right now they are the case for R1 and Tiago. Josiah is making sure that for all robots, this is the case (i.e. gripper z-axis point out of the gripper, y-axis point from left finger to right finger, x-axis pointing upward)

omnigibson/action_primitives/starter_semantic_action_primitives.py
ChengshuLi: Yeah I was mistaken before. curobo did do internal batching, but it won't return the result (it will just throw away the result)

omnigibson/action_primitives/starter_semantic_action_primitives.py
ChengshuLi: added a TODO

setup.py
ChengshuLi: This is the latest commit on main lol

tests/test_symbolic_primitives.py
ChengshuLi: Yeah can you file a small PR to fix this. I don't want to be blocked by this.

ChengshuLi (approved):

omnigibson/action_primitives/starter_semantic_action_primitives.py
ChengshuLi: Added TODO.

PR #1033, opened 2024-11-21
fix torch compiled transform utils

ChengshuLi:


omnigibson/utils/transform_utils.py
```diff
@@ -481,8 +481,8 @@ def vec2quat(vec: torch.Tensor, up: torch.Tensor = torch.tensor([0.0, 0.0, 1.0])
     if up.dim() == 1:
         up = up.unsqueeze(0)
 
-    vec_n = torch.nn.functional.normalize(vec, dim=-1)
-    up_n = torch.nn.functional.normalize(up, dim=-1)
+    vec_n = torch.nn.functional.normalize(vec, dim=-1, eps=1e-10)
```
cgokmen, line 484: Can you change this to also use `normalize` or just do it manually with linalg.norm? We shouldn't use this nn library here - they usually have a lot of gradient-related bloat.

hang-yin (approved):

PR #1042, opened 2024-12-03
fix flaky tests

ChengshuLi:


tests/test_data_collection.py
```diff
@@ -7,6 +7,9 @@
 from omnigibson.envs import DataCollectionWrapper, DataPlaybackWrapper
 from omnigibson.macros import gm
 from omnigibson.objects import DatasetObject
+from omnigibson.utils.ui_utils import create_module_logger
+
+log = create_module_logger(module_name=__name__)
```
hang-yin, line 12: Do we want logging within test environments?

hang-yin (approved):

tests/test_data_collection.py
ChengshuLi: removed

PR #1063, opened 2024-12-25
fix cloth loading problems: caused by non-watertight, and potential multiple pieces

Jianghanxiao:
1. Fix the clothes loading problem due to the non-watertight original mesh
2. Fix the clothes with multiple pieces: only keep the largest piece
3. Tune the default parameters for clothes


cgokmen:
I'm merging this into my WIP branch for asset export. Will include it in that PR later. Thanks!

PR #1065, opened 2024-12-27
Enable and comply with ruff linter

cgokmen:
Switches us from black and isort to ruff, which is a combined linter and formatter. For now, this PR enables just the default checks, including unused imports and variables, star imports, `__all__` statements etc. - already caught at least 3-4 bugs just while fixing the linter results. Later on we can continue enabling more features as we go.


omnigibson/__init__.py
```diff
@@ -24,15 +26,14 @@
...
-import nest_asyncio
-
 nest_asyncio.apply()
 
 __version__ = "1.1.1"
 
 root_path = os.path.dirname(os.path.realpath(__file__))
 
 # Store paths to example configs
+# TODO: Move this elsewhere.
```
hang-yin, line 36: Do we want to do this in this PR?

hang-yin (approved):

omnigibson/__init__.py
cgokmen: Lets do it later.

PR #1087, opened 2025-02-04
New cloth mechanisms, fixes to asset import, new meta link format, and more

cgokmen:
* Introduced Yunzhu's group's cloth contributions into the ClothPrim class
* Fix issues in asset import
* Make asset import compatible with fillable meta links baked into the URDF from previous stages
* Update code to make use of new meta link metadata, new names for fields, etc.
* Unified the use of the phrase `meta link` into two words e.g. separated with an underscore for consistent find/replace
* torch.compile has a runtime dependency on torch.dynamo which does not seem to work on Windows. It also seems more unstable than we'd like. We also dont have the strict performance requirements on this code anymore. Switching back to JIT.

TODO:
- [x] fix init py changes (these break pipeline)
- [x] add backwards compatibility with existing dataset formats
- [x] fix jitscript errors


omnigibson/object_states/attached_to.py
```diff
@@ -76,10 +76,10 @@ def remove(self):
...
 
     @classproperty
-    def metalink_prefix(cls):
+    def meta_link_type(cls):
         """
         Returns:
-            str: Unique keyword that defines the metalink associated with this object state
+            str: Unique keyword that defines the meta link associated with this object state
         """
         return m.ATTACHMENT_LINK_PREFIX
```
ChengshuLi, line 84: should these macros still be called `XXX_PREFIX`?

omnigibson/prims/geom_prim.py
```diff
@@ -50,8 +50,9 @@ def _post_load(self):
         # run super first
         super()._post_load()
 
-        # By default, GeomPrim shows up in the rendering.
-        self.purpose = "default"
```
ChengshuLi, line 54: Is this deletion intended?

```
        self.purpose = "default"
```

omnigibson/prims/geom_prim.py
```diff
@@ -50,8 +50,9 @@ def _post_load(self):
...
+        # Temporarily set the opacity threshold to be 0.1. This will be automated in later exports.
+        if self.material:
+            self.material.opacity_threshold = 0.1
```
ChengshuLi, line 55: Can you add some comments? e.g. ray traicing can only handle binary opacity using `opacity_threshold`

omnigibson/prims/material_prim.py
```diff
@@ -231,6 +231,20 @@ def set_input(self, inp, val):
...
+            if isinstance(val, str):
+                input_type = lazy.pxr.Sdf.ValueTypeNames.String
+            elif isinstance(val, int):
+                input_type = lazy.pxr.Sdf.ValueTypeNames.Int
+            elif isinstance(val, float):
+                input_type = lazy.pxr.Sdf.ValueTypeNames.Float
+            elif isinstance(val, bool):
+                input_type = lazy.pxr.Sdf.ValueTypeNames.Bool
+            else:
+                raise ValueError(f"Invalid input type {type(val)} for input {inp}")
```
ChengshuLi, line 245: I remember Hang has a helper function that does this. If not, maybe put this into a helper function in one of the utils files?

omnigibson/prims/rigid_prim.py
```diff
@@ -796,6 +798,51 @@ def visual_aabb_center(self):
...
+        assert self.is_meta_link, f"{self.name} is not a meta link"
+        if self.prim.HasAttribute("ig:metaLinkType"):
+            return self.get_attribute("ig:metaLinkType")
+
+        # Check using the old format.
+        # TODO: Remove this after the next dataset release
+        return m.LEGACY_META_LINK_PATTERN.fullmatch(self.name).group(1)
+
+    @cached_property
+    def meta_link_id(self):
```
ChengshuLi, line 827: Can we add some docstring about meta_link_id and sub_id? I actually don't quite remember what these are..

omnigibson/utils/asset_conversion_utils.py
```diff
@@ -282,6 +282,13 @@ def _set_mtl_opacity(mtl_prim, texture):
     lazy.omni.usd.create_material_input(mtl_prim, mtl, texture, lazy.pxr.Sdf.ValueTypeNames.Asset)
     lazy.omni.usd.create_material_input(mtl_prim, "enable_opacity", True, lazy.pxr.Sdf.ValueTypeNames.Bool)
     lazy.omni.usd.create_material_input(mtl_prim, "enable_opacity_texture", True, lazy.pxr.Sdf.ValueTypeNames.Bool)
+
+    # Set the opacity to use the alpha channel for its mono-channel value.
+    # This defaults to some other value, which takes opaque black channels in the
+    # image to be fully transparent. This is not what we want.
+    lazy.omni.usd.create_material_input(mtl_prim, "opacity_mode", 0, lazy.pxr.Sdf.ValueTypeNames.Int)
+    lazy.omni.usd.create_material_input(mtl_prim, "opacity_threshold", 0.1, lazy.pxr.Sdf.ValueTypeNames.Float)
```
ChengshuLi, line 290: Why is this done twice in the code base? Are both needed?

omnigibson/utils/asset_conversion_utils.py
```diff
@@ -1112,7 +1099,7 @@ def _create_urdf_import_config(
     import_config.set_merge_fixed_joints(merge_fixed_joints)
     import_config.set_convex_decomp(use_convex_decomposition)
     import_config.set_fix_base(False)
-    import_config.set_import_inertia_tensor(False)
+    import_config.set_import_inertia_tensor(True)
```
ChengshuLi, line 1102: I assume this change is intended?

omnigibson/prims/cloth_prim.py
```diff
@@ -88,8 +108,31 @@ def _post_load(self):
         if "mass" in self._load_config and self._load_config["mass"] is not None:
             self.mass = self._load_config["mass"]
 
+        # Save the default point configuration
+        self.save_configuration("default", self.points)
+
+        # Remesh the object if necessary
+        force_remesh = self._load_config.get("force_remesh", False)
```
ChengshuLi, line 115: what's "force_remesh" for?

omnigibson/prims/cloth_prim.py
```diff
@@ -136,6 +179,361 @@ def _post_load(self):
...
+        for config in CLOTH_CONFIGURATIONS:
+            attr_name = f"points_{config}"
+            if self.prim.HasAttribute(attr_name):
+                self.prim.RemoveProperty(attr_name)
+
+        # Then update the configuration-dependent fields
+        self.save_configuration("default", th.tensor(tm.vertices, dtype=th.float32))
+
+        # Then update the points to the default configuration
+        self.reset_points_to_configuration("default")
```
ChengshuLi, line 299: Is this needed?

omnigibson/prims/cloth_prim.py
```diff
@@ -136,6 +179,361 @@ def _post_load(self):
...
+        tmp_dir = os.path.join(tempfile.gettempdir(), tmp_name)
+        tmp_fpath = os.path.join(tmp_dir, f"{tmp_name}.obj")
+        Path(tmp_dir).mkdir(parents=True, exist_ok=True)
+        tm.export(tmp_fpath)
+
+        # Start with the default particle distance
+        particle_distance = self.cloth_system.particle_contact_offset * 2 / 1.5
+
+        # Repetitively re-mesh at lower resolution until we have a mesh that has less than MAX_CLOTH_PARTICLES vertices
+        import pymeshlab  # We import this here because it takes a few seconds to load.
```
ChengshuLi, line 204: sorry I am not sure I fully understand this. what benefit do we gain by importing here exactly?

ChengshuLi (commented):
LGTM with some minor comments!

omnigibson/object_states/attached_to.py
cgokmen: Changed.

omnigibson/prims/cloth_prim.py
cgokmen: In the pipeline, we need to force this class to remesh the object even though the scale is 1, 1, 1. That also happens through this logic so we need to allow forcing it.

omnigibson/prims/cloth_prim.py
cgokmen: I didn't add this, I just moved it over from the micro particle system class. But I'm guessing pymeshlab takes a few seconds to import so the author wanted to save some time by only importing it if remeshing is needed (now VERY infrequently). I guess pymeshlab is actually now an optional dependency so we could even remove it from the setup.py later if needed.

omnigibson/prims/cloth_prim.py
cgokmen: This is what actually sets the `points` attribute to also use the remeshed configuration (note that we don't set it elsewhere here). I guess it's kind of a sanitycheck too.

omnigibson/prims/material_prim.py
cgokmen: He does in an unmerged PR (the 4.2 and 4.5 branches), yes - he'll merge that PR in a bit so this can quickly get removed when he does that. I can't put it here because it uses the new API for checking what inputs are available which is 4.2+.

omnigibson/utils/asset_conversion_utils.py
cgokmen: This is the actually needed one (for the pipeline). The other one is a temporary hotfix that I'll remove after the next export.

omnigibson/utils/asset_conversion_utils.py
cgokmen: Yes, intended. This lets the moment of inertia tensors from the URDF be used.

omnigibson/prims/geom_prim.py
cgokmen: Yes, it's intended. If we don't do this, all visual meta links (e.g. inside volumes) are forced to visibility.

ChengshuLi (approved):

PR #1094, opened 2025-02-07
Isaac 4.5 Upgrade

hang-yin:
TODO:
- [ ] omni hydra log suppress does not work (for create_joint)
- [x] verify glass material change
- [x] revisit kit file - add remote streaming and VR extensions
- [x] refactor rigid prim to 1) RigidPrim 2) RigidBodyPrim 3) RigidKinematicPrim
- [x] curobo test bug
- [x] check point instancer particle system get/set pose methods when flatcache is on
- [ ] double check installation
- [x] VR? Or should this be a separate PR?
- [ ] check warnings in controllers test
- [ ] open a ticket for particle applier visualization
- [ ] open a ticket for adding common community questions to `Known Issues` doc page
- [ ] open a ticket to clean up VR teleop

Major changes:
- Refactor RigidPrim into 1) RigidPrim 2) RigidBodyPrim (with RigidPrimView) 3) RigidKinematicPrim (kinematic-only)
- Isaacsim module renaming
- Remove particle applier visualization for now (deprecated since 4.2)
- Material handling
- Enable point-instancer particle systems when flatcache is on
- Make sure physics view is always valid when needed (Isaac 4.5 is a lot more strict with invalidating the physics view; this is a problem for us since we do many physics-breaking things when simulation is playing; the fix here is to always re-initialize the physics view whenever we break it; some common reasons why the physics view break: 1) prim deletion/addition 2) AG release)


omnigibson/prims/rigid_kinematic_prim.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        Kinematic objects don't have dynamic velocities
+
+        Args:
+            clone (bool): Whether to clone the internal buffer or not when grabbing data (ignored)
+
+        Returns:
+            th.tensor: zero velocity. Shape (3,).
+        """
+        return th.zeros(3)
```
hang-yin, line 100: Question: we can either keep these no-op methods for kinematic-only rigid prims or force upstream methods to check whether this is kinematic-only or not. What do people think?

omnigibson/object_states/particle_modifier.py
```diff
@@ -1035,79 +1033,6 @@ def _initialize(self):
         # This will initialize the system if it's not initialized already.
         system = self.obj.scene.get_system(system_name)
 
-        if self.visualize:
```
cremebrule, line 1038: is this being removed because it doesn't work, or because our paradigm changed (e.g.: our assets now come shipped with the particle visualization mesh attached?)

omnigibson/objects/dataset_object.py
```diff
@@ -214,12 +214,8 @@ def recursive_light_update(child_prim):
 
         # Apply any forced roughness updates
         for material in self.materials:
-            if (
-                "reflection_roughness_texture_influence" in material.shader_input_names
```
cremebrule, line 218: are these values guaranteed to always exist in the material?

omnigibson/omnigibson_4_5_0.kit
```diff
@@ -0,0 +1,39 @@
+[package]
```
cremebrule, line 1: should we also delete the outdated kit files?

omnigibson/prims/entity_prim.py
```diff
@@ -219,39 +227,44 @@ def update_links(self):
...
-                "belongs_to_articulation": self._articulation_view is not None and link_name != self._root_link_name,
+                "kinematic_only": is_kinematic,
+                "belongs_to_articulation": self._articulation_view is not None and not is_root_link,
                 "remesh": self._load_config.get("remesh", True),
                 "xform_props_pre_loaded": self._load_config.get("xform_props_pre_loaded", False),
                 "scale": self._load_config.get("scale", None),
             }
+
+            # Determine the correct class based on link type and kinematic property
+            if link_type == "rigid":
```
cremebrule, line 262: hmmm is there a better way than checking directly for a string? Maybe a Enum or something?

omnigibson/prims/material_prim.py
```diff
@@ -194,7 +196,7 @@ def shader_update_asset_paths_with_root_path(self, root_path, relative=False):
                 Otherwise, @root_path will be pre-appended to the original asset paths
         """
 
-        for inp_name in self.shader_input_names_by_type("SdfAssetPath"):
+        for inp_name in self.get_shader_input_names_by_type("SdfAssetPath", include_default=True):
```
cremebrule, line 199: what does this extra flag do?

omnigibson/prims/rigid_body_prim.py
```diff
@@ -0,0 +1,271 @@
...
+import torch as th
+
+import omnigibson as og
+import omnigibson.lazy as lazy
+from omnigibson.utils.usd_utils import PoseAPI
+
+from .rigid_prim import RigidPrim
+
+
+class RigidBodyPrim(RigidPrim):
```
cremebrule, line 13: nit: can we rename this to be RigidDynamicPrim? That way the distinction is clear between Kinematic <--> Dynamic, which also aligns with how omni docs describe the two

omnigibson/prims/rigid_body_prim.py
```diff
@@ -0,0 +1,271 @@
...
+    dynamics like gravity, forces, and collisions.
+
+    Args:
+        relative_prim_path (str): Scene-local prim path of the Prim to encapsulate or create.
+        name (str): Name for the object. Names need to be unique per scene.
+        load_config (None or dict): If specified, should contain keyword-mapped values that are relevant for
+            loading this prim at runtime.
+    """
+
+    def __init__(
```
cremebrule, line 26: i assume the bulk of this is copied vertabim from the original RigidPrim class lol

omnigibson/prims/rigid_kinematic_prim.py
```diff
@@ -0,0 +1,235 @@
...
+        pass
+
+    def set_linear_velocity(self, velocity):
+        """
+        Kinematic objects don't have dynamic velocities - this is a no-op
+
+        Args:
+            velocity (th.tensor): linear velocity to set (ignored). Shape (3,).
+        """
+        pass
```
cremebrule, line 67: should we raise an error here?

omnigibson/prims/rigid_kinematic_prim.py
```diff
@@ -0,0 +1,235 @@
...
+        return th.zeros(3)
+
+    def set_angular_velocity(self, velocity):
+        """
+        Kinematic objects don't have dynamic velocities - this is a no-op
+
+        Args:
+            velocity (th.tensor): angular velocity to set (ignored). Shape (3,).
+        """
+        pass
```
cremebrule, line 88: same here

omnigibson/prims/rigid_kinematic_prim.py
cremebrule: I'm inclined towards the latter to avoid silent bugs from occurring (or rather, unintended / unexpected behaviors)

omnigibson/robots/manipulation_robot.py
```diff
@@ -613,6 +613,7 @@ def _release_grasp(self, arm="default"):
 
         # Remove joint and filtered collision restraints
         og.sim.stage.RemovePrim(self._ag_obj_constraint_params[arm]["ag_joint_prim_path"])
+        og.sim.update_handles()
```
cremebrule, line 616: what was this needed for? was this because we just removed a prim?

omnigibson/simulator.py
```diff
@@ -143,22 +142,21 @@ def _launch_app():
...
-    assert lazy.omni.isaac.core.utils.stage.close_stage()
+    assert lazy.isaacsim.core.utils.stage.close_stage()
 
     # Omni overrides the global logger to be DEBUG, which is very annoying, so we re-override it to the default WARN
     # TODO: Remove this once omniverse fixes it
     logging.getLogger().setLevel(logging.WARNING)
 
     # Enable additional extensions we need
-    lazy.omni.isaac.core.utils.extensions.enable_extension("omni.flowusd")
-    lazy.omni.isaac.core.utils.extensions.enable_extension("omni.particle.system.bundle")
```
cremebrule, line 157: seems like we don't have the particle extension still?? )):

omnigibson/simulator.py
```diff
@@ -1023,7 +1038,9 @@ def play(self):
                 if gm.ENABLE_FLATCACHE:
                     channels.append("omni.physx.plugin")
                 with suppress_omni_log(channels=channels):
-                    super().play()
+                    # Playing takes some steps, so we temporarily change the dt to 0 to avoid moving things.
+                    with self.slowed():
```
cremebrule, line 1042: FYI when i tried slowing and unslowing the sim back years ago (maybe during the 4.0.0 or even before??), the resulting unslowed-sim could be unstable. not sure if that's still an issue?

omnigibson/simulator.py
```diff
@@ -1279,18 +1306,16 @@ def paused(self):
...
             instantaneous kinematics with minimal impact on physics propagation.
 
-            NOTE: This will set both the physics dt and rendering dt to the same value during this scope.
-
             Upon leaving the scope, the prior simulator state is restored.
             """
             # Set dt, yield, then restore the original dt
             physics_dt, rendering_dt = self.get_physics_dt(), self.get_rendering_dt()
-            self.set_simulation_dt(physics_dt=dt, rendering_dt=dt)
+            self._physics_context.set_physics_dt(slow_dt, 1)
```
cremebrule, line 1318: won't this mean that taking a sim.step() (which calls og.app.update()) will take a ton of rendering steps (`1 / slow_dt`)?

omnigibson/systems/micro_particle_system.py
```diff
@@ -473,10 +473,8 @@ def initialize(self, scene):
         super().initialize(scene)
 
         # Run sanity checks
-        if not gm.USE_GPU_DYNAMICS or gm.ENABLE_FLATCACHE:
-            raise ValueError(
-                f"Failed to initialize {self.name} system. Please set gm.USE_GPU_DYNAMICS=True and gm.ENABLE_FLATCACHE=False."
-            )
+        if not gm.USE_GPU_DYNAMICS:
+            raise ValueError(f"Failed to initialize {self.name} system. Please set gm.USE_GPU_DYNAMICS=True.")
```
cremebrule, line 477: so does this mean fluids has been confirmed to work with flatcache on?? :D

omnigibson/systems/system_base.py
```diff
@@ -203,7 +203,14 @@ def _clear(self):
...
         self.reset()
-        lazy.omni.isaac.core.utils.prims.delete_prim(self.prim_path)
+        lazy.isaacsim.core.utils.prims.delete_prim(self.prim_path)
+        og.sim.update_handles()
+
+        # This is a bit of a hack to address a very niche situation where we clear a system and immediately reinitialize it
+        # If we do not take a physics step here, the system will not be reinitialized properly
+        # We need to find a way around this since this causes some physics leakage
+        with og.sim.slowed():
+            og.sim.step()
```
cremebrule, line 213: can we just take a physics step instead of a whole og step?

cremebrule: in general this applies to whenever we're running `og.sim.slowed()` elsewhere

omnigibson/utils/deprecated_utils.py
```diff
@@ -82,96 +78,6 @@ def __init__(self, prim_type: str, **kwargs):
         assert isinstance(self._evaluator_class, type)
 
 
-class Utils(OmniUtils):
```
cremebrule, line 85: nice!!! glad we could finally get rid of this

omnigibson/utils/deprecated_utils.py
```diff
@@ -869,7 +788,9 @@ def get_world_poses(
              [ 1.0000000e+00 -9.5171310e-07 -2.2615541e-07  5.5922797e-08]
              [ 1.0000000e+00 -7.9806580e-07 -1.3064776e-07  5.3154917e-08]]
         """
-        if not omni.timeline.get_timeline_interface().is_stopped() and self._physics_view is not None:
```
cremebrule, line 872: in general, do we not need to query the timeline interface anymore?

omnigibson/utils/physx_utils.py
```diff
@@ -93,6 +93,7 @@ def bind_material(prim_path, material_path):
         material_path=material_path,
         strength=None,
     )
+    og.sim.update_handles()
```
cremebrule, line 96: why do we need to update handles after binding a material?

omnigibson/utils/usd_utils.py
```diff
@@ -746,6 +747,9 @@ def mark_valid(cls):
     @classmethod
     def _refresh(cls):
         if og.sim is not None and not cls.VALID:
+            # Check that no reads from PoseAPI are happening during a physics step.
+            assert not og.sim.currently_stepping, "Cannot refresh poses during a physics step!"
```
cremebrule, line 751: what's the reason for this? I assume this would possibly only occur during the physics callback, i.e.: the robot controller callabck code?

cremebrule (commented):
awesome work!! I don't think I have any major comments, just a lot of clarification questions. Happy to discuss in person too!

omnigibson/prims/rigid_kinematic_prim.py
hang-yin: Sounds good, just made the change

omnigibson/object_states/particle_modifier.py
hang-yin: This stopped working since 4.2, will open a ticket to approximate this.

omnigibson/objects/dataset_object.py
hang-yin: These always exist now if the material isn't glass. Just added a check for this.

omnigibson/omnigibson_4_5_0.kit
hang-yin: Done

omnigibson/prims/entity_prim.py
hang-yin: Rewrote this with enum

omnigibson/prims/material_prim.py
hang-yin: This was a change made on the 4.2 branch, which essentially go through all shader input names, including the default. Before 4.2, many inputs are somehow automatically created, but they changed it in a way that make a lot of the inputs we need "missing" (they stay in the default inputs, not 'initialized' in a sense). I understand it as a form of lazy material input loading. This flag essentially tell us whether to include all the uninitialized inputs.

omnigibson/robots/manipulation_robot.py
hang-yin: Exactly! In this new version, many operations, e.g. remove prim, add prim, and material binding would all "destroy" the physics sim view.

omnigibson/simulator.py
hang-yin: Sadly no, and it seems like this extension might be gone forever. The isaac people said they were going to look into it but no luck yet.

omnigibson/utils/deprecated_utils.py
hang-yin: I guess not, except in deprecated utils?

omnigibson/simulator.py
```diff
@@ -748,13 +747,30 @@ def adding_objects(self, objs):
             Args:
                 objs (Iterable[BaseObject]): list of objects to add
             """
-            # Immediately yield
+            SimulationManager = lazy.isaacsim.core.simulation_manager.SimulationManager
```
cgokmen, line 750: It makes no sense to have this context with the objs arg, because if you already know how many objects you're importing you can just call post_import yourself. Let's leave a TODO here to remove this context and call post_import directly.

omnigibson/simulator.py
cgokmen: It would take a ton of physics step you mean? I think that's true. @hang-yin can you also update the render step to match here?

cgokmen (approved):

omnigibson/prims/prim_base.py
```diff
@@ -4,7 +4,11 @@
 import omnigibson.lazy as lazy
 from omnigibson.utils.python_utils import Recreatable, Serializable
 from omnigibson.utils.ui_utils import create_module_logger
-from omnigibson.utils.usd_utils import delete_or_deactivate_prim, scene_relative_prim_path_to_absolute
+from omnigibson.utils.usd_utils import (
+    delete_or_deactivate_prim,
```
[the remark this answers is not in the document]
cremebrule: Not sure what the current status is, but at the time I had implemented prim deletion there was an issue where certain prims couldn't actually be "truly" deleted -- e.g.: removing a subprim from an imported asset. So there were some APIs that worked better than others at accomplishing this more uniformly.

FWIW I'm ambivalent to whatever API we end up going with for 4.5, as long as the behavior is consistent across different use cases.

omnigibson/prims/rigid_prim.py
```diff
@@ -463,18 +432,20 @@ def has_collision_meshes(self):
...
+        Sets the visual only state of this link
 
         Args:
             val (bool): Whether this link should be a visual-only link (i.e.: no gravity or collisions applied)
         """
         # Set gravity and collisions based on value
         if val:
             self.disable_collisions()
-            self.disable_gravity()
+            if not self.kinematic_only:
```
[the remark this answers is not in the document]
cremebrule: Yeah great point. But in some sense it's weird to actually have APIs are strictly not supported, no? (e.g.: there is no sense of gravity for kinematic only objects, so it's weird to have an API that supports that).

I think the guiding principle I had in mind was clarity for the end user -- what do you think would be best? I think the separate classes are nice but I'm also not a fan of having APIs across classes that clearly don't support those functionalities

omnigibson/prims/prim_base.py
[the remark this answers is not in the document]
cgokmen: @hang-yin let's replace all usages of deleteprim and removeprim with this.

omnigibson/prims/rigid_prim.py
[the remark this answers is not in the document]
cgokmen: Thats fair. @hang-yin let's:

1. remove the self.kinematic_only member.
2. in every entityprim use case that uses self.kinematic_only, lets instead check if the rigid prim isinstance(dynamic) or isinstance(kinematic) etc. and call the function only then
3. in every rigidprim use case (like this one), implement every kinematic/dynamic dependent function (like this above) separately in the two subclasses instead of relying on a check in the common code

omnigibson/examples/teleoperation/vr_robot_control_demo.py
```diff
@@ -0,0 +1,99 @@
...
+                "name": "InverseKinematicsController",
+                "mode": "absolute_pose",
+                "command_input_limits": None,
+                "command_output_limits": None,
+            },
+            "gripper_left": {"name": "MultiFingerGripperController", "command_input_limits": "default"},
+            "gripper_right": {"name": "MultiFingerGripperController", "command_input_limits": "default"},
+        },
+        "action_normalize": False,
+        "reset_joint_pos": [
```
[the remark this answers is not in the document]
hang-yin: Ooooh this is just here so that the robot "kneels" down a little. It R1 stands up straight, it's relatively hard to see and grasp.

omnigibson/prims/rigid_dynamic_prim.py
```diff
@@ -0,0 +1,391 @@
...
+    @property
+    def mass(self):
+        """
+        Returns:
+            float: mass of the rigid body in kg.
+        """
+        mass = self._rigid_prim_view.get_masses()[0]
+
+        # Fallback to analytical computation of volume * density
+        if mass == 0:
```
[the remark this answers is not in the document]
hang-yin: What do you think we should have it here though? Do you think I should just assert mass doesn't equal to zero if that's not expected?

omnigibson/simulator.py
hang-yin: There isn't a set_rendering_dt, do you mean reverting back to do `set_simulation_dt` with both physics dt and rendering dt?

omnigibson/prims/entity_prim.py
```diff
@@ -1,6 +1,7 @@
 import math
 from functools import cached_property
 from typing import Literal
+from enum import Enum, auto
```
cgokmen, line 4: Remove. Doesn't `ruff` catch this? Do you have the precommit set up?

omnigibson/prims/entity_prim.py
```diff
@@ -219,39 +220,44 @@ def update_links(self):
...
             # Fixed child links of kinematic-only objects are not kinematic-only, to avoid the USD error:
             # PhysicsUSD: CreateJoint - cannot create a joint between static bodies, joint prim: ...
+            is_kinematic = self._load_config.get("kinematic_only", False) if is_root_link else False
+
             link_load_config = {
-                "kinematic_only": (
-                    self._load_config.get("kinematic_only", False) if link_name == self._root_link_name else False
-                ),
-                "belongs_to_articulation": self._articulation_view is not None and link_name != self._root_link_name,
+                "kinematic_only": is_kinematic,
```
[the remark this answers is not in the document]
cgokmen: Let's actually remove kinematic_only members everywhere. Search and destroy `.kinematic_only`

omnigibson/prims/rigid_kinematic_prim.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Enables gravity for this rigid body
+        """
+        pass
+
+    def disable_gravity(self):
+        """
+        Disables gravity for this rigid body
+        """
+        pass
```
cgokmen, line 167: Are these necessary?

omnigibson/prims/rigid_prim.py
```diff
@@ -155,10 +138,6 @@ def _initialize(self):
             for mesh in mesh_group.values():
                 mesh.initialize()
 
-        # Get contact info first
-        if self.contact_reporting_enabled:
-            og.sim.contact_sensor.get_rigid_body_raw_data(self.prim_path)
```
cgokmen, line 160: What happened to this?

omnigibson/utils/usd_utils.py
```diff
@@ -602,10 +604,10 @@ def clear(cls):
         """
         # Remove all the collision group prims
         for col_group_prim in cls.ACTIVE_COLLISION_GROUPS.values():
-            og.sim.stage.RemovePrim(col_group_prim.GetPath().pathString)
+            delete_or_deactivate_prim(col_group_prim.GetPath().pathString)
 
         # Remove the collision groups tree
-        og.sim.stage.RemovePrim("/World/collision_groups")
+        delete_or_deactivate_prim("/World/collision_groups")
```
cgokmen, line 610: These are problematic: in a multiple envs context delete_or_deactivate_prim will get called lots of times which will mean it will call its downstream update_handles implication multiple times. Instead, let's make it accept a list of prim paths (delete_or_deactivate_prims), call it with a flattened list of prim paths everywhere it's called in an obvious loop, and make it call update handles just once.

omnigibson/simulator.py
cgokmen: Lets discuss this quickly tomorrow. What is sim_step_dt, why does it not change too?

omnigibson/prims/entity_prim.py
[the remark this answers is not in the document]
cremebrule: @cgokmen  I am discussing with @hang-yin , would it make more sense to keep the kinematic_only flag at the object level?

We do want to keep track of which objects are not dynamic, because it has implications for other physics-related features, eg. sleeping, views, etc.

omnigibson/prims/rigid_kinematic_prim.py
hang-yin: This is added because of this comment: https://github.com/StanfordVL/OmniGibson/pull/1094#discussion_r1994426070
The visual_only setter calls enable and disable gravity

omnigibson/prims/rigid_dynamic_prim.py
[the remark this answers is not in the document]
hang-yin: For our reference: 
```
    @property
    def mass(self):
        """
        Returns:
            float: mass of the rigid body in kg.
        """
        mass = self._rigid_prim_view.get_masses()[0]

        # Fallback to analytical computation of volume * density
        if mass == 0:
            return self.volume * self.density

        return mass

    @property
    def density(self):
        """
        Returns:
            float: density of the rigid body in kg / m^3.
        """
        mass = self._rigid_prim_view.get_masses()[0]
        # We first check if the mass is specified, since mass overrides density. If so, density = mass / volume.
        # Otherwise, we try to directly grab the raw usd density value, and if that value does not exist,
        # we return 1000 since that is the canonical density assigned by omniverse
        if mass != 0.0:
            density = mass / self.volume
        else:
            density = self._rigid_prim_view.get_densities()[0]
            if density == 0.0:
                density = 1000.0

        return density
```
I might be missing something but I don't see why this can cause infinite recursion. If `self._rigid_prim_view.get_masses()[0]` returns zero, and we call the density getter, it doesn't seem to be making any recursive calls. Is that what you mean here?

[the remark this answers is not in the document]
cgokmen: Hm yeah it doesnt use self.mass. Makes sense.

PR #1120, opened 2025-02-18
Fixes/curobo particle

ChengshuLi:


omnigibson/action_primitives/curobo.py
```diff
@@ -197,6 +198,16 @@ def __init__(
...
 
+            # Make sure all cuda graphs have been warmed up
+            for solver in [mg.ik_solver, mg.trajopt_solver, mg.finetune_trajopt_solver]:
+                if solver.solver.use_cuda_graph_metrics:
+                    assert solver.solver.safety_rollout._metrics_cuda_graph_init
+                    if isinstance(solver, lazy.curobo.wrap.reacher.trajopt.TrajOptSolver):
+                        assert solver.interpolate_rollout._metrics_cuda_graph_init
+                for opt in solver.solver.optimizers:
+                    if opt.use_cuda_graph:
+                        assert opt.cu_opt_init
```
cremebrule, line 209: rubber stamping this, as i have no idea what this does LOL

omnigibson/systems/micro_particle_system.py
```diff
@@ -1444,14 +1443,9 @@ def use_isosurface(self):
...
     def particle_radius(self):
         # Magic number from omni tutorials
-        # See https://docs.omniverse.nvidia.com/prod_extensions/prod_extensions/ext_physics.html#offset-autocomputation
+        # See https://docs.omniverse.nvidia.com/extensions/latest/ext_physics/physics-particles.html#offset-autocomputation
         return 0.99 * 0.6 * self._particle_contact_offset
 
-    @property
-    def particle_particle_rest_distance(self):
-        # Magic number, based on intuition from https://docs.omniverse.nvidia.com/extensions/latest/ext_physics/physics-particles.html#particle-particle-interaction
-        return self.particle_radius * 2.0 * m.FLUID_PARTICLE_PARTICLE_DISTANCE_SCALE
```
cremebrule, line 1453: is this not needed anymore?

omnigibson/utils/usd_utils.py
```diff
@@ -519,11 +519,7 @@ def get_column_filters(cls):
 
     @classmethod
     def get_max_contact_data_count(cls):
-        # 4x per finger link, to be safe.
-        # 4 here is not the finger count, it's the number of items we will record contacts with, per finger.
-        # e.g. it's N such that if the finger is touching more than N items at once, only the first N are recorded.
-        # This number should very rarely go above 4.
-        return len(cls.get_column_filters()[0]) * 4
+        return len(cls.get_column_filters()[0]) * 128
```
cremebrule, line 522: why does this need to change?

tests/test_curobo.py
```diff
@@ -387,7 +382,9 @@ def test_curobo():
                     eef_link_name = robot.eef_link_names[arm_name]
                     marker.set_position_orientation(position=target_pos[eef_link_name][traj_idx])
 
-                q_traj = cmg.path_to_joint_trajectory(traj_path)
+                q_traj = cmg.path_to_joint_trajectory(traj_path).cpu().float()
+                if not bypass_physics:
+                    q_traj = cmg.add_linearly_interpolated_waypoints(traj=q_traj, max_inter_dist=0.01)
```
cremebrule, line 387: maybe add comment here? i assume this is done to avoid jerky motions

cremebrule (approved):
just a few clarification questions / comments, otherwise lgtm!

omnigibson/action_primitives/curobo.py
ChengshuLi: it made sure that the warmup is actually successful (and the cuda graphs are built correctly). Previously it's not successful because our retract_config is wrong.

omnigibson/systems/micro_particle_system.py
ChengshuLi: No, I can no longer find any reference to the arbitrary 0.8 scale on their documentation. Empirically, if we keep this 0.8 scale, the sampled particles (e.g. filling a container) will be too close together, which leads to violent explosion when physics starts.

omnigibson/utils/usd_utils.py
ChengshuLi: If there are many contact points during assisted or sticky grasping, we will exceed the limits, which makes the grasping not work.

PR #1132, opened 2025-02-27
Add support for untextured skybox and checkerboard floor

cgokmen:
Both of these improvements are meant primarily to help improve VLM and other pretrained models' ability to interpret an object. The replaced floor plane helps understand object scale and make a backdrop on which any colored object is visible (although we can make the different scales clearer later e.g. which one is which). The skybox thing fixes the too-yellow lighting and keeps us from having to use sky with clouds on the back. It also switches us from the quite uninterpretable skybox color mode (which is NOT in regular 0-1 RGB but 0-2??) to color temperature.

Example object:

image


omnigibson/examples/objects/load_object_selector.py
```diff
@@ -48,6 +48,9 @@ def main(random_selection=False, headless=False, short_exec=False):
     }
     if scene_type == "InteractiveTraversableScene":
         cfg["scene"]["scene_model"] = "Rs_int"
+    else:
+        cfg["scene"]["floor_plane_grid"] = True
+        cfg["scene"]["use_skybox_texture"] = False
```
ChengshuLi, line 53: should we keep the original behavior as default? I think the skybox actually looks better than the checkerboard, but I guess "better" is also subjective.

omnigibson/scenes/scene_base.py
```diff
@@ -59,20 +59,35 @@ def __init__(
         self,
         scene_file=None,
         use_floor_plane=True,
+        floor_plane_size=1000.0,
```
ChengshuLi, line 62: what is the unit? meter? Can you add a comment if so?

omnigibson/scenes/scene_base.py
```diff
@@ -59,20 +59,35 @@ def __init__(
...
         floor_plane_visible=True,
         floor_plane_color=(1.0, 1.0, 1.0),
+        floor_plane_grid=False,
+        floor_plane_grid_resolutions=(0.001, 0.01, 0.1, 1.0),
         use_skybox=True,
+        use_skybox_texture=True,
```
ChengshuLi, line 68: hmm I am not fan of a variable named `use_skybox_texture` can be True/False/an string. Can we create another variable for the string case?

omnigibson/scenes/scene_base.py
```diff
@@ -59,20 +59,35 @@ def __init__(
...
+            floor_plane_grid (bool): whether to render a checkerboard floor plane with grid squares
+            floor_plane_grid_resolutions (Tuple[float]): A tuple of integers where each integer represents a
+                grid cell size, e.g. (0.001, 0.01, 1.) will render a floor plane with 3 different grid cell sizes at 1mm,
+                1cm, and 1m respectively. Will be ignored if @floor_plane_grid is False.
             use_skybox (bool): whether to load a skybox into the simulator
+            use_skybox_texture (Union[str, bool]): whether to use a texture for the skybox or not. False will disable the
+                texture and expose the skybox color, True will use the default texture, and a string will use the
+                specified texture.
+            skybox_color_temperature (float): Color temperature of the skybox if @use_skybox is True. This is the color of the
+                light emitted by the skybox in Kelvin, and the color of the skybox itself if no texture is used.
```
ChengshuLi, line 90: This is also a bit confusing. It's the color of the skybox AND it's the color of the light emitted by the skybox? Shouldn't we then split these into two variables in case they might be different in the future?

omnigibson/utils/usd_utils.py
```diff
@@ -1659,6 +1660,104 @@ def mesh_prim_to_trimesh_mesh(mesh_prim, include_normals=True, include_texcoord=
...
+    face_uvs = []
+
+    # Iterate over cell sizes and the areas they will need to cover (which is the next cell size)
+    last_size_vert_idxes = None
+    for cell_size, area_size in zip(edge_sizes[:-1], edge_sizes[1:]):
+        # First, create the vertices for the area. These are named after the quadrants.
+        new_1 = (area_size, area_size, 0)
+        new_2 = (-area_size, area_size, 0)
+        new_3 = (-area_size, -area_size, 0)
+        new_4 = (area_size, -area_size, 0)
```
ChengshuLi, line 1692: maybe better variable names here?

omnigibson/utils/usd_utils.py
```diff
@@ -1659,6 +1660,104 @@ def mesh_prim_to_trimesh_mesh(mesh_prim, include_normals=True, include_texcoord=
...
+            # If this is the first iteration, we only have the area vertices. Connect them in two
+            # triangles to form the area.
+            face_vert_indices.extend([[new_1_idx, new_2_idx, new_3_idx], [new_1_idx, new_3_idx, new_4_idx]])
+
+            # For this case we can just get the indices of the UV coordinates for the area vertices.
+            repetitions = area_size / cell_size
+            new_1_uv = (repetitions, repetitions)
+            new_2_uv = (0, repetitions)
+            new_3_uv = (0, 0)
+            new_4_uv = (repetitions, 0)
```
ChengshuLi, line 1711: same here? hmm I feel like I don't quite understand this piece of code because of these variable names lol

omnigibson/utils/usd_utils.py
```diff
@@ -1659,6 +1660,104 @@ def mesh_prim_to_trimesh_mesh(mesh_prim, include_normals=True, include_texcoord=
...
+
+            area_face_verts = [
+                [old_1_idx, new_1_idx, new_2_idx],
+                [old_1_idx, new_2_idx, old_2_idx],
+                [old_2_idx, new_2_idx, new_3_idx],
+                [old_2_idx, new_3_idx, old_3_idx],
+                [old_3_idx, new_3_idx, new_4_idx],
+                [old_3_idx, new_4_idx, old_4_idx],
+                [old_4_idx, new_4_idx, new_1_idx],
+                [old_4_idx, new_1_idx, old_1_idx],
```
ChengshuLi, line 1726: holy cow

omnigibson/examples/objects/load_object_selector.py
cgokmen: This example is the whole point of adding the checkerboard, it lets you see the object regardless of its color, get a feel for its size, and also not have the object floating around in front of some clouds. This is the only example that I changed.

omnigibson/scenes/scene_base.py
cgokmen: There isn't a way to separate these two things currently. The skybox emits light in its color.

omnigibson/examples/objects/load_object_selector.py
cremebrule: FYI, slightly unrelated, but we should probably update the default skybox to be an actual HDRI background from [here](https://polyhaven.com/hdris). They're free to use for non-commerical purposes, are incredibly high quality (can go up to 8k), and have nuanced lighting / toning baked in based on the backdrop. Was experimenting with them for our digital cousin project and it's literally a drop in replacement for our current janky skybox that I had originally added to OG lol (literally just replace via `og.sim.skybox.texture_path = ...`)

omnigibson/scenes/scene_base.py
cgokmen: I think for now it's fine to let the texture path come from a macro rather than a config str, it's not quite important to change from a config. So I'm changing this to just True/False.

omnigibson/examples/objects/load_object_selector.py
cgokmen: I tried some of these the other day (from this exact website) - it's kind of hard to get this right. The problem is that many scenes have full height windows, and what you see outside through them makes the scene look very weird regardless of it's the sky both up and down, or grass and sky. I think the ideal solution involves using some kind of backdrop city mesh, which I will look into later.

cgokmen: Resolving for now, opening a ticket for this.

omnigibson/utils/usd_utils.py
cgokmen: These form the triangles that create a kind of ring mesh that has the previous square on the inside and the new square on the outside. It's a hardcoded list, no easy way to generate it otherwise. Looks like this. alt text

cgokmen: image

omnigibson/utils/usd_utils.py
cgokmen: Note that the `if` side of the block here generates the white shape in this image, and the else part generates the blue one (there may be multiple layers of the blue)

image

PR #1150, opened 2025-03-11
add automatic pre-processing of wildcard bddl instances

cremebrule:


ChengshuLi (approved):
Does this break backward compatibility?

PR #1152, opened 2025-03-13
add checkpointing mechanism during demo collection

cremebrule:


omnigibson/simulator.py
```diff
@@ -1568,6 +1568,17 @@ def restore(self, scene_files):
...
+                    # Synchronize systems -- we need to check for pruning currently-existing systems,
+                    # as well as creating any non-existing systems
+                    current_systems = set(scene.active_systems.keys())
+                    load_systems = set(scene_info["state"]["system_registry"].keys())
+                    systems_to_remove = current_systems - load_systems
+                    systems_to_add = load_systems - current_systems
+                    for name in systems_to_remove:
+                        scene.clear_system(name)
+                    for name in systems_to_add:
+                        scene.get_system(name, force_init=True)
```
hang-yin, line 1580: Wooooo, is this a bug fix?

tests/test_data_collection.py
```diff
@@ -76,6 +76,15 @@ def test_data_collect_and_playback():
...
             env.step(env.robots[0].action_space.sample())
 
+        # Checkpoint state here for our first episode
+        if i == 0:
+            env.update_checkpoint()
+            robot_eef_state = {arm: env.robots[0].get_eef_position(arm=arm) for arm in env.robots[0].arm_names}
+
+            # Take one step to avoid creating the system immediately after the checkpoint is updated, which
+            # will cause downstream errors during playback
+            env.step(env.robots[0].action_space.sample())
```
hang-yin, line 86: Hmmmm I'm curious why this is needed here - seems like a related issue to something I'm fixing on 4.5

omnigibson/envs/data_wrapper.py
```diff
@@ -489,6 +530,17 @@ def add_transition_info(self, obj, add=True):
...
+        # If we're at the current checkpoint idx, this means that we JUST created a checkpoint and we're still at
+        # the same sim step.
+        # This is dangerous because it means that a transition is happening that will NOT be tracked properly
+        # if we rollback the state -- i.e.: the state will be rolled back to just BEFORE this transition was executed,
+        # and will therefore not be tracked properly in subsequent states during playback. So we assert that the current
+        # idx is NOT the current checkpoint idx
+        if self.checkpoint_step_idx is not None:
+            assert (
+                self.checkpoint_step_idx - 1 != self.env.episode_steps
+            ), "A checkpoint was just updated. Any subsequent transitions at this immediate timestep will not be replayed properly!"
```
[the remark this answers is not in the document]
cremebrule: yes and no ... basically if this situation occurs, without the assert the behavior is a bit undefined.

What this means is that there's transitions happening at the same timestep we recorded a checkpoint, but occurring _after_ the checkpoint was saved.

Usually this should never happen during normal runtime, but there's some edge cases, like the test_data_collection case where we manually add / remove water particles in between time steps.

Concretely, this causes an issue. Consider the following:
- We add water particles to the sim
- We step a few times
- We update the checkpoint (at t = T)
- Immediately after updating the checkpoint, we remove the water system (still at t = T). Without the assert, this means that our data collection wrapper is recording the water deletion event at t = T.
- We step a few times
- We try to rollback to the checkpoint at t = T (where the water was existing)
- We step a few times (with the water still existing)

When we try to playback the data, we now have incongruous information -- when we get to t = T, we see that there's a water deletion event, so we remove the water system. But for t > T, the recorded sim states still have water data (since the rolled back state included water particles), even though now in the actual sim the water was just deleted! So this is how OG can crash if there's add / deletion events occurring immediately after caching a checkpoint state.

omnigibson/simulator.py
cremebrule: Yep!! We weren't even handling systems before 😅

tests/test_data_collection.py
cremebrule: Oh well at least for me this is a specific safeguard against the behavior described above...I don't think this should have implications outside of the data collection wrapper checkpointing code.

hang-yin (approved):

cgokmen (commented):
LGTM, sorry for delay

PR #1174, opened 2025-04-09
Fix: friction issue for non-controllable objects

hang-yin:
1. Tune damping and friction for non-controllable objects
2. Decouple position/velocity instantaneous setter and target setter

### DIFF
--- a/omnigibson/objects/controllable_object.py
+++ b/omnigibson/objects/controllable_object.py
@@ -461,6 +461,14 @@ class ControllableObject(BaseObject):
             idx += controller.command_dim
 
     @property
+    def _is_controllable(self) -> bool:
+        """
+        Returns:
+            bool: Whether this object is controllable or not
+        """
+        return True
+
+    @property
     def control_enabled(self):
         return self._control_enabled
 
--- a/omnigibson/objects/controllable_object.py
+++ b/omnigibson/objects/controllable_object.py
@@ -461,10 +461,10 @@ class ControllableObject(BaseObject):
             idx += controller.command_dim
 
     @property
-    def _is_controllable(self) -> bool:
+    def is_driven(self) -> bool:
         """
         Returns:
-            bool: Whether this object is controllable or not
+            bool: Whether this object is actively controlled/driven or not
         """
         return True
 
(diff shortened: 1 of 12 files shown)


omnigibson/prims/geom_prim.py
```diff
@@ -460,9 +460,9 @@ def _post_load(self):
         # run super first
         super()._post_load()
 
-        # TODO: tmp fix for visible fillable volume
-        if "fillable" not in self.name:
-            # Make sure mesh is visible by default
+        # TODO: tmp fix for visible metalinks
```
cremebrule, line 463: ooof this is super ugly but i guess it works 😂

omnigibson/prims/joint_prim.py
```diff
@@ -797,9 +797,14 @@ def set_pos(self, pos, normalized=False, drive=False):
             pos = self._denormalize_pos(pos)
 
         # Set the DOF(s) in this joint
-        if drive:
+        if self.driven and drive:
```
cremebrule, line 800: Can we add comments here for the semantic differences between these, and the motivations? (same applies to set_vel)

To make this clearer, maybe we should split the if/ else into nested, like so:

```python
if self.driven:
    # Basically any controllable object
    if drive:
        ...
    else:
        ...
else:
    # Any other object, e.g.: furniture
    ...
```

omnigibson/prims/entity_prim.py
```diff
@@ -365,6 +366,14 @@ def _update_joint_limits(self):
             joint.lower_limit = joint.lower_limit * scale_along_axis
             joint.upper_limit = joint.upper_limit * scale_along_axis
 
+    @property
+    def _is_controllable(self) -> bool:
```
cremebrule, line 370: should we rename this to be more clear -- e.g.: "is_driven"?

omnigibson/objects/controllable_object.py
```diff
@@ -460,6 +460,14 @@ def apply_action(self, action):
             # Update idx
             idx += controller.command_dim
 
+    @property
+    def _is_controllable(self) -> bool:
```
cremebrule, line 464: maybe rename, see other comment

cremebrule (approved):
lgtm, just some small things, feel free to merge after!

omnigibson/prims/geom_prim.py
hang-yin: Temporary fix on this data collection branch before the final dataset release😂

PR #1182, opened 2025-04-19
Update inside

wensi-ai:


cgokmen (changes requested):

omnigibson/prims/rigid_prim.py
```diff
@@ -544,3 +545,51 @@ def meta_link_sub_id(self):
         # Check using the old format.
         # TODO: Remove this after the next dataset release
         return int(m.LEGACY_META_LINK_PATTERN.fullmatch(self.name).group(3))
+
+    def check_points_in_volume(
+        self,
+        particle_positions_world,
+        use_visual_meshes=True,
+        mesh_name_prefixes=None,
```
cremebrule, line 553: probably should be renamed to `mesh_name_prefix` for better clarity

omnigibson/prims/rigid_prim.py
```diff
@@ -544,3 +545,51 @@ def meta_link_sub_id(self):
...
+    ):
+        """
+        Args:
+            particle_positions_world (th.tensor): (N, 3) array of particle positions to check
+            use_visual_meshes (bool): Whether to use @volume_link's visual or collision meshes to generate points fcn
```
cremebrule, line 558: we should also clarify here that we assume that these meshes are convex (and if not, we enforce it anyways by converting them into convex)

omnigibson/prims/geom_prim.py
```diff
@@ -160,6 +167,64 @@ def geom_type(self):
...
+    def mesh_face_normals(self):
+        trimesh_mesh = mesh_prim_mesh_to_trimesh_mesh(
+            self.prim, include_normals=False, include_texcoord=False
+        ).convex_hull
+        assert trimesh_mesh.is_convex, f"Trying to get mesh face normals for a non-convex mesh {self.prim_path}"
+        return th.tensor(trimesh_mesh.face_normals, dtype=th.float32)
+
+    def check_local_points_in_volume(self, particle_positions_in_mesh_frame):
+        mesh_type = self.prim.GetTypeName()
+        if mesh_type == "Mesh":
```
cremebrule, line 182: given that the mesh type is static, maybe we can define the proper one during init / load time so we don't have to do a bunch of if / else at runtime?

omnigibson/prims/geom_prim.py
```diff
@@ -160,6 +167,58 @@ def geom_type(self):
...
+        vertices = self.points
+        face_indices = self.faces
+        return vertices[face_indices].mean(dim=1)
+
+    @cached_property
+    def mesh_face_normals(self):
+        return vtarray_to_torch(self.prim.GetAttribute("normals").Get())
+
+    def check_local_points_in_volume(self, particle_positions_in_mesh_frame):
+        mesh_type = self.prim.GetTypeName()
```
cremebrule, line 181: cache mesh type to avoid slow reads during runtime

omnigibson/object_states/inside.py
```diff
@@ -1,8 +1,9 @@
 import torch as th
 
 import omnigibson as og
+from omnigibson.macros import macros
 from omnigibson.object_states.aabb import AABB
-from omnigibson.object_states.adjacency import HorizontalAdjacency, VerticalAdjacency, flatten_planes
+from omnigibson.object_states.adjacency import HorizontalAdjacency, VerticalAdjacency
```
cgokmen, line 6: Why is this not complaining about unnecessary imports, can you check if they are used somewhere and remove them?

cgokmen (approved):

omnigibson/object_states/inside.py
cremebrule: oooh great point, forgot to remove them

PR #1764, opened 2025-06-24
One stop install

hang-yin:


OmniGibson/setup.py
```diff
@@ -34,7 +34,6 @@
         "trimesh>=4.0.8",
         "h5py>=3.10.0",
         "cryptography>=41.0.7",
-        "bddl~=3.5.0",
```
cgokmen, line 37: Lets not remove this. Just pin the current version so that when BDDL is installed first it won't try to pull it again. Let's also set all the versions to 3.7.0-alpha (including BDDL, JoyLo, and OmniGibson), and 3.7.0rc1 for ig_pipeline including the entry missed above and the one in __init__.py - at release time we'll remove the alpha part.

README.md
```diff
@@ -1,3 +1,50 @@
...
+- **GPU**: NVIDIA RTX 2080+
+
+## Quick Start
+
+For most users, we recommend the full installation with a new conda environment:
+
+```bash
+git clone https://github.com/StanfordVL/BEHAVIOR-1K
+cd BEHAVIOR-1K
+chmod +x setup.sh
```
cgokmen: Do the chmod +x ahead of time so the user doesn't have to.

joylo/setup.py
```diff
@@ -4,14 +4,13 @@
...
-    author_email="philippwu@berkeley.edu",
-    description="software for GELLO",
+    name="joylo",
+    version="3.6.0",
+    author="Stanford University",
+    description="JoyLo for OmniGibson",
     long_description=long_description,
     long_description_content_type="text/markdown",
-    url="https://github.com/wuphilipp/gello_software",
+    url="https://github.com/StanfordVL/og-gello",
```
cgokmen: Fix

PR #1782, opened 2025-07-10
Misc. Fixes

hang-yin:
- [ ] Make installation script more robust
- [ ] Replay minor bug fix
- [ ] Sampling bug fix - quit when 1) not success 2) not applying dynamic scaling

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 
 # BEHAVIOR-1K Installation Script
-# Usage: ./setup.sh [OPTIONS]
+# Usage: . ./setup.sh [OPTIONS]
+
+# Enable strict error handling - exit on any command failure
+set -e
 
 # Read Arguments
 TEMP=`getopt -o h --long help,new-env,omnigibson,bddl,teleop,dataset,primitives,dev,cuda-version: -n 'setup.sh' -- "$@"`
@@ -80,8 +83,8 @@ if [ "$HELP" = true ] ; then
 fi
 
 # Check if we're in the right directory
-if [ ! -d "omnigibson" ] && [ ! -d "bddl" ] && [ ! -d "joylo" ] ; then
-    echo "[ERROR] Cannot find omnigibson, bddl, or joylo directories"
+if [ ! -d "OmniGibson" ] && [ ! -d "bddl" ] && [ ! -d "joylo" ] ; then
+    echo "[ERROR] Cannot find OmniGibson, bddl, or joylo directories"
     echo "[ERROR] Please run this script from the BEHAVIOR-1K root directory"
     return 1
 fi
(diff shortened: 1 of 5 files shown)


setup.sh
```diff
@@ -1,283 +1,239 @@
...
-        --bddl) BDDL=true ; shift ;;
-        --teleop) TELEOP=true ; shift ;;
-        --dataset) DATASET=true ; shift ;;
-        --primitives) PRIMITIVES=true ; shift ;;
-        --asset-pipeline) ASSET_PIPELINE=true ; shift ;;
-        --dev) DEV=true ; shift ;;
-        --cuda-version) CUDA_VERSION="$2" ; shift 2 ;;
-        --) shift ; break ;;
-        *) ERROR=true ; break ;;
+CUDA_VERSION="12.4"
```
cgokmen, line 13: Why?

OmniGibson/omnigibson/utils/object_state_utils.py
```diff
@@ -99,7 +99,7 @@ def sample_kinematics(
     max_trials=None,
     z_offset=0.05,
     skip_falling=False,
-    use_last_ditch_effort=False,
+    use_last_ditch_effort=True,
```
cgokmen, line 102: You sure about this? It makes the "last ditch effort", which presumably has some downside, part of the default behavior...

setup.sh
```diff
@@ -1,283 +1,381 @@
...
-# Initialize flags
+# Parse arguments
 HELP=false
 NEW_ENV=false
 OMNIGIBSON=false
 BDDL=false
 TELEOP=false
 DATASET=false
 PRIMITIVES=false
-ASSET_PIPELINE=false
```
cgokmen, line 19: Bad merge removes my pipeline setup features. Please undo and reflect in the ps1 script also.

setup.sh
```diff
@@ -1,283 +1,381 @@
...
+            download_og_dataset(accept_license=${DATASET_ACCEPT_FLAG})
+        
+        if not assets_exist:
+            print('Downloading assets...')
+            download_assets()
+    else:
+        print('Datasets already exist, skipping download.')
+except Exception as e:
+    print(f'ERROR: Dataset installation failed: {e}')
+    exit(1)
```
cgokmen, line 359: Lets either convert this to bash also, or keep this as a python script in the omnigibson scripts directory and just call it. 

Converting to bash makes more sense imo, you can move the if exists logic into the download functions themselves to get rid of those parts here (no reason ever to download if it already exists). Here you'd just have two separate python calls for `python -c "import ...; download_og_dataset()"` and `python -c "import ...; download_assets()"` in place of lines 350 and 354.

setup.sh
hang-yin: @wensi-ai why was this again?🤣

PR #1792, opened 2025-08-06
Small Misc. Fixes

cremebrule:
Random fixes observed:

- CoACD: Set `decimate=True` to enforce that our `max_ch_vertices` arg is respected. See official [CoACD README docs](https://github.com/SarahWeiii/CoACD?tab=readme-ov-file#parameters) about their args
- `franka.py` / `franka_mounted.py`: Update typo in ag_start_/end_point property causing raycast errors, update franka mounted to use default (auto-generated) ag points instead of hardcoded franka ones since the gripper morphology is different
- `manipulation_robot.py`: Make gripper raycast checking numerically stable by adding small epsilon to raycast end points to avoid start / end point overlap

### DIFF
--- a/OmniGibson/omnigibson/robots/franka.py
+++ b/OmniGibson/omnigibson/robots/franka.py
@@ -314,7 +314,7 @@ class FrankaPanda(ManipulationRobot):
 
     @property
     def _assisted_grasp_end_points(self):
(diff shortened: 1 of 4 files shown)


hang-yin (approved):
LGTM!

PR #1796, opened 2025-08-09
Presample robot poses

hang-yin:
Implements https://github.com/StanfordVL/BEHAVIOR-1K/issues/846

Note: we only presample base poses for mobile manipulators

TODO:
- [x] patch all task instances
- [x] fix tro reset (directly read presampled robot pose instead of parsing robot state)


OmniGibson/omnigibson/scenes/scene_base.py
```diff
@@ -403,6 +411,9 @@ def _load_metadata_from_scene_file(self):
 
         # Write the metadata
         for key, data in scene_info.get("metadata", dict()).items():
+            if key == "task":
+                # Robot poses data type can't get stored in the scene prim; we will store this with the task
+                data.pop("robot_poses")
```
cgokmen: Sorry for the uninformed comment, but - why does this need to be here? Wasn't there already task metadata (e.g. the object scope - object assignments)? What changed that requires the addition of this?

OmniGibson/omnigibson/scenes/scene_base.py
```diff
@@ -136,6 +137,9 @@ def __init__(
                 obj = create_object_from_init_info(obj_info)
                 self._init_objs[obj_name] = obj
 
+            # Store presampled robot poses from metadata
+            self._presampled_robot_poses = task_metadata.get("robot_poses", None)
+
```
cgokmen: I would advocate not having this here at all and having the task instance maintain this list. It's behavior task specific anyway.

OmniGibson/omnigibson/tasks/behavior_task.py
```diff
@@ -242,9 +244,19 @@ def _load(self, env):
     def reset(self, env):
         super().reset(env)
 
+        # Use presampled robot pose if specified (only available for officially supported mobile manipulators)
+        if self.use_presampled_robot_pose:
+            robot = self.get_agent(env)
+            presampled_poses = env.scene.presampled_robot_poses
```
cgokmen: Yeah this should not be on the scene - clear reasoning is that it's only used in this class which has access to the metadata itself.

OmniGibson/omnigibson/tasks/behavior_task.py
```diff
@@ -123,6 +124,7 @@ def __init__(
 
         # Object info
         self.online_object_sampling = online_object_sampling  # bool
+        self.use_presampled_robot_pose = use_presampled_robot_pose
```
cgokmen, line 132: There should also be some kind of setup that allows the user to decide if it should always go to pose 0 or to randomly one of the stored poses.

joylo/gello/robots/sim_robot/og_sim.py
```diff
@@ -875,12 +875,13 @@ def reset(self, increment_instance=True):
                 tro_state = recursively_convert_to_torch(json.load(f))
             self.env.scene.reset()
             for bddl_name, obj_state in tro_state.items():
-                if "agent" in bddl_name:
+                if bddl_name == "robot_poses":
```
cgokmen, line 878: This seems like a weird setup. Why is this data mixed in with tro state info? Can't we store it separately?

cgokmen (changes requested):
FYI I can't review the diff on the sampled tasks, github won't show it. Would be great if you can screenshot the diff on one of them or something.

OmniGibson/omnigibson/scenes/scene_base.py
hang-yin: Yeah I'm not sure, this is why we need @cremebrule to help review this. It seems like currently this metadata is stored on the scene prim as metadata. The original dictionary with strings seem to store there fine but not these floats.

joylo/gello/robots/sim_robot/og_sim.py
hang-yin: Where should we store it though? This tro setup is never meant as a permanent solution (it's only implemented to save space).

hang-yin: Storing it in the tro file clearly assigns these poses to this instance. Alternatively, we want another file just for all the robot poses?

hang-yin:
image (1)
image (2)

@cgokmen fyi; also deleted all robot states (in init info and states)

OmniGibson/omnigibson/envs/env_base.py
```diff
@@ -617,9 +617,16 @@ def step(self, action, n_render_iterations=1):
                 - dict: info, i.e. dictionary with any useful information
         """
         # Pre-processing before stepping simulation
-        if isinstance(action, Iterable) and not isinstance(action, (dict, OrderedDict)):
+        if isinstance(action, (dict, OrderedDict)):
```
cremebrule: nit but should probably remove the OrderedDict check since all dicts are ordered implicitly since python 3.6 IIRC

OmniGibson/omnigibson/tasks/behavior_task.py
```diff
@@ -81,6 +85,8 @@ def __init__(
         activity_instance_id=0,
         predefined_problem=None,
         online_object_sampling=False,
+        use_presampled_robot_pose=False,
```
cgokmen, line 88: This should default to True imo.

joylo/gello/robots/sim_robot/og_sim.py
cgokmen: Man I really have no idea what this file does or what's happening here or what the read file is. Feel free to merge if you think this is right. This part of the code generally seems a little lower quality.

cgokmen (approved):

joylo/gello/robots/sim_robot/og_sim.py
hang-yin: Yeah I agree, a lot of the code in here was added ad-hoc to get data collection going. Will need to clean it up at some point.

PR #1814, opened 2025-08-30
Feat/isaac 5.0

cremebrule:


OmniGibson/setup.py
```diff
@@ -71,7 +71,7 @@
         ],
         "primitives": [
             "nvidia-curobo @ git+https://github.com/StanfordVL/curobo@cbaf7d32436160956dad190a9465360fad6aba73",
-            "ompl @ https://storage.googleapis.com/gibson_scenes/ompl-1.6.0-cp310-cp310-manylinux_2_28_x86_64.whl",
+            # "ompl @ https://storage.googleapis.com/gibson_scenes/ompl-1.6.0-cp310-cp310-manylinux_2_28_x86_64.whl", # Todo? update to python3.11
```
cgokmen: It's safe to remove this now I think. It had quite a few advantages compared to curobo but it's mostly disconnected now.

OmniGibson/omnigibson/objects/object_base.py
```diff
@@ -108,6 +109,9 @@ def __init__(
         load_config["self_collisions"] = self_collisions
         load_config["prim_type"] = prim_type
 
+        # Store this category
+        OBJECT_CATEGORIES.add(self.category)
```
cgokmen, line 113: Not a good idea - this makes it not be a constant. We should just scan all available datasets for categories in the semantic module instead of doing this at load time this way.

OmniGibson/omnigibson/prims/geom_prim.py
```diff
@@ -59,6 +65,43 @@ def _post_load(self):
...
+            assert self.scene is not None, "Cannot get position and orientation relative to scene without a scene"
+            return self.scene.convert_world_pose_to_scene_relative(*PoseAPI.get_world_pose(self.prim_path))
+        elif frame == "link":
+            assert self._link is not None, "Cannot get position and orientation relative to link without a link"
+            return T.mat2pose(
+                T.pose2mat(self._link.get_position_orientation()) @ T.pose2mat(PoseAPI.get_world_pose(self.prim_path))
+            )
+        else:  # parent
+            position, orientation = lazy.isaacsim.core.utils.xforms.get_local_pose(self.prim_path)
+            return th.as_tensor(position, dtype=th.float32), th.as_tensor(orientation[[1, 2, 3, 0]], dtype=th.float32)
```
cgokmen, line 103: Can you change the implementation of this to defer to super() for everything but the parent? In fact even for parent we can call super() with "world" and then postprocess. So we shouldn't do any actual pose getting here. Pose-getting is finicky so it's best to keep it limited to xformprim and rigidprim.

OmniGibson/omnigibson/prims/prim_base.py
```diff
@@ -137,6 +137,10 @@ def _post_load(self):
         object and grab its handles and internal references. By default, this is a no-op.
         """
         pass
+        # # TODO: Remove this once we have support for instanceable prims
+        # # Disable instanceable prims because it breaks many fundamental R/W paradigms in OG
+        # if self._prim.IsInstanceable() or self._prim.IsInstanceProxy():
+        #     from IPython import embed; embed()
```
cgokmen, line 143: Remove before merging

OmniGibson/omnigibson/scenes/scene_base.py
```diff
@@ -1158,6 +1158,10 @@ def _load_state(self, state):
         # TODO: Remove backwards compatible check once new scene RC is updated
         if "pos" in state:
             self.set_position_orientation(position=state["pos"], orientation=state["ori"])
+            # We need to propagate these changes or else we get a crash
+            og.sim.pi.update_simulation(elapsedStep=0.0, currentTime=og.sim.current_time)
+            og.sim.psi.fetch_results()
```
cgokmen, line 1182: Can we put this inside simulator.py as `fake_step` or something and call it (here and elsewhere where update_simulation is currently called)? It's actually probably useful in a lot of places where we really call step, too!

OmniGibson/omnigibson/utils/asset_conversion_utils.py
```diff
@@ -750,6 +751,19 @@ def import_obj_metadata(usd_path, obj_category, obj_model, dataset_root, force_a
...
+
+        def remove_instanceable_recursive(_root):
+            # See https://openusd.org/docs/api/_usd__page__scenegraph_instancing.html ("Traversing Into Instances with Instance Proxies")
+            children = _root.GetFilteredChildren(lazy.pxr.Usd.TraverseInstanceProxies())
+            for child in children:
+                if child.IsInstanceable():
+                    child.SetInstanceable(False)
+                remove_instanceable_recursive(_root=child)
+
+        remove_instanceable_recursive(_root=prim)
```
cgokmen, line 765: I think that something else I do in this file (where I flatten the imported object hierarchy) inevitably does this. But regardless it should be fine to keep.

OmniGibson/omnigibson/utils/asset_conversion_utils.py
```diff
@@ -2468,9 +2491,19 @@ def import_og_asset_from_urdf(
         use_omni_convex_decomp=False,  # We already pre-decomposed the values, so don' use omni convex decomp
         use_usda=use_usda,
         merge_fixed_joints=merge_fixed_joints,
+        import_inertia_tensor=import_inertia_tensor,
     )
 
-    prim = import_obj_metadata(usd_path=usd_path, obj_category=category, obj_model=model, dataset_root=dataset_root)
+    # Copy meta links URDF to original name of object model
+    shutil.copy2(urdf_path, os.path.join(dataset_root, "objects", category, model, "urdf", f"{model}.urdf"))
```
cgokmen: This overwrites the file if the file is already here, which may not be desirable. Why are we not OK with the current format?

OmniGibson/omnigibson/simulator.py
```diff
@@ -70,7 +71,8 @@
 m.INITIAL_SCENE_PRIM_Z_OFFSET = -100.0
 
 m.KIT_FILES = {
-    (4, 5, 0): "omnigibson_4_5_0.kit",
+    # (4, 5, 0): "omnigibson_4_5_0.kit",
```
cgokmen, line 75: OK to delete this and also the file. In fact the (new) file should be called omnigibson.kit.

setup.sh
```diff
@@ -12,7 +12,7 @@ PRIMITIVES=false
 EVAL=false
 ASSET_PIPELINE=false
 DEV=false
-CUDA_VERSION="12.4"
+CUDA_VERSION="12.8"
```
cgokmen, line 16: This is a really high CUDA version. I don't have it on the cluster I'm working on for example. Do we have to do this? What's requiring it?

setup.sh
```diff
@@ -303,21 +304,38 @@ if [ "$OMNIGIBSON" = true ]; then
...
+                "isaacsim_extscache_physics-5.0.0.0"
+                "isaacsim_extscache_kit-5.0.0.0"
+                "isaacsim_extscache_kit_sdk-5.0.0.0"
             )
             
             local wheel_files=()
             for pkg in "${packages[@]}"; do
                 local pkg_name=${pkg%-*}
-                local filename="${pkg}-cp310-none-manylinux_2_34_x86_64.whl"
+                local filename="${pkg}-cp311-none-manylinux_2_35_x86_64.whl"
```
cgokmen: Even higher glibc requirement :(

OmniGibson/omnigibson/install.py
cgokmen, line 1: Can you re-remove this file?

OmniGibson/omnigibson/prims/xform_prim.py
```diff
@@ -63,9 +63,14 @@ def _post_load(self):
         super()._post_load()
 
         # Make sure all xforms have pose and scaling info
-        # These only need to be done if we are creating this prim from scratch.
+        # These only need to be done if we are creating this prim from scratch AND it is not an instanceable / proxy prim.
```
cgokmen, line 66: Why not? The pointer side should still have this no?

OmniGibson/omnigibson/prims/rigid_prim.py
```diff
@@ -165,48 +166,97 @@ def update_meshes(self):
...
+                    relative_prim_path=absolute_prim_path_to_scene_relative(self.scene, prim.GetPrimPath().__str__()),
+                    name=f"{prim.GetName()}_non_instanced",
+                    load_config=None,
+                )
+                xform.load(self.scene)
+            return True
+
+        # We don't do this recurisvely to save on compute, since we know the immediate children are instances if this
+        # does happen
+        # self.modify_children_prims(fcn=set_non_instanced, recursive=False, include_instances=True)
```
cgokmen, line 186: What is this, why is it commented out?

OmniGibson/omnigibson/prims/rigid_prim.py
```diff
@@ -165,48 +166,97 @@ def update_meshes(self):
...
-                    self._visual_meshes[mesh_name] = VisualGeomPrim(**mesh_kwargs)
-                    self._visual_meshes[mesh_name].load(self.scene)
+                    nested_tf = current_local_tf
+                    nested_scale = current_scale
+                if prim_type in GEOM_TYPES:
+                    assert len(children_dict) == 0, "No children prims should be owned by a Mesh prim!"
+                    # Raw meshes themselves could be nested, and share the same name so we convert their (unique) prim path
+                    # into a string we'll use for their name
+                    mesh_name = prim_path.split(self.prim_path)[-1].strip("/").replace("/", "_")
+                    is_collision = prim.HasAPI(lazy.pxr.UsdPhysics.CollisionAPI)
```
cgokmen, line 220: **Important** I don't think this works with new-import objects. At least in 4.5 the collision API is on an xformprim between the actual geoms and the link (e.g. link > collisions > collision1_xform > collision1_geom). So it's no longer possible to have a "collision geom" - the geoms are identical collision or visual.

That said, if this doesnt break the existing objects lets merge it as is and worry about it later.

OmniGibson/omnigibson/prims/rigid_prim.py
```diff
@@ -165,48 +166,97 @@ def update_meshes(self):
...
-                    local_pos, local_orn = mesh.get_position_orientation(frame="parent")
-                    vols.append(volume * th.prod(mesh.scale))
-                    coms.append(T.quat2mat(local_orn) @ (com * mesh.scale) + local_pos)
-                    # If the ratio between the max extent and min radius is too large (i.e. shape too oblong), use
-                    # boundingCube approximation for the underlying collision approximation for GPU compatibility
-                    if not check_extent_radius_ratio(mesh, com):
-                        log.warning(f"Got overly oblong collision mesh: {mesh.name}; use boundingCube approximation")
-                        mesh.set_collision_approximation("boundingCube")
+
+        # Need to explicitly check for instanced children here since they may include instanced meshes
```
cgokmen, line 170: I'm getting a feeling that this PR should just assume/assert that nothing is instanceable. There's no purpose in trying to change those, things will break. The importer already disables this instancing. Let's save it for a separate PR.

fan-ziqi:
When will this be merged?

wensi-ai:
Migrated to #1955

PR #1827, opened 2025-09-14
Fixing setup.ps1

husain-zaidi:
Adding missing }


wensi-ai:
Hi, thanks for pointing this out! We already have it fixed in the main branch, please pull the latest changes and the setup should work now.

PR #1831, opened 2025-09-18
[feat]: adds support for uv package manager

wpumacay:
- Adds support to setup.sh script for the UV package manager, when using a custom environment. Just changes call to `pip` with `uv pip` if the `--use-uv` flag is given
- Tested it on Arch Linux as platform, and `uv 0.8.17`
- There's a weird issue when installing `bddl` in development mode with `-e`, so had to install it in normal mode like this: `uv pip install $WORKDIR/bddl`


wensi-ai:
Closing this for now, feel free to reopen if needed.

PR #1836, opened 2025-09-20
Add storage requirements to docs

Cascoopman:
I ran out of disk space running the installation steps on a fresh VM with 100GB. Got out of disk error. After installation the final size required is about 70GB, but during the process it requires more.


wensi-ai:
Hi @Cascoopman , we have updated our installation process of the dataset so now it should be good. Feel free to reopen if you still spot the same issue!

PR #1841, opened 2025-09-29
Fixed right gripper range

IliaLarchenko:
It is a tiny fix.

For some reason, the left gripper has a range from -1 to +1:
https://github.com/StanfordVL/BEHAVIOR-1K/blob/12b0aeeeb755e39aee820a365da01673ec34f739/OmniGibson/omnigibson/learning/utils/eval_utils.py#L164

But the right gripper is from 0 to 0.05:
https://github.com/StanfordVL/BEHAVIOR-1K/blob/12b0aeeeb755e39aee820a365da01673ec34f739/OmniGibson/omnigibson/learning/utils/eval_utils.py#L169

I think the correct one is -1 to 1 for both, based on the range of actions in the training data.

The PR is just this simple fix that makes it -1 to 1 for the right gripper.

### DIFF
--- a/OmniGibson/omnigibson/learning/utils/eval_utils.py
+++ b/OmniGibson/omnigibson/learning/utils/eval_utils.py
@@ -166,7 +166,7 @@ JOINT_RANGE = {
             th.tensor([-4.4506, -3.1416, -2.3562, -2.0944, -2.3562, -1.0472, -1.5708], dtype=th.float32),
             th.tensor([1.3090, 0.1745, 2.3562, 0.3491, 2.3562, 1.0472, 1.5708], dtype=th.float32),
         ),
-        "right_gripper": (th.tensor([0.00], dtype=th.float32), th.tensor([0.05], dtype=th.float32)),
+        "right_gripper": (th.tensor([-1], dtype=th.float32), th.tensor([1], dtype=th.float32)),
     },
 }
 


wensi-ai:
Hi @IliaLarchenko, thanks for pointing this out! After inspection I believe this should be [0, 0.05] ([-1, 1] is actually the action range, not the joint angle range). I have pushed the fix to main, feel free to check out!

IliaLarchenko:
Thanks!

Can you also explain the meaning of these ranges for actions and states?

Is [-1, 1] for actions a scaled-up [0, 0.05]. Or are actions more like velocities in this case, and -1/1 means to close/open the gripper more?

The actual states of grippers are [0, 0.1] in the training data. Is it because [0, 0.05] is for one finger, and the state represents the distance between fingers in the symmetric gripper?

Having different ranges everywhere is a bit confusing.

wensi-ai:
Hi,

[-1, 1] is the normalized action specified by MultiFingerJointController(mode=smooth), where -1 means gripper close and 1 means gripper open (aboslute pose). For more information, please refer to our [documentaiton](https://behavior.stanford.edu/reference/controllers/multi_finger_gripper_controller.html)

For the actual state, if you are referring to openpi repo, we use joint gripper distances as specified [here](https://github.com/wensi-ai/openpi/blob/behavior/src/openpi/policies/b1k_policy.py#L31-L32)

IliaLarchenko:
Thank you!

PR #1842, opened 2025-09-30
No credit for just achieving initial state

hang-yin:


bddl/bddl/config.py
```diff
@@ -24,5 +24,5 @@ def get_domain_filename(domain_name):
 
 
 # MISC
-GROUND_GOALS_MAX_OPTIONS = 20
-GROUND_GOALS_MAX_PERMUTATIONS = 10
+GROUND_GOALS_MAX_OPTIONS = None  # No limit - generates all combinations
```
cgokmen: Lets just remove this feature. Let the user worry about it. It's as simple as calling itertools.islice on the generator. Remove the truncated product functions too.

OmniGibson/omnigibson/metrics/task_metric.py
```diff
@@ -19,16 +19,44 @@ def start_callback(self, env):
...
+                    initial_state = initial_states[j]
+                    # Only count if it went from False to True
+                    if current_state and not initial_state:
+                        newly_satisfied_count += 1
+
+                # Calculate score as proportion of initially unsatisfied predicates that are now satisfied
+                option_score = newly_satisfied_count / predicates_to_complete
+            else:
+                # All predicates were already satisfied initially, no progress to measure
+                option_score = 0.0
```
cgokmen: This is unlikely/impossible. No point in having this performance hack. We can remove the predicates_to_complete variable/condition.

cgokmen (changes requested):

OmniGibson/omnigibson/metrics/task_metric.py
```diff
@@ -19,16 +19,41 @@ def start_callback(self, env):
...
+                if not initial_state:
+                    initially_unsatisfied_count += 1
+                    # Check if it's now satisfied
+                    if predicate.evaluate():
+                        newly_satisfied_count += 1
+
+            # Calculate score as proportion of initially unsatisfied predicates that are now satisfied
+            option_score = (
+                newly_satisfied_count / initially_unsatisfied_count if initially_unsatisfied_count > 0 else 0.0
+            )
```
cgokmen: I don't think this is what we want to be doing. I think the partial credit divides by the total number of predicates regardless of how many are initially satisfied. Otherwise you still end up in the scenario where people can get full credit despite not finishing the task.

The one caveat is that we do ofc want to give full credit if the full task is finished - you should check the task for completion prior to the outer loop and just return 1.0 instead of doing all this math for no reason. (I dont think that check requires enumerating the conditions, either)

bddl/bddl/condition_evaluation.py
```diff
@@ -42,7 +38,7 @@ def evaluate(self):
         return all(self.child_values)
 
     def get_ground_options(self):
-        options = list(truncated_product(*[child.flattened_condition_options for child in self.children]))
+        options = list(itertools.product(*[child.flattened_condition_options for child in self.children]))
```
cgokmen, line 41: It's quite annoying that we do this thing in a whole bunch of places in this tree where we convert the iterators to lists which prevents the user from being able to stop generating these early. It's a bit complex to fix so please open a ticket for it and lets worry about it later.

hang-yin: Done: https://github.com/StanfordVL/BEHAVIOR-1K/issues/1845

cgokmen (approved):

PR #1843, opened 2025-09-30
Switch back from USDZ to USD due to load time and caching issues

cgokmen:


cgokmen:
It's actually two different "constants" - one for the URL of the current version that should be downloaded, another for the minimum version that's supported by the current codebase. I don't know if those two should always be the same or not. It's fine for now I think.

PR #1848, opened 2025-10-02
[WIP] Refactor Conjunction and Disjunction for lazy evaluation

### LINKED ISSUE #1845: BDDL: Refactor goal state generation to support lazy evaluation with iterators
The current implementation in bddl/bddl/condition_evaluation.py eagerly converts iterators to lists in multiple places during goal state generation, which prevents users from being able to stop generation early and causes unnecessary memory usage for large combinatorial spaces.

Jongwan93:
Fixes #1845 

## Hello this is a **Draft Pull Request** to get some early feedback on my approach on the issue #1845.
My goal is to refactor the eager evaluation patterns into lazy ones using generators.

### Changes
I have started by refactoring the `Conjunction` and `Disjunction` classes.
- **`evaluate()` method**: Removed `List` library in `Conjunction.evaluate()` and `Disjunction.evaluate()`. This allow `all()` and `any()` to utilize short-circuit evaluation, improving performance.
- **`Conjunction.get_ground_options()`**: Refactored to iterate directly over `itertools.product` instead of creating a large list in memory.

### Question
In the original `evaluate()` method, there was an `assert` statement to check for `None` values.
`assert all([val is not None for val in self.child_values])`
My current refactoring removes this explicit check for simplicity.
If this `assert` check is important, I can implement a helper function to accommodate this feature

What are your thoughts on this approach? Please let me know your preference.

### DIFF
--- a/bddl/bddl/condition_evaluation.py
+++ b/bddl/bddl/condition_evaluation.py
@@ -32,15 +32,22 @@ class Conjunction(Expression):
         if generate_ground_options:
             self.get_ground_options()
 
+    # def evaluate(self):
+    #     self.child_values = [child.evaluate() for child in self.children]
+    #     assert all([val is not None for val in self.child_values]), "child_values has NoneTypes"
+    #     return all(self.child_values)
+
     def evaluate(self):
-        self.child_values = [child.evaluate() for child in self.children]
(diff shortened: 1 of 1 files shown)


Jongwan93:
I will send new PR soon

PR #1858, opened 2025-10-11
Move appdata directory into OmniGibson and make it configurable

### LINKED ISSUE #1851: Improved cache handling on HPC
Over the years I was never able to figure out how to get the shader cache working on SLURM. Usually the problem is that the shader cache ends up at some unknown location that's on a shared drive, and is space-limited, so you dont get the right things in the cache for any of the GPUs. Today I realized the extension that manages these is located in a very interesting place:

`~/miniconda3/envs/behavior/lib/python3.10/site-packages/omni/data/Kit/OmniGibson/3.7/exts/3/omni.gpu_foundation.shadercache.vulkan-1.0.0+d02c707b.lx64.r/omni/gpu_foundation/shadercache/vulkan/shadercache_vulkan.py`

It caches the actual shaders at these places:
`['~/miniconda3/envs/behavior/lib/python3.10/site-packages/omni/data/Kit/OmniGibson/3.7/exts/3/omni.gpu_foundation.shadercache.vulkan-1.0.0+d02c707b.lx64.r/cache/shadercache', '~/miniconda3/envs/behavior/lib/python3.10/site-packages/omni/data/Kit/OmniGibson/3.7/exts/3/omni.hydra.rtx.shadercache.vulkan-1.0.0+d02c707b.lx64.r/cache/shadercache']`

These are symlinks to the extscache copy of those plugins.

We can do one of two things:
1. Replace those paths with symlinks to /tmp etc - or some other local path.
2. Configure the shadercache exts to not use those paths but instead let us specify the paths.

Either way we should do that and at launch time assert that the path is on a local disk.

cgokmen:
Previously, caches and logs were getting thrown into multiple different locations:

1. app logs into the conda env's site_packages/omni/logs
2. app caches (including shaders) into site_packages/omni/data/... and site_packages/omni/cache/...
3. texture caches into ~/.cache/ov/texturecache
4. other arbitrary data into ~/.nvidia-omniverse/data/...

This distribution causes issues when the conda env and/or the home directory are on networked drives like in most HPC clusters.

This PR standardizes all of this data to go to a centralized appdata location which defaults to OmniGibson/appdata in this repo but can be changed with a macro.

This fixes #1851

### DIFF
--- a/docs/challenge/dataset.md
+++ b/docs/challenge/dataset.md
@@ -1,6 +1,6 @@
 # Dataset
 
-### **NOTE: The [joint efforts data](https://github.com/StanfordVL/BEHAVIOR-1K/blob/main/OmniGibson/omnigibson/learning/utils/eval_utils.py#L90) in the robot state entry of the parquet files are wrong. This is because we do not store observations during our initial data collection, and all observations are collected through a round of "data replay" in which we restore sim state every step without stepping physics, and thus the joint effort reading is wrong. Please do not use them for training. They will be removed in the next dataset release.**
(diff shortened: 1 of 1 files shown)


wensi-ai (commented):
lgtm

hang-yin (approved):

PR #1876, opened 2025-10-26
fix(setup.sh): expand --cuda-version arg instead of literal $2

yerbymatey:
prev line used `CUDA_VERSION="\$2"` preventing variable expansion and left `$2` as a literal string so setting `--cuda-version` flag didn't properly override default 12.4 value

found it while testing on cuda 13, wheel url came out as cu$2


yerbymatey:
hey @hang-yin , should've hit yall up first before suggesting contribs but i can go back and address linter so it passes ci + reparent to og_develop to adhere to yall's guidelines. lmk what you wanna do! will follow up in discord

wensi-ai:
Thanks for catching this, we have included this in the Isaac 5.1 upgrade PR. See #1814

PR #1905, opened 2025-11-17
Refactor robot subclasses into single robot class + YAML definition files

Andorexad:
This PR refactors the robot-related code.  Key changes:
- all robot classes are merged into one class called Robot()
- ControllableObject is also merged into Robot()
- for each robot, there is a YAML file in robot_configs/ that contains robot-config args


OmniGibson/omnigibson/objects/controllable_object.py
cgokmen, line 1: As a meta-note, I think controllableobject and robot should also be unified into a single class (Robot). There's no non-robot controllable object IIUC - but do check with Hang and Josiah.

OmniGibson/omnigibson/objects/controllable_object.py
```diff
@@ -947,7 +947,7 @@ def reset_joint_pos(self, value):
         self._reset_joint_pos = value
 
     @property
-    @abstractmethod
+    # @abstractmethod
```
cgokmen: Why, and also remove it fully if you have to, dont comment

OmniGibson/omnigibson/robots/robot_configs/a1.yaml
```diff
@@ -0,0 +1,79 @@
...
+  arm_joint_names:
+    "0": ['arm_joint1', 'arm_joint2', 'arm_joint3', 'arm_joint4', 'arm_joint5', 'arm_joint6']
+    
+
+  gripper:
+    _eef_link_names: "eef_link"
+    eef_link_names:
+      "0": "eef_link"
+    _finger_link_names: ["gripper1", "gripper2"]
+    finger_link_names:
```
cgokmen: In a config file there should not be arguments that start with _ in their name - that should be handled in Python if necessary. I think the name overlap can be resolved by changing the dicts to be called finger_link_names_by_arm etc.

OmniGibson/omnigibson/robots/locomotion_robot.py
cgokmen, line 1: These classes should also be unified. The end goal is exactly ONE robot class "Robot".

OmniGibson/omnigibson/robots/manipulation_robot.py
cgokmen, line 1: These classes should also be unified. The end goal is exactly ONE robot class "Robot".

OmniGibson/omnigibson/robots/active_camera_robot.py
cgokmen, line 1: These classes should also be unified. The end goal is exactly ONE robot class "Robot".

OmniGibson/omnigibson/robots/behavior_robot.py
cgokmen, line 1: Can you check with everybody that they are OK discontinuing this robot?

OmniGibson/omnigibson/robots/robot_base.py
```diff
@@ -181,6 +181,21 @@ def _post_load(self):
...
+            for wheel_name in self.floor_touching_base_link_names:
+                wheel_link = self.links[wheel_name]
+                assert set(wheel_link.collision_meshes) == {"collisions"}, "Wheel link should only have 1 collision!"
+                wheel_link.collision_meshes["collisions"].set_collision_approximation("boundingSphere")
+        elif self.model_name.lower() == "tiago":
+            # The eef gripper links should be visual-only. They only contain a "ghost" box volume
+            # for detecting objects inside the gripper, in order to activate attachments (AG for Cloths).
+            for arm in self.arm_names:
+                self.eef_links[arm].visual_only = True
+                self.eef_links[arm].visible = False
```
cgokmen: We don't want to have ANY robot-type-specific code ANYWHERE in our codebase. These should all be gated with some mechanism that can be configured in the robot file. e.g. a field `force_sphere_wheel_approximation` for the first one, `visual_only_eef_links` for the latter?

OmniGibson/omnigibson/robots/two_wheel_robot.py
cgokmen, line 1: These classes should also be unified. The end goal is exactly ONE robot class "Robot".

OmniGibson/omnigibson/utils/python_utils.py
```diff
@@ -192,12 +192,22 @@ def get_class_init_kwargs(cls):
...
         cls (object): Class from which to grab __init__ kwargs
 
     Returns:
         list: All keyword arguments (excluding "self") specified by @cls __init__ constructor method
     """
+    # Check if class has stored init param names from YAML config
+    if hasattr(cls, "_init_param_names_from_yaml"):
+        return cls._init_param_names_from_yaml
+
+    # Fall back to signature inspection for regular classes
```
cgokmen: At the end there will be just one class, so this should not be necessary.

OmniGibson/omnigibson/robots/__init__.py
```diff
@@ -1,28 +1,33 @@
...
+from omnigibson.robots.robot_configs.franka import FrankaPanda
+from omnigibson.robots.robot_configs.franka_mounted import FrankaMounted
+from omnigibson.robots.robot_configs.freight import Freight
+from omnigibson.robots.robot_configs.husky import Husky
+from omnigibson.robots.robot_configs.locobot import Locobot
+from omnigibson.robots.robot_configs.r1 import R1
+from omnigibson.robots.robot_configs.r1pro import R1Pro
+from omnigibson.robots.robot_configs.stretch import Stretch
+from omnigibson.robots.robot_configs.tiago import Tiago
+from omnigibson.robots.robot_configs.turtlebot import Turtlebot
```
cgokmen: Instead of having extra classes like this, follow my recommendation on the top-level PR comment.

OmniGibson/omnigibson/robots/robot_config_parser.py
cgokmen, line 1: the structure of this will need to change as you follow the top-level comment to get rid of the subclassing, there's a lot that is needed only for the dynamic subclassing here. so go ahead and do that and I'll give you a more in depth review.

OmniGibson/omnigibson/robots/robot_configs/franka_mounted.yaml
```diff
@@ -0,0 +1,118 @@
...
+  urdf_path: "models/franka/franka_mounted/urdf/franka_mounted.urdf"
+  curobo_path: "models/franka/franka_mounted/curobo/franka_mounted_description_curobo_default.yaml"
+
+
+  arm_link_names:
+    "0": ['panda_link0', 'panda_link1', 'panda_link2', 'panda_link3', 'panda_link4', 'panda_link5', 'panda_link6', 'panda_link7']
+  arm_joint_names:
+    "0": ['panda_joint1', 'panda_joint2', 'panda_joint3', 'panda_joint4', 'panda_joint5', 'panda_joint6', 'panda_joint7']
+
+  gripper:
```
wensi-ai: I don't think franka_mounted support these end effectors, could you double check?

OmniGibson/omnigibson/robots/robot_configs/kinova.yaml
```diff
@@ -0,0 +1,39 @@
...
...], ['left_outer_knuckle', 'left_inner_knuckle'], ['left_outer_knuckle', 'right_outer_knuckle'], ['left_outer_knuckle', 'right_outer_finger'], ['left_outer_knuckle', 'right_inner_finger'], ['left_outer_knuckle', 'right_inner_knuckle'], ['left_outer_finger', 'left_inner_finger'], ['left_outer_finger', 'left_inner_knuckle'], ['left_outer_finger', 'right_outer_knuckle'], ['left_outer_finger', 'right_outer_finger'], ['left_outer_finger', 'right_inner_finger'], ['left_outer_finger', 'right_inner_knuckle'], ['left_inner_finger', 'left_inner_knuckle'], ['left_inner_finger', 'right_outer_knuckle'], ['left_inner_finger', 'right_outer_finger'], ['left_inner_finger', 'right_inner_finger'], ['left_inner_finger', 'right_inner_knuckle'], ['left_inner_knuckle', 'right_outer_knuckle'], ['left_inner_knuckle', 'right_outer_finger'], ['left_inner_knuckle', 'right_inner_finger'], ['left_inner_knuckle', 'right_inner_knuckle'], ['right_outer_knuckle', 'right_outer_finger'], ['right_outer_knuckle', 'right_inner_finger'], ['right_outer_knuckle', 'right_inner_knuckle'], ['right_outer_finger', 'right_inner_finger'], ['right_outer_finger', 'right_inner_knuckle'], ['right_inner_finger', 'right_inner_knuckle']]
```
wensi-ai: nit: could you make all the lists in yamls into the format similar to that of the r1.disabled_collision_pairs so that we don't have very long lines?

wensi-ai: Also would be great to have all the spacing consistent across configs

joylo/gello/robots/sim_robot/og_teleop_utils.py
```diff
@@ -1023,9 +1022,9 @@ def optimize_sim_settings(vr_mode=False):
     settings.set("/app/vsync", True)
 
 def setup_ghost_robot_info(ghost, robot):
-    if isinstance(robot, R1Pro):
+    if robot.roobot_type_name=="r1pro":
         robot_arm_dof = 7
-    elif isinstance(robot, R1):
+    elif robot.roobot_type_name=="r1":
         robot_arm_dof = 6
```
wensi-ai, line 1026: typo

OmniGibson/tests/test_controllers.py
```diff
@@ -305,13 +308,13 @@ def check_ori_error(curr_orientation, init_orientation, tol=0.1):
                         if pos_check is not None:
                             is_valid_pos = pos_check(target_pos, curr_pos, init_pos)
                             assert is_valid_pos, (
-                                f"Got mismatch for controller [{controller}], mode [{controller_mode}], robot [{robot.model_name}], action [{action_name}]\n"
+                                f"{robot.model_name} 11Got mismatch for controller [{controller}], mode [{controller_mode}], robot [{robot.model_name}], action [{action_name}]\n"
```
wensi-ai: typo

OmniGibson/omnigibson/robots/robot_configs/a1.yaml
```diff
@@ -0,0 +1,44 @@
+capabilities:
```
wensi-ai: I think we should remove the files under configs/robots (they're not used anywhere I think), and move these yaml files over (so we have all configs in one place), but @cgokmen should confirm this.

wensi-ai (commented):
Left some comments. Overall LGTM

joylo/gello/robots/sim_robot/og_sim.py
```diff
@@ -72,9 +70,18 @@ def __init__(
         for rule in DISABLED_TRANSITION_RULES:
             rule.ENABLED = False
 
-        robot_cls = REGISTERED_ROBOTS.get(robot, None)
-        assert robot_cls is not None, f"Got invalid OmniGibson robot class: {robot}"
-        assert issubclass(robot_cls, ManipulationRobot), f"Robot class {robot} is not a manipulation robot! Cannot use GELLO"
+        # Check if robot is valid and supports manipulation by reading from YAML config
+        robot_config_dir = os.path.dirname(inspect.getfile(Robot))
```
cgokmen: This is a lot of work for an assertion. Just move the assertion to happen after the og.Environment load, so you dont have to do this loading yourself.

joylo/gello/robots/sim_robot/og_sim.py
```diff
@@ -386,7 +393,7 @@ def command_joint_state(self, joint_state: th.tensor, component=None) -> None:
         """
         # If R1, process manually
         state = joint_state.clone()
-        if isinstance(self.robot, R1) and not isinstance(self.robot, R1Pro):
+        if self.robot.model_name=="r1" and self.robot.model_name!="r1pro":
```
cgokmen: All of these in this file have formatting issues (space before/after == or !=). 

More importantly, the second condition here is trivially true if the first one is.

joylo/gello/robots/sim_robot/og_teleop_utils.py
```diff
@@ -276,7 +275,7 @@ def setup_cameras(robot, external_sensors, resolution):
     og.sim.viewer_camera.image_width = resolution[1]
 
     # Adjust wrist cameras for R1
-    if isinstance(robot, R1) and not isinstance(robot, R1Pro):
+    if robot.model_name.startswith("r1"):
```
cgokmen: Both here and in the above file, for this r1-or-r1pro check I think it would be better if you just did `robot.model_name in ("r1, "r1pro")`.  Like dont rely on the thing being a prefix etc.

OmniGibson/omnigibson/action_primitives/curobo.py
```diff
@@ -160,7 +160,7 @@ def __init__(
 
             robot_cfg_obj = lazy.curobo.types.robot.RobotConfig.from_dict(robot_cfg_dict, self._tensor_args)
 
-            if isinstance(robot, HolonomicBaseRobot):
+            if isinstance(robot, Robot) and robot.is_holonomic_base:
```
cgokmen: I think it's safe to assume here that robot is a Robot instance.

OmniGibson/omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -61,28 +47,26 @@
 m = create_module_macros(module_path=__file__)
 
 m.KP_LIN_VEL = {
```
cgokmen: move into robot definition files

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -507,13 +507,13 @@ def clear(self):
...
     def get_column_filters(cls):
-        from omnigibson.robots.manipulation_robot import ManipulationRobot
+        from omnigibson.robots.robot import Robot
 
         filters = dict()
         for scene_idx, scene in enumerate(og.sim.scenes):
             filters[scene_idx] = []
             for robot in scene.robots:
-                if isinstance(robot, ManipulationRobot):
+                if isinstance(robot, Robot) and robot.is_manipulation:
```
cgokmen: first part is guaranteed, remove

OmniGibson/omnigibson/simulator.py
```diff
@@ -1243,7 +1243,7 @@ def _on_pre_physics_step(self):
                 # Run the controller step on every controllable object
                 for scene in self.scenes:
                     for obj in scene.objects:
```
cgokmen: use scene.robots

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -507,13 +507,13 @@ def clear(self):
 class GripperRigidContactAPIImpl(RigidContactAPIImpl):
     @classmethod
     def get_column_filters(cls):
-        from omnigibson.robots.manipulation_robot import ManipulationRobot
+        from omnigibson.robots.robot import Robot
```
cgokmen: remove

OmniGibson/tests/test_robot_states_flatcache.py
```diff
@@ -149,7 +150,7 @@ def test_robot_load_drive():
     og.sim.stop()
 
     # Iterate over all robots and test their motion
-    for robot_name, robot_cls in REGISTERED_ROBOTS.items():
+    for robot_name in REGISTERED_ROBOTS.items():
```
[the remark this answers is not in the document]
cgokmen: Clear bug here. Make sure these tests actually run

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -304,6 +304,7 @@ def update_goal(self, command, control_dict):
         ), f"Commands must be dimension {self.command_dim}, got dim {len(command)} instead."
 
         # Preprocess and run internal command
+
```
cgokmen: undo

OmniGibson/omnigibson/examples/action_primitives/rs_int_example.py
```diff
@@ -8,7 +8,7 @@
     StarterSemanticActionPrimitives,
     StarterSemanticActionPrimitiveSet,
 )
-from omnigibson.robots.tiago import Tiago
+from omnigibson.robots.robot import Robot
```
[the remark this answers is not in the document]
cgokmen: remove

OmniGibson/omnigibson/envs/env_base.py
```diff
@@ -280,13 +286,9 @@ def _load_robots(self):
...
                     )
 
-                # Make sure robot exists, grab its corresponding kwargs, and create / import the robot
-                robot = create_class_from_registry_and_config(
-                    cls_name=robot_config["type"],
-                    cls_registry=REGISTERED_ROBOTS,
-                    cfg=robot_config,
-                    cls_type_descriptor="robot",
-                )
+                cls_kwargs = extract_class_init_kwargs_from_dict(cls=Robot, dic=robot_config, copy=False)
```
cgokmen: What's the point of this?

OmniGibson/omnigibson/scenes/scene_base.py
```diff
@@ -15,7 +15,8 @@
 from omnigibson.macros import gm
 from omnigibson.objects.object_base import BaseObject
 from omnigibson.prims.xform_prim import XFormPrim
-from omnigibson.robots.robot_base import REGISTERED_ROBOTS, m as robot_macros
+from omnigibson.robots.robot import m as robot_macros
```
cgokmen: Importing another module's macros should not be a thing. Let's move that macro to constants.py and remove this import.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+}
+GraspingPoint = namedtuple("GraspingPoint", ["link_name", "position"])  # link_name (str), position (x,y,z tuple)
+
+
+class Robot(USDObject, GymObservable):
+    def __init__(
+        self,
+        # Shared kwargs in hierarchy
+        name,
+        robot_type_name,
```
cgokmen: Let's use "model" instead of robot_type_name and also robot.model_name. Unify

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+            cfg = yaml.safe_load(f) or {}
+            self._robot_cfg = cfg["property"]
+            self._init_capabilities(cfg["capabilities"])
+
+        # Unique to Tiago
+        if "support_variant" in self._robot_cfg.keys():
+            if default_arm_pose is None:
+                default_arm_pose = "diagonal15"
+            assert variant in ("default", "wrist_cam"), f"Invalid Tiago variant specified {variant}!"
+            self._variant = variant
```
cgokmen: Add into definition a key for what should be the default arm pose for that robot, default_arm_pose_key="vertical" to be diagonal15 for Tiago, and when default_arm_pose is None in this script, and make default_arm_pose default to definition.default_arm_pose_key instead of just "vertical".

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+                If reset_joint_pos is not None, this will be ignored (since _default_joint_pos won't be used during initialization).
+            end_effector (str): The end effector type to use.
+            kwargs (dict): Additional keyword arguments that are used for other super() calls from subclasses, allowing
+                for flexible compositions of various object subclasses (e.g.: Robot is USDObject + ControllableObject).
+        """
+        self.robot_type_name = robot_type_name.lower()
+        # Read robot config YAML file
+        robot_config_dir = os.path.dirname(__file__)
+        robot_cfg_path = os.path.join(robot_config_dir, "robot_configs", self.robot_type_name + ".yaml")
+        with open(robot_cfg_path, "r") as f:
```
cgokmen: These new yaml files that define the robot format that used to be in the Python class files should be referred to as **model definitions** and not configs.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+            cfg = yaml.safe_load(f) or {}
```
cgokmen: Use OmegaConf + dataclasses to define, parse, and enforce schema for definition files

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+                default_arm_pose = "diagonal15"
+            assert variant in ("default", "wrist_cam"), f"Invalid Tiago variant specified {variant}!"
+            self._variant = variant
+
+        if "supported_end_effector" in self._robot_cfg.keys():
+            self.end_effector = end_effector
+            grasping_direction = "lower" if end_effector == "gripper" else "upper"
+            self._init_ag_points()
+
+        if self.is_manipulation:
```
cgokmen, line 227: Let's add functioanlity-based config groups like the below

```
@dataclass
class RobotDefinition:
  usd_path: str

  manipulation: Optional[ManipulationDefinition]

@dataclass
class ManipulationDefinition:
  ag_config: ...
```




and then here

```
@property
def is_manipulation(self):
  return self.definition.manipulation is not None
```

OmniGibson/omnigibson/robots/robot_configs/fetch.yaml
```diff
@@ -0,0 +1,39 @@
+capabilities: 
+  - is_two_wheel
```
cgokmen: Remove is_ from these

OmniGibson/omnigibson/robots/robot_configs/fetch.yaml
```diff
@@ -0,0 +1,39 @@
+capabilities:
```
cgokmen: Let's remove the string list of capabilities and distribute them to subdefinitions and/or just simple true/false fields on the main robot definition if the capability does not require any config

OmniGibson/omnigibson/robots/robot.py
cgokmen: Remove the variant feature.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+        if self.is_mobile_manipulation:
+            assert_valid_key(key=default_reset_mode, valid_keys=RESET_JOINT_OPTIONS, name="default_reset_mode")
+            self.default_reset_mode = default_reset_mode
+
+        if self.is_untucked_arm_pose:
+            if default_arm_pose is not None:
+                assert_valid_key(key=default_arm_pose, valid_keys=self.default_arm_poses, name="default_arm_pose")
+            else:
+                default_arm_pose = "vertical"
+            self.default_arm_pose = default_arm_pose
```
cgokmen: Fold this into mobile manipulator also

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+                for arm, finger_link_names in self.finger_link_names.items():
+                    for finger_link_name in finger_link_names:
+                        if finger_link_name not in self._link_physics_materials:
+                            self._link_physics_materials[finger_link_name] = dict()
+                        if finger_static_friction is not None:
+                            self._link_physics_materials[finger_link_name]["static_friction"] = finger_static_friction
+                        if finger_dynamic_friction is not None:
+                            self._link_physics_materials[finger_link_name]["dynamic_friction"] = finger_dynamic_friction
+
+    def _init_capabilities(self, capabilities):
```
cgokmen: Use @property-based functions to return these instead of setting fields

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+                # Call the super() method to move the robot frame first
+                self._base_set_position_orientation(position, orientation, frame)
+                # Move the joint frame for the world_base_joint
+                if self._world_base_fixed_joint_prim is not None:
+                    self._world_base_fixed_joint_prim.GetAttribute("physics:localPos0").Set(tuple(position))
+                    self._world_base_fixed_joint_prim.GetAttribute("physics:localRot0").Set(
+                        lazy.pxr.Gf.Quatf(*orientation[[3, 0, 1, 2]].tolist())
+                    )
+            return
+        if self.is_manipulation:
```
cgokmen, line 1206: Leave a gap

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+        )
+
+        # Concatenate and return
+        state_flat = th.cat([state_flat, controller_states_flat])
+        if self.is_manipulation:
+            # No additional serialization needed if we're using physical grasping
+            if self.grasping_mode == "physical":
+                return state_flat
+
+            # TODO AG
```
cgokmen: Remove the TODO AG comment here and in deserialize

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+        return pos
+
+    @property
+    def tucked_default_joint_pos(self):
+        assert self.is_mobile_manipulation
+        return self._get_tucked_default_joint_pos(self._robot_cfg["tucked_default_joint_pos"])
+
+    @property
+    def untucked_default_joint_pos(self):
+        if self.model_name == "r1":
```
cgokmen: Figure out how to get rid of these. In general, find and remove any reference to "model_type == xyz" logic inside this class

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+            # Keep the current joint positions for the base joints
+            pos[self.base_idx] = self.get_joint_positions()[self.base_idx]
+            for arm in self.arm_names:
+                pos[self.gripper_control_idx[arm]] = th.tensor([0.05, 0.05])  # open gripper
+                pos[self.arm_control_idx[arm]] = th.tensor([0.0, 1.906, -0.991, 1.571, 0.915, -1.571])
+            return pos
+        if self.model_name == "r1pro":
+            pos = th.zeros(self.n_dof)
+            # Keep the current joint positions for the base joints
+            pos[self.base_idx] = self.get_joint_positions()[self.base_idx]
```
cgokmen: I think the point of this logic is that it prevents accidentally resetting the holonomic joints' positions when resetting to the untucked/tucked positions. Instead of this, whenever this untucked_default_joint_pos is being used, make sure that it only includes the real joints (not the holonomic ones), and DONT reset the holonomic joints. Also add a unit test for this: resetting a holonomic joint robot to tucked/untucked should not move the robot's body.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+        return obs_keys
+
+    @property
+    def model_name(self):
+        """
+        Returns:
+            str: name of this robot model. usually corresponds to the class name of a given robot model
+        """
+        if "supported_end_effector" in self._robot_cfg.keys():
+            return self._robot_cfg[self.end_effector].get("model_name", self.robot_type_name)
```
cgokmen: Don't do these variant typenames. The type name should simply be the same as the filename of the robot definition file.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4412 @@
...
+        self._action_normalize = action_normalize
+
+        # Store internal placeholders that will be filled in later (from ControllableObject)
+        self._dof_to_joints = None  # dict that will map DOF indices to JointPrims
+        self._last_action = None
+        self._controllers = None
+        self.dof_names_ordered = None
+        self._control_enabled = True
+
+        class_name = self.robot_type_name.lower()
```
cgokmen: Instead of the typename, create a property that uniquely defines the controllable variant of the robot. The goal is that everything that has the same the kinematic tree (same joint hierarchy) should have the same kinematic tree ID such the controllable object view APIs can assume same shape for everything that has the same kinematic tree identifier.

```
@property
def kinematic_tree_identifier(self):
  return self.type_name + "_" + ...
```

OmniGibson/omnigibson/robots/robot_configs/a1.yaml
cgokmen: Rename this directory to be `definitions`

OmniGibson/omnigibson/envs/env_base.py
Andorexad: REGISTERED_ROBOTS is a list now and doesn't work with create_class_from_registry_and_config.

joylo/gello/robots/sim_robot/og_sim.py
```diff
@@ -4,20 +4,16 @@
 import numpy as np
 from typing import Dict, Optional
 import json
-
+import inspect
```
cgokmen: Is this used?

joylo/gello/robots/sim_robot/og_sim.py
```diff
@@ -32,6 +28,7 @@
 import gello.robots.sim_robot.og_teleop_utils as utils
 
 from bddl.activity import Conditions
+import yaml
```
cgokmen: Is this used?

OmniGibson/omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -1688,13 +1649,11 @@ def _navigate_to_pose_direct(self, pose_2d, low_precision=False):
                     assert (
                         self.robot.controllers["base"].motor_type == "velocity"
                     ), "Holonomic base controller must be in velocity mode"
-                    direction_vec = (
-                        body_target_pose[0][:2] / th.norm(body_target_pose[0][:2]) * m.KP_LIN_VEL[type(self.robot)]
-                    )
+                    direction_vec = body_target_pose[0][:2] / th.norm(body_target_pose[0][:2]) * self.robot.kp_lin_vel
```
cgokmen: Can we rename these kp_lin_vel and kp_ang_vel to "linear_velocity_gain_for_primitives" and "angular_velocity_gain_for_primitives" here and on the robot just to be clearer.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -88,7 +88,7 @@ def global_update(cls):
             {
                 link.prim_path
                 for robot in scene.robots
-                if isinstance(robot, ManipulationRobot)
+                if isinstance(robot, Robot) and robot.manipulation
```
cgokmen: Remove isinstance

OmniGibson/omnigibson/robots/definition_schema.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Root configuration for a robot, with optional capability sub-configs.
+    """
+
+    raw_controller_order: List[str] = field(default_factory=list)
+    default_controllers: Dict[str, str] = field(default_factory=dict)
+    default_joint_pos: Optional[List[Any]] = None
+    usd_path: Optional[str] = None
+    urdf_path: Optional[str] = None
+    curobo_path: Optional[str] = None
```
cgokmen, line 112: Can you make a pass through all the configs in this file and find out exactly what's optional? Like for example usd_path is NOT optional. I'm guessing most of these things actually arent. So let's enforce them.

OmniGibson/omnigibson/robots/definition_schema.py
```diff
@@ -0,0 +1,128 @@
...
+    untucked_default_joint_pos: Optional[List[Any]] = None
+    tucked_default_joint_pos: Optional[List[Any]] = None
+    # Multiple arm pose support (optional)
+    default_arm_pose_key: Optional[str] = None
+    default_arm_poses: Optional[Dict[str, List[Any]]] = None
+
+
+# === Main Robot Definition ===
+
+
```
cgokmen: Remove unnecessary space

OmniGibson/omnigibson/robots/definition_schema.py
```diff
@@ -0,0 +1,128 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+
+# === Capability-specific definitions ===
+
+
```
cgokmen: Remove unnecessary space

OmniGibson/tests/test_controllers.py
```diff
@@ -16,39 +15,42 @@ def test_arm_control():
         "objects": [],
         "robots": [
             {
-                "type": "FrankaPanda",
+                "model": "Franka",
```
cgokmen: This should be case-sensitive, and we should change to lowercase.

OmniGibson/tests/test_robot_states_flatcache.py
```diff
@@ -274,8 +276,9 @@ def object_is_in_hand(robot, obj, grasping_mode):
             raise ValueError(f"Unknown grasping mode: {grasping_mode}")
 
     for grasping_mode in grasping_modes:
-        robot = Fetch(
+        robot = Robot(
             name="Fetch",
+            model="Fetch",
```
cgokmen: This should be case-sensitive, and we should change to lowercase.

OmniGibson/tests/test_robot_states_no_flatcache.py
```diff
@@ -45,3 +46,60 @@ def test_object_in_FOV_of_robot():
...
+    ), f"tuck() changed base joint positions! Initial: {initial_base_joint_pos}, After tuck: {base_joint_pos_after_tuck}"
+
+    # Test untuck() - should preserve base joint positions
+    robot.untuck()
+    base_joint_pos_after_untuck = robot.get_joint_positions()[robot.base_idx]
+    assert th.allclose(
+        initial_base_joint_pos, base_joint_pos_after_untuck, atol=1e-6
+    ), f"untuck() changed base joint positions! Initial: {initial_base_joint_pos}, After untuck: {base_joint_pos_after_untuck}"
+
+    og.clear()
```
cgokmen, line 111: As it is set up right now, I don't think this test is very interesting, and it strays quite a lot from the default test setup in this file. Is there anything important that it checks? If not, let's remove it.

OmniGibson/tests/test_robot_teleoperation.py
```diff
@@ -15,6 +15,7 @@ def test_teleop():
         "robots": [
             {
                 "type": "Fetch",
+                "model": "Fetch",
```
cgokmen: See comment on `env_base.py`.

OmniGibson/omnigibson/envs/env_base.py
```diff
@@ -270,6 +271,11 @@ def _load_robots(self):
                 # Add a name for the robot if necessary
                 if "name" not in robot_config:
                     robot_config["name"] = "robot_" + "".join(random.choices(string.ascii_lowercase, k=6))
+                if "model" not in robot_config:
+                    assert (
+                        robot_config["type"] in REGISTERED_ROBOTS
+                    ), f"{robot_config['type']} is not a registered robot."
+                    robot_config["model"] = robot_config["type"]
```
cgokmen: Remove the assertion - why would this be checked only for `type` and not `model`? Let the robot init check this.

More importantly, update the logic to look like this:

```
if "model" is in config:
  assert "type" is not in config, "CANNOT SPECIFY BOTH TYPE AND MODEL"
elif "type" is in config:
   # TODO(andi) write better message
  log.warning("Type field is deprecated. Please switch to model. New model IDs are lowercase.")
  robot_config["model"] = robot_config["type"].lower()
  del robot_config["type"]
```

OmniGibson/omnigibson/envs/env_base.py
```diff
@@ -280,13 +286,9 @@ def _load_robots(self):
...
-                # Make sure robot exists, grab its corresponding kwargs, and create / import the robot
-                robot = create_class_from_registry_and_config(
-                    cls_name=robot_config["type"],
-                    cls_registry=REGISTERED_ROBOTS,
-                    cfg=robot_config,
-                    cls_type_descriptor="robot",
-                )
+                cls_kwargs = extract_class_init_kwargs_from_dict(cls=Robot, dic=robot_config, copy=False)
+
+                robot = Robot(**cls_kwargs)
```
cgokmen: I don't think this is necessary. Just do **robot_config.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4419 @@
...
+m.MAX_ASSIST_FORCE = 100
+m.MIN_AG_DEFAULT_GRASP_POINT_PROP = 0.2
+m.MAX_AG_DEFAULT_GRASP_POINT_PROP = 0.95
+m.AG_DEFAULT_GRASP_POINT_Z_PROP = 0.4
+
+m.CONSTRAINT_VIOLATION_THRESHOLD = 0.1
+m.GRASP_WINDOW = 3.0  # grasp window in seconds
+m.RELEASE_WINDOW = 1 / 30.0  # release window in seconds
+
+# ---HolonomicBase---
```
cgokmen: Remove comment

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4419 @@
...
+
+    @property
+    def manipulation(self) -> bool:
+        """Returns True if this robot has manipulation capabilities."""
+        return self._definition.manipulation is not None or self.mobile_manipulation or self.articulated_trunk
+
+    @property
+    def locomotion(self) -> bool:
+        """Returns True if this robot has locomotion capabilities."""
+        return self.holonomic_base or self.two_wheel or self._definition.locomotion is not None
```
cgokmen: Sorry I wasn't super clear about this before, and I know this is going to cause you a bit of pain, but can we convert all of these to be called "is_x" again? I just meant that the definition's fields should be called locomotion etc. (which you did correctly implement) but these properties on the robot should still be called is_locomotion etc.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4419 @@
...
+            return None
+        return self._definition.manipulation.end_effectors.get(self.end_effector)
+
+    @property
+    def kinematic_tree_identifier(self):
+        if self._definition.manipulation and self._definition.manipulation.supported_end_effector:
+            eef_def = self._get_end_effector_definition()
+            if eef_def and eef_def.model:
+                return eef_def.model
+        return self.model
```
cgokmen, line 469: Add a docstring for what this does

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -78,7 +78,7 @@ def get_optional_dependencies(cls):
     @classmethod
     def global_update(cls):
         # Avoid circular imports
-        from omnigibson.robots.manipulation_robot import ManipulationRobot
+        from omnigibson.robots.robot import Robot
```
cgokmen: Remove

OmniGibson/omnigibson/examples/robots/all_robots_visualizer.py
```diff
@@ -22,13 +22,11 @@ def main(random_selection=False, headless=False, short_exec=False):
...
-    for robot_name, robot_cls in REGISTERED_ROBOTS.items():
-        if robot_name in ["BehaviorRobot"]:
-            continue
-
+    for robot_name in REGISTERED_ROBOTS:
         # Create and import robot
-        robot = robot_cls(
+        robot = Robot(
             name=robot_name,
+            model=robot_name.lower(),
```
cgokmen: Remove lower

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -0,0 +1,4421 @@
...
+            finger_dynamic_friction (None or float): If specified, specific dynamic friction to use for robot's fingers.
+                Note: If specified, this will override any ways that are found within @link_physics_materials for any
+                robot finger gripper links
+            default_reset_mode (str): Default reset mode for the robot. Should be one of: {"tuck", "untuck"}
+                If reset_joint_pos is not None, this will be ignored (since _default_joint_pos won't be used during initialization).
+            end_effector (str): The end effector type to use.
+            kwargs (dict): Additional keyword arguments that are used for other super() calls from subclasses, allowing
+                for flexible compositions of various object subclasses (e.g.: Robot is USDObject).
+        """
+        self.model = model.lower()
```
cgokmen: Remove this

PR #1916, opened 2025-11-21
Working WebRTC Streaming for OmniGibson (Replaces Deprecated Streamclient)

yerbymatey:
WebRTC streaming supporting ICE/SDP + TCP only WS streaming to cover the now deprecated `omni.services.streamclient.webrtc` and their built in js viewer for local machines and cloud instances. Setup is mainly **env driven** (grab and set public ip if you have one, any mapped ports, pick between streaming modes). There's a small [setup script](OmniGibson/remote_viewer/setup.sh) to grab Nvidia's WebRTC streaming library since we can't ship it directly :/ Setup guide is [here](OmniGibson/remote_viewer/README.md).

Kit directly serves a [static site](OmniGibson/remote_viewer/kit/exts/omnigibson.remote_viewer.web/web-ui/index.html) with some custom [extensions](OmniGibson/remote_viewer/kit/exts/) to help out with session coordination and some stopgaps for Nvidia's event plumbing.

Tested on two Ubuntu 22.04 boxes under v different network setups:
- local + Tailscale
- local + external IP + port-forwarding
- local + firewalled UDP (to test TCP only websocket)
- provider full VM (Vast.ai) with mapped UDP/TCP

Haven’t tested an unprivileged Runpod-style Docker image yet with TCP only and proxied HTTPS, but the design should survive it since the backend URL can be pulled from origin headers and the SPA routing lives in the [web extension](OmniGibson/remote_viewer/kit/exts/omnigibson.remote_viewer.web/omnigibson/remote_viewer/web/extension.py). TCP only streams are view-only rn, so keyboard or mouse events don't get registered.

_Note on extension load + enabling_: The custom exts use a [lazy ExtensionManager](OmniGibson/omnigibson/simulator.py#L234) to set the [extension path](OmniGibson/omnigibson/macros.py#L152) during kit runtime instead of the app dir copy pattern y'all had. Ext loading is kind of in this weird halfway place right now where `isaacsim.core.utils.extension.enable_extension()` calls will work for ext mgmt, but will sometimes fall short for recursive dep solving.

Tracking Nvidia's doc drift to swap out and extend functionality for exts that still worked with Isaac Sim 4.5 + Kit 106.5 was absolute dogwater but streaming should be way more straightforward once y'all bump to Isaac Sim 5.0 lol

### DIFF
--- a/OmniGibson/omnigibson/simulator.py
+++ b/OmniGibson/omnigibson/simulator.py
@@ -265,10 +265,10 @@ def _launch_app():
             lazy.isaacsim.core.utils.extension.enable_extension("omni.kit.livestream.native")
             print(f"Now streaming on {ip} via Omniverse Streaming Client")
 
-        elif gm.REMOTE_STREAMING == "webrtc":            
+        elif gm.REMOTE_STREAMING == "webrtc":
             # Set WebRTC port
             app.set_setting("/app/livestream/port", gm.WEBRTC_PORT)
-            
+
             # Enable custom OmniGibson remote viewer extensions
             custom_exts = [
                 "omnigibson.remote_viewer.setup",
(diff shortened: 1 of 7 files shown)


Seyeon321:
Thanks for this PR! 
I tried it while setting up remote visualization, and it clarified a lot of the issues I was running into with the deprecated streaming path. 
Really appreciate the work :) !

PR #1921, opened 2025-12-01
Feat/leaderboard

wensi-ai:
Leaderboard for 2025 Behavior Challenge


hang-yin (approved):
LGTM, can we also 
1) open a ticket to move this leaderboard to huggingface and dynamically read from there 
2) add a note at the top of the leaderboard page saying we will update this to have more information, including per-task ranking and efficiency metrics, etc.

PR #1927, opened 2025-12-03
Commit evaluation system used for challenge

cgokmen:


wensi-ai:
overall lgtm. Some high level questions:

1. For submission wrappers, most of them are just using the ones we provided, is it ok that we have a lot of duplicated code in there?
2. I'm thinking maybe we should put eval_with_job_queue somewhere else from OmniGibson, but that could be a future todo where we do a proper cleanup on the whole codebase. 
3. Do we still want to keep the north star specific early stopping stuff in our codebase? That's actually sort-of cheating because early stopping will increase the time and movement score

OmniGibson/omnigibson/learning/policies.py
```diff
@@ -52,12 +52,21 @@ def __init__(
         **kwargs,
     ) -> None:
         logging.info(f"Creating websocket client policy with host: {host}, port: {port}")
+        self.last_action = None
+        self.policy = None  # WebsocketClientPolicy(host=host, port=port)
```
wensi-ai: This will probably break the normal non-jobqueue evaluation pipeline. Can we maybe revert it back, or do a if-else that only set self.policy=None if host and port are not provided?

OmniGibson/omnigibson/learning/wrappers/challenge_submissions/submission_tobi.py
```diff
@@ -0,0 +1,39 @@
+import omnigibson as og
```
wensi-ai: We should remove this since we haven't evaluated this and I'm unsure whether this will work.

OmniGibson/omnigibson/learning/utils/network_utils.py
```diff
@@ -55,27 +61,33 @@ def _wait_for_server(self) -> Tuple[websockets.sync.client.ClientConnection, Dic
         host_port = self._uri.replace("ws://", "").replace("wss://", "")
         if ":" in host_port:
             host, port = host_port.split(":")
-            health_url = f"http://{host}:{port}/healthz"
+            health_url = f"https://{host}:{port}/healthz" if int(port) == 443 else f"http://{host}:{port}/healthz"
```
wensi-ai, line 64: I remembered with the changes it won't work on local websocket eval or something, this need to be double checked.

wensi-ai (approved):
LGTM

PR #1937, opened 2025-12-10
Fix curobo holonomic joint limit

hang-yin:
There's this weird issue with curobo prismatic joint planning - @ChengshuLi found it that if we are planning outside of a 5 meters range, curobo would fail super randomly. Thus, in `curobo.py`, we have a macro `m.HOLONOMIC_BASE_PRISMATIC_JOINT_LIMIT` to add an artificial joint limit to the holonomic base joints.

This works fine if we are only moving the robot within a circle of 5m radius - this is why MoMaGen works. But as we move to general behavior tasks, we might want to move across rooms - even the robot start pose might already be more than 5m away from the origin (holonomic base root link fixed there). 

This PR is a temporary fix - we fix some low-hanging fruit so that we can set the root link pose to be closer to the robot workspace. This still doesn't fix everything though - perhaps we might want to implement a simple path planning algorithm for robot base with curobo collision check? This would also help us add diff drive base planning back.

- [x] what do we do about q_to_action?
- [x] dynamically move root link?

### DIFF
--- a/OmniGibson/omnigibson/action_primitives/curobo.py
+++ b/OmniGibson/omnigibson/action_primitives/curobo.py
@@ -116,17 +116,22 @@ class CuRoboMotionGenerator:
                 CuRoboEmbodimentSelection.DEFAULT: robot_cfg_path_dict[CuRoboEmbodimentSelection.DEFAULT]
             }
         print("!" * 100)
+        # TODO [Wensi]: Check whether this is still true for future releases.
         print(
-            "NOTE: Currently for 50 series, only Default embodiment works for Tiago, and only non-DEFAULT embodiment works for R1Pro."
+            """
+            NOTE: Currently (v3.8.0), for cuda architecture 12.0 (e.g. RTX 50-series), using Default embodiment for Tiago or non-DEFAULT embodiment for R1Pro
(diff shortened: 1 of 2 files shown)


OmniGibson/omnigibson/action_primitives/curobo.py
```diff
@@ -115,6 +115,22 @@ def __init__(
             robot_cfg_path_dict = {
                 CuRoboEmbodimentSelection.DEFAULT: robot_cfg_path_dict[CuRoboEmbodimentSelection.DEFAULT]
             }
+        print("!" * 100)
+        print(
+            "NOTE: Currently for 50 series, only Default embodiment works for Tiago, and only non-DEFAULT embodiment works for R1Pro."
+        )
+        if th.cuda.get_device_capability(device) == (12, 0):
+            if robot.model == "tiago":
+                print("Detected you are using Tiago with 50-series GPU: excluding non-DEFAULT embodiment.")
```
cremebrule: Can we actually be very verbose here? Otherwise we're going to forget why it's excluded haha

wensi-ai: done

PR #1957, opened 2026-02-02
Added ImplicitPD controller for robots and fixed open_state metadata

nadunRanawaka1:


asset_pipeline/b1k_pipeline/export_objs_global.py
```diff
@@ -817,12 +817,26 @@ def process_object(root_node, target, relevant_nodes, requried_meta_types, outpu
...
-            for i, joint in enumerate(tree.findall("joint"))
-            if "openable"
-            in out_metadata["link_tags"].get(joint.find("child").attrib["link"], [])
-        ]
+        # Build openable_joint_ids with direction info
+        # Format: [[index, joint_name, direction], ...] where direction is 1 (positive) or -1 (negative)
+        openable_joint_ids = []
+        for i, joint in enumerate(tree.findall("joint")):
+            child_link = joint.find("child").attrib["link"]
+            if "openable" in out_metadata["link_tags"].get(child_link, []):
```
cremebrule: Just so I understand this correctly -- this auto-exports URDFs to have proper joint annotation based on the inferred open direction, where 0 is assumed to be the default (closed) state?

OmniGibson/omnigibson/controllers/implicit_pd_controller.py
```diff
@@ -0,0 +1,351 @@
...
+
+# Create settings for this module
+m = create_module_macros(module_path=__file__)
+
+# Default gains matching Isaac Lab's typical values for Franka
+m.DEFAULT_STIFFNESS = 400.0
+m.DEFAULT_DAMPING = 80.0
+
+
+class ImplicitPDController(LocomotionController, ManipulationController, GripperController):
```
cremebrule: Sorry, I'm a bit confused -- how is this different from `JointController` with `isaac_kp` and `isaac_kd` set?

OmniGibson/omnigibson/envs/data_wrapper.py
```diff
@@ -1291,14 +1292,26 @@ def allocate_traj(
...
             record_data (bool): Whether to record data during playback or not
         """
-        for episode_id in range(self.input_hdf5["data"].attrs["n_episodes"]):
+        if n_episodes is None:
+            episode_ids = range(self.input_hdf5["data"].attrs["n_episodes"])
+        else:
+            if random_sample:
+                import numpy as np
+
+                episode_ids = np.random.choice(
```
cremebrule, line 1313: nit: let's assert that n_episodes <= number of actual episodes in the dataset

OmniGibson/omnigibson/learning/utils/lerobot_utils.py
```diff
@@ -1,5 +1,9 @@
 import argparse
 import av
+
+# Suppress verbose FFmpeg/libx264 encoder output at module load time
+av.logging.set_level(av.logging.PANIC)
```
cremebrule, line 5: I didn't know this was an option LOL

OmniGibson/omnigibson/learning/utils/lerobot_utils.py
```diff
@@ -145,10 +149,10 @@ def encode_video_frames(
     imgs_dir: Path | str,
     video_path: Path | str,
     fps: int,
-    vcodec: str = "libsvtav1",
+    vcodec: str = "h264",
```
cremebrule: I think lerobot's default codec is libsvtav1. Is switching to h264 because this is what DROID uses? Or what's the benefit here again?

OmniGibson/omnigibson/object_states/open_state.py
```diff
@@ -17,8 +17,8 @@
 # Should be a number in the range [0, 1] which will be transformed
 # to a position in the joint's min-max range.
 m.JOINT_THRESHOLD_BY_TYPE = {
-    JointType.JOINT_REVOLUTE: 0.05,
-    JointType.JOINT_PRISMATIC: 0.05,
+    JointType.JOINT_REVOLUTE: 0.10,  # TODO: revert back to 0.05
```
cremebrule: Let's revert this? These macros can be set externally too so that we don't mess up the underlying BEHAVIOR code. We should set these in the CDC repo. Can be set via:

```python
from omnigibson.macros import macros as m
m.object_states.open_state.JOINT_THRESHOLD_BY_TYPE[JointType.JOINT_REVOLUTE] = ...
```

OmniGibson/omnigibson/object_states/open_state.py
```diff
@@ -98,17 +98,28 @@ def _get_relevant_joints(obj):
...
     both_sides = obj.metadata[m.BOTH_SIDES_METADATA_FIELD] if m.BOTH_SIDES_METADATA_FIELD in obj.metadata else False
-    joint_metadata = obj.metadata[m.METADATA_FIELD].items()
-
-    # The joint metadata is in the format of [(joint_id, joint_name), ...] for legacy annotations and
-    # [(joint_id, joint_name, joint_direction), ...] for direction-annotated objects.
-    joint_names = [m[1] for m in joint_metadata]
-    joint_directions = [m[2] if len(m) > 2 else 1 for m in joint_metadata]
+    joint_metadata = obj.metadata[m.METADATA_FIELD]
+
+    # The joint metadata is in the format of {joint_id: [joint_name]} for legacy annotations and
```
cremebrule, line 103: Looping in @cgokmen  here. I think this is a good change that allows for external users to import their custom articulated objects and still use our Open state easily out of the box. Is this good with you?

OmniGibson/omnigibson/object_states/robot_related_states.py
```diff
@@ -17,8 +17,22 @@ def robot(self):
...
+        Check if the robot is grasping the given object.
+        Works with both assisted grasping and physical grasping.
+
+        Uses the robot's built-in is_grasping() method which handles:
+        - Assisted grasping: checks _ag_obj_in_hand
+        - Physical grasping: checks gripper controller state AND finger contact with object
+        """
+        from omnigibson.controllers import IsGraspingState
+
+        for arm in self.robot.arm_names:
```
cremebrule, line 30: Nice! This is clean. Just to make sure, for the assisted grasping case, this is exactly what the code was doing before, correct?

OmniGibson/omnigibson/utils/asset_conversion_utils.py
```diff
@@ -791,9 +791,21 @@ def remove_instanceable_recursive(_root):
 
     # Manually modify metadata
     if "openable_joint_ids" in data["metadata"]:
+        # Convert list format [[idx, name], [idx, name, direction]] to dict format
+<<<<<<< HEAD
```
cremebrule: resolve merge conflict

OmniGibson/omnigibson/learning/utils/lerobot_utils.py
cgokmen: talk about a one-line change with big consequences haha

asset_pipeline/b1k_pipeline/export_objs_global.py
cgokmen: Why is this file getting changed? Are you actually running this file at all? This is part of our asset pipeline specifically and NOT the external user import path.

OmniGibson/omnigibson/object_states/open_state.py
cgokmen: Yes this is OK. Thanks for checking in.

OmniGibson/omnigibson/envs/data_wrapper.py
nadunRanawaka1: Or should we always limit to the number of actual episodes?

OmniGibson/omnigibson/learning/utils/lerobot_utils.py
nadunRanawaka1: h264 is faster for encoding/decoding unless you run on GPU. Shouldn't affect downstream learning, but might increase data usage. 

Reverted for now.

OmniGibson/omnigibson/controllers/implicit_pd_controller.py
nadunRanawaka1: It's a simpler controller when `self._use_explicit_pd=True`, i.e. there is no dynamics compensation. It's just another option if users don't need true impedance control. Might be more specific to the CDC project, so can be removed

OmniGibson/omnigibson/object_states/robot_related_states.py
nadunRanawaka1: This additionally checks `self._ag_release_counter[arm] is None`  for the ag grasping case, but I think this is what we want?

OmniGibson/omnigibson/utils/asset_conversion_utils.py
nadunRanawaka1: Oof, good catch. Had it fixed locally but not pushed. Resolved now.

OmniGibson/omnigibson/object_states/open_state.py
nadunRanawaka1: Reverted.

asset_pipeline/b1k_pipeline/export_objs_global.py
nadunRanawaka1: @cremebrule Yes, this uses a heuristic to infer open direction (joint_position ~= 0 is assumed to be closed). 

@cgokmen I added this to the export pipeline to have future exports agree with the metadata expected in `open_state.py` . 

Was trying to figure out where to include the logic for determining joint direction, maybe `record_obj_metadata_from_urdf()` in `asset_conversion_utils.py` is a better place?

nadunRanawaka1:
Addressed the comments. Let me know the best way to proceed re the following:

1. Including ImplicitPD controller
2. Adding joint direction logic to `asset_pipeline/b1k_pipeline/export_objs_global.py`

OmniGibson/omnigibson/object_states/robot_related_states.py
cremebrule: Ah I see. Yeah that makes sense! So basically making sure that the user isn't trying to actively release the object

OmniGibson/omnigibson/learning/utils/lerobot_utils.py
cremebrule: Thanks. For our use cases with DROID if we find it is faster we can probably modify it on our end, but we'll keep the main BEHAVIOR stack compatible with vanilla lerobot since that's their default

OmniGibson/omnigibson/envs/data_wrapper.py
cremebrule: Hmmm I'm fine either way. I guess in that case, the code will silently only generate up to the number of actual episodes, even if you request more?

OmniGibson/omnigibson/controllers/implicit_pd_controller.py
cremebrule: Wait sorry I'm still confused haha. I think JointController with `motor_type=position` uses the implicit isaac PD implementation, i.e.: it sets the stiffness / damping gains and then directly sets the joint position targets. Is the ImplicitPDController doing something different than this?

asset_pipeline/b1k_pipeline/export_objs_global.py
cremebrule: Thanks for the clarification. I'll defer to @cgokmen 's opinion here, as he is the master of our asset exporting pipeline.

I do generally agree there should be some default convention for populating the joint metadata if it's not annotated explicitly, and I think Nadun's convention makes sense. But should the code live in the asset exporting pipeline or only downstream / at runtime as a fallback if we don't find the explicit metadata file for a given articulated object? My guess is the latter, but I'll defer to you @cgokmen

cgokmen: I see. Please revert to the original form but update the tuple to include the direction, simply defaulting to the "positive is opening" direction value. All objects in the pipeline have the default/lower position being the closed position.

OmniGibson/omnigibson/envs/data_wrapper.py
nadunRanawaka1: Yes, correct. (unless we want to repeat episodes)

asset_pipeline/b1k_pipeline/export_objs_global.py
nadunRanawaka1: Updated

OmniGibson/omnigibson/controllers/implicit_pd_controller.py
nadunRanawaka1: Removed controller and included functionality in JointController to turn off inertia compensation.

OmniGibson/omnigibson/envs/data_wrapper.py
cremebrule: Gotcha. I'll defer to you then. Either just assert as mentioned above or clip the upper bound to be n_episodes (and if the latter, maybe print out a warning to the user). Feel free to resolve this convo after doing this

nadunRanawaka1: Good point about the warning. Added warning before clipping.

PR #1958, opened 2026-02-05
Prevent joylo forcing numpy upgrade

svdeepak99:
Joylo is forcing numpy>2 upgrade, which is not compatible with isaacsim-core 4.5.0.0 & omnigibson 3.7.2.

So when I do a full install with the following command (on my Ubuntu 22.04.5 PC with miniforge3-25.11):
`./setup.sh --new-env --omnigibson --bddl --joylo --dataset --eval --primitives`

I get the following error:
_ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts._

To fix this error, I have added "numpy<2" constrain for joylo, which made setup.sh finish the installation in one go, without errors.

If required, feel free to make modification to the installation setup anywhere, to accomplish the same goal too.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -371,7 +371,10 @@ fi
 if [ "$JOYLO" = true ]; then
     echo "Installing JoyLo..."
     [ ! -d "joylo" ] && { echo "ERROR: joylo directory not found"; exit 1; }
-    pip install -e "$WORKDIR/joylo"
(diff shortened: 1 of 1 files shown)


wensi-ai:
Thanks for the PR, we have updated this in another PR (#1952 ), and will also update this in isaac 5.1 #1955 . Closing this now.

PR #1973, opened 2026-02-11
Convert visual and collision geom prims to plain geom prim

cgokmen:
This pull request converts specialized `VisualGeomPrim`, `CollisionGeomPrim`, and `CollisionVisualGeomPrim` classes to use a unified `GeomPrim` class. This architectural change is necessary for supporting instanceable assets, where collision and visual properties are no longer distinguished at the class level but rather at the link level based on the scope prim structure.

**Changes:**
- Removed `VisualGeomPrim`, `CollisionGeomPrim`, and `CollisionVisualGeomPrim` classes and unified them into `GeomPrim`
- Added `setup_collision_apis()` function in `usd_utils.py` to dynamically set up collision-related APIs on geom prims
- Updated collision vs visual distinction logic to check if geom prims appear under a "collisions" scope prim at the link level

#543

### DIFF
--- a/OmniGibson/omnigibson/prims/geom_prim.py
+++ b/OmniGibson/omnigibson/prims/geom_prim.py
@@ -27,7 +27,7 @@ class GeomPrim(XFormPrim):
     created from scratch.
 
     Geom prims are not inherently distinguished as collision or visual. Instead, at the link level
-    (RigidPrim), they are tracked separately based on whether they appear under a collision scope prim.
+    (RigidPrim), they are tracked separately based on whether they appear with a CollisionAPI or as a child of one.
     Collision-related APIs and methods live on RigidPrim and operate on all collision meshes of a link.
 
     Args:


OmniGibson/omnigibson/scenes/static_traversable_scene.py
```diff
@@ -83,11 +83,12 @@ def _load(self):
         )
 
         # Grab the actual mesh prim
-        self._scene_mesh = CollisionVisualGeomPrim(
+        self._scene_mesh = GeomPrim(
             relative_prim_path=f"/scene/mesh_z_up/{self.scene_model}_mesh_texture",
             name=f"{self.scene_model}_mesh",
         )
         self._scene_mesh.load(self)
+        setup_collision_apis(self._scene_mesh.prim)
```
wensi-ai, line 91: why are we doing this?

wensi-ai (approved):
lgtm

OmniGibson/omnigibson/scenes/static_traversable_scene.py
cgokmen: We don't know if the scene mesh is just a plain USD containing only a mesh, or does it actually have the physics/collision properties attached, so we have to do this.

OmniGibson/omnigibson/prims/geom_prim.py
```diff
@@ -27,7 +24,11 @@ class GeomPrim(XFormPrim):
     """
     Provides high level functions to deal with a geom prim and its attributes / properties.
     If there is an geom prim present at the path, it will use it. By default, a geom prim cannot be directly
-    created from scratch.at
+    created from scratch.
+
+    Geom prims are not inherently distinguished as collision or visual. Instead, at the link level
+    (RigidPrim), they are tracked separately based on whether they appear under a collision scope prim.
```
cremebrule: High level question: is this backwards-compatible? eg.: what if collision geoms don't live under the "collision" scope prim explicitly?

OmniGibson/omnigibson/prims/rigid_prim.py
```diff
@@ -162,48 +169,92 @@ def remove(self):
...
+        # the link-level lists as they are encountered during descent.
+        geom_prims = []
+
+        def _find_geom_prims(prim, is_collision=False):
+            # If this prim has collision APIs, add them to the link-level lists
+            if prim.HasAPI(lazy.pxr.UsdPhysics.CollisionAPI):
+                self._collision_apis.append(lazy.pxr.UsdPhysics.CollisionAPI(prim))
+                is_collision = True
+            if prim.HasAPI(lazy.pxr.PhysxSchema.PhysxCollisionAPI):
+                self._physx_collision_apis.append(lazy.pxr.PhysxSchema.PhysxCollisionAPI(prim))
```
cremebrule, line 194: Ah I see. So we assume the prims already have their appropriate collision APIs, and infer whether they're collision / visual geoms that way, correct?

OmniGibson/omnigibson/prims/rigid_prim.py
```diff
@@ -162,48 +169,92 @@ def remove(self):
...
+                is_collision = True
+
+            if prim.GetPrimTypeInfo().GetTypeName() in GEOM_TYPES:
+                # MeshCollisionAPI is inherently per-geom-prim; store alongside prim reference
+                if prim.HasAPI(lazy.pxr.UsdPhysics.MeshCollisionAPI):
+                    self._mesh_collision_apis.append((prim, lazy.pxr.UsdPhysics.MeshCollisionAPI(prim)))
```
cremebrule, line 200: Is this mainly done for performance reasons? or what's the reason for tracking the collision APIs separately from the mesh prim instances themselves?

OmniGibson/omnigibson/prims/rigid_prim.py
cgokmen: Yes, correct.

OmniGibson/omnigibson/prims/rigid_prim.py
cgokmen: In the future version of this setup, the collision APIs are not attached to the geom prims but instead to the parent xform prim, thats the main reason. So they're N:N currently but it'll be 1:N later.

OmniGibson/omnigibson/prims/geom_prim.py
cgokmen: Comment is misguiding - fixed.

Issue #338, opened 2023-12-11
Standardize and sanitycheck prim path splitting logic


cgokmen:
This is done in a number of places ad-hoc right now - we should unify and we should also make it assert the pattern

PR #1977, opened 2026-02-12
Persistent-cached, step-updated, robust-to-sleep RigidContactAPI for contact checking

cgokmen:
This PR overhauls the contact-checking infrastructure in OmniGibson, centered around a persistent-cached, step-updated `RigidContactAPI` that is robust to sleeping rigid bodies. It also includes several related cleanups and refactors.

1. **RigidContactAPI rewrite** Contact data is now persistently cached per-physics-step and only refreshed for body pairs whose positions have changed, fixing the long-standing issue where sleeping rigid bodies would lose contact information. The old impulse-matrix indexing API is replaced with clean high-level methods (`is_in_contact`, `get_contact_pairs`, `is_in_contact_batch`). The separate `GripperRigidContactAPI` is removed in favor of the unified API.

2. **Contact state cleanup** Removes `ContactBodies`, `ContactSubscribedStateMixin`, and the simulator's `_on_contact` callback. We also remove `contact_list` (which no longer returns correct info) and reimplement this just for cloth objects in a limited manner as `get_contacts`. All object states (`AttachedTo`, `Touching`, `SlicerActive`, `ToggledOn`, etc.) and transition rules now query `RigidContactAPI` directly.

3. **Contact APIs cleanup** Removes all the utility functions in `sim_utils.py` which relied on stepping and contact checking for sampling robots etc. for navigation tasks. These are not primary use cases of OmniGibson and have been simplified to just sample at a known position from the traversability map.

4. **Profiling infrastructure** Adds `cProfile`-based profiling wrapped around simulator step methods, controllable via `gm.ENABLE_PROFILING`. This was used to develop this feature and I decided to merge it since it's useful for future PRs.

5. **Physics step exception handling**  Pre/post physics step callbacks now properly surface exceptions instead of silently swallowing them (previously, any exceptions raised in the callbacks were silently ignored by PhysX). This was obfuscating errors in this PR and can generally be problematic.

### DIFF
--- a/OmniGibson/omnigibson/simulator.py
+++ b/OmniGibson/omnigibson/simulator.py
@@ -430,9 +430,9 @@ def _launch_simulator(*args, **kwargs):
             self.pre_step_exception = None
             self.post_step_exception = None
 
-            self._pre_physics_step_profiler = None
-            self._post_physics_step_profiler = None
-            self._non_physics_step_profiler = None
+            self._pre_physics_step_profiler = Profile() if gm.ENABLE_PROFILING else None
+            self._post_physics_step_profiler = Profile() if gm.ENABLE_PROFILING else None
+            self._non_physics_step_profiler = Profile() if gm.ENABLE_PROFILING else None
 
             self._floor_plane = None
             self._skybox = None
--- a/OmniGibson/tests/benchmark/profiling.py
+++ b/OmniGibson/tests/benchmark/profiling.py
@@ -216,10 +216,10 @@ def main():
         json.dump(ret, f, indent=4)
 
     # Save the simulation profilers
-    # if args.deep_profiling:
-    #     og.sim._pre_physics_step_profiler.dump_stats("pre_physics_step.prof")
-    #     og.sim._post_physics_step_profiler.dump_stats("post_physics_step.prof")
-    #     og.sim._non_physics_step_profiler.dump_stats("non_physics_step.prof")
+    if args.deep_profiling:
+        og.sim._pre_physics_step_profiler.dump_stats("pre_physics_step.prof")
+        og.sim._post_physics_step_profiler.dump_stats("post_physics_step.prof")
+        og.sim._non_physics_step_profiler.dump_stats("non_physics_step.prof")
 
     og.shutdown()
 


OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -99,17 +99,13 @@ def global_update(cls):
...
-                for idx in touching_bodies_idxs:
-                    body_prim_path = RigidContactAPI.get_row_idx_prim_path(scene_idx, idx=idx)
-                    obj = scene.object_registry("prim_path", "/".join(body_prim_path.split("/")[:-1]))
-                    if obj is not None:
-                        cls._finger_contact_objs.add(obj)
+
+            # Get the robot finger prim paths' contacts
+            contact_pairs = RigidContactAPI.get_contact_pairs(scene_idx, scene_robot_finger_paths)
+            for contact_pair in contact_pairs:
+                obj = scene.object_registry("prim_path", "/".join(contact_pair[1].split("/")[:-1]))
```
stefren, line 106: do we have utilities for this type of manipulation on the prim path?

OmniGibson/omnigibson/simulator.py
```diff
@@ -1226,67 +1250,58 @@ def step_physics(self):
...
+                        for robot in scene.robots:
+                            robot.step()
 
-                # Flush the controls from the ControllableObjectViewAPI
-                ControllableObjectViewAPI.flush_control()
+                    # Flush the controls from the ControllableObjectViewAPI
+                    ControllableObjectViewAPI.flush_control()
+            except Exception as e:
+                self.pre_step_exception = e
+                raise e
```
stefren: hm, does the later reporting for exceptions work if we raise e here?

OmniGibson/omnigibson/utils/bddl_utils.py
```diff
@@ -1200,7 +1200,7 @@ def _filter_object_scope(self, input_object_scope, conditions, condition_type):
                         # If we're sampling kinematics, sort children based on (a) whether they are cloth or not, and
                         # then (b) their AABB, so that first all rigid objects are sampled before all cloth objects,
                         # and within each group the larger objects are sampled first. This is needed because rigid
-                        # objects currently don't detect collisions with cloth objects (rigid_obj.states[ContactBodies]
+                        # objects currently don't detect collisions with cloth objects (rigid-object contact checks
```
stefren, line 1203: wait is this still true? I thought we are calculating contacts with cloth

OmniGibson/omnigibson/simulator.py
cgokmen: This raising gets printed and then ignored by PhysX which is the caller of this callback function. I guess there's not much value in raising it here since it's guaranteed to get ignored, but here the actual traceback might be better.

OmniGibson/omnigibson/utils/bddl_utils.py
cgokmen: The rigid body checkers don't check for cloth. The cloth checkers do check for rigid bodies. So it's asymmetrical. This is not changed in this PR.

OmniGibson/omnigibson/object_states/toggle.py
cgokmen: Agreed that we should (and I have opened tickets about this like #338 a long time ago), but actually for various reasons the better approach seems to be structuring all code to avoid all these cases of needing to lookup prim path -> object. Indeed this particular case will soon be removed next week (we'll look up with both column and row indices instead starting with an upcoming PR from @Andorexad )

PR #1982, opened 2026-02-26
Refactor assisted grasping and state restoration flow

cgokmen:
## Summary
- refactor robot assisted-grasp internals around explicit target object/link names, including joint creation helpers and grasp decision flow
- simplify assisted-grasp contact and release handling, remove cloth-specific AG pathways, and drop obsolete AG cloth macro / entity attachment-point logic
- update AG state dump/load serialization, including legacy state conversion support and symbolic semantic grasp callsite updates to the new grasp API


yinhang-oai:
Very tangentially related, we never serialize AG state unfortunately. We used to have a comment in `manipulation_robot` that says something like "TODO: serialize AG", but even that is gone after the robot class refactoring.
https://github.com/StanfordVL/BEHAVIOR-1K/blob/68acabd0a01d5bc096cd3e5b564b06a2eca4531b/OmniGibson/omnigibson/robots/robot.py#L1328

PR #1990, opened 2026-03-06
Controller refactor for batched compute

Andorexad:
## Summary

- Added `ControllerView`-based controller grouping so controllers with the same key share one controller group.
- Grouping key is based on robot kinematic tree pattern + body part + controller config.
- Batching is now executed per group, so control compute/write paths run once per group instead of once per robot. 
- Update goal is not batched yet.

## Controller Batching

- Controller groups maintain per-member state (goals, enabled mask, control outputs, etc.) and shared state (control_limit, mode, etc).
- `ControllerView.step_all()` iterates groups and calls `controller.step()` once per group.
- Inside each controller step:
  - Build/refresh goals for active members
  - Run batched `compute_control(...)` over group members
  - Clip outputs
  - Write controls through batched `ControllableObjectViewAPI` setters
- This preserves member-level behavior while reducing per-robot Python overhead.

## ControllableObjectViewAPI / BatchControlViewAPIImpl Updates

- Added batched write APIs for position/velocity/effort targets over selected rows.
- Added batched read helpers (including new all-* transform helpers) to support group-level compute paths.

## Tests

- Expanded controller regression coverage for grouped execution and reload behavior.

### DIFF
--- a/OmniGibson/omnigibson/controllers/multi_finger_gripper_controller.py
+++ b/OmniGibson/omnigibson/controllers/multi_finger_gripper_controller.py
@@ -247,8 +247,8 @@ class MultiFingerGripperController(GripperController):
             else:
                 u = target_batch  # (N, ctrl_dim)
 
-        # If we're near the joint limits and we're using velocity / torque control, we zero out the action
-        if self._motor_type in {"velocity", "torque"}:
+        # If we're near the joint limits and we're using velocity / effort control, we zero out the action
+        if self._motor_type in {"velocity", "effort"}:
             pos_hi = self._control_limits[ControlType.POSITION][1][self.dof_idx]  # (ctrl_dim,)
             pos_lo = self._control_limits[ControlType.POSITION][0][self.dof_idx]  # (ctrl_dim,)
             violate_upper_limit = all_joint_pos > pos_hi - self._limit_tolerance  # (N, ctrl_dim)
@@ -297,8 +297,8 @@ class MultiFingerGripperController(GripperController):
                 # For joint position control, if the desired positions are the same as the current positions, is_grasping unknown
                 if self._motor_type == "position":
                     no_move_mask = cb.mean(cb.abs(control - joint_pos), dim=1) < m.POS_TOLERANCE  # (N,)
-                # For joint velocity / torque control, if the desired velocities / torques are zeros, is_grasping unknown
-                elif self._motor_type in {"velocity", "torque"}:
+                # For joint velocity / effort control, if the desired velocities / efforts are zeros, is_grasping unknown
+                elif self._motor_type in {"velocity", "effort"}:
                     no_move_mask = cb.mean(cb.abs(control), dim=1) < m.VEL_TOLERANCE  # (N,)
                 else:
                     no_move_mask = cb.bool_zeros(self.n_members)  # all-False
--- a/OmniGibson/omnigibson/utils/backend_utils.py
+++ b/OmniGibson/omnigibson/utils/backend_utils.py
@@ -127,7 +127,7 @@ class _ComputeTorchBackend(_ComputeBackend):
     where = th.where
     squeeze = lambda arr, dim=None: arr.squeeze(dim=dim)
     T = TT
-    indices_where = lambda mask: th.nonzero(mask)[0].tolist()
+    indices_where = lambda mask: th.nonzero(mask, as_tuple=False).squeeze(-1)
     item_bool = lambda x: bool(x.item())
     item_int = lambda x: int(x.item())
     item_float = lambda x: float(x.item())


OmniGibson/omnigibson/robots/robot.py
```diff
@@ -792,126 +828,42 @@ def control_enabled(self):
     @control_enabled.setter
     def control_enabled(self, value):
         self._control_enabled = value
+        if hasattr(self, "_controllers"):
```
wensi-ai: I think this is always true?

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -792,126 +828,42 @@ def control_enabled(self):
...
-            u_type_vec[idx] = ctrl["type"]
-
-        u_vec, u_type_vec = self._postprocess_control(control=u_vec, control_type=u_type_vec)
-
-        # Deploy control signals
-        self.deploy_control(control=u_vec, control_type=u_type_vec)
-
-    def _postprocess_control(self, control, control_type):
+        1. If control is disabled or the articulation view is not ready, skip.
+        2. Override frozen gripper joint targets in the buffer (assisted grasping freeze).
```
wensi-ai: iirc we already removed this

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,276 @@ def _reverse_preprocess_command(self, processed_command):
...
+        # mirroring the old behavior where self._goal was only set by update_goal().
+        for i in range(N):
+            # Skip control_disabled or unregistered members
+            if self._control_enabled[i] == 0 or self._unregistered_controllers[i] == 1:
+                continue
+            if not self._goal_set[i]:
+                no_op = self.compute_no_op_goal(i)
+                for k, v in no_op.items():
+                    self._goals[k][i] = v
+                self._goal_set[i] = True
```
wensi-ai, line 458: this seems to contradict line 470?

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,263 @@
...
+        Register a controller into the appropriate group.
+
+        If no group exists for the given key, one is created. A new member is then
+        added to that group's controller via add_member().
+
+        Args:
+            body_part (str): name of the body part being controlled (e.g., "arm_right", "base")
+            controller_cfg (dict): controller configuration dict (must include "name" key)
+            articulation_root_path (str): articulation root prim path of the new group member
+            link_name (None or str): if specified, the name of the EEF or trunk link (for IK/OSC controllers)
```
wensi-ai, line 61: nit: missing control_enabled

OmniGibson/omnigibson/controllers/ik_controller.py
```diff
@@ -129,11 +126,8 @@ def limiter(target_pos: Array[float], target_quat: Array[float], control_dict: D
         """
         # Store arguments
         control_dim = len(dof_idx)
-        self.control_filter = (
-            None
-            if smoothing_filter_size in {None, 0}
-            else MovingAverageFilter(obs_dim=control_dim, filter_width=smoothing_filter_size)
-        )
+        self._smoothing_filter_size = smoothing_filter_size
```
wensi-ai: FYI we are moving this to JointController in Isaac 5.1. See #1992 . @cgokmen do you think I should separate out that change to a separate PR and merge it first to avoid future merge conflicts?

OmniGibson/omnigibson/controllers/osc_controller.py
```diff
@@ -371,107 +386,158 @@ def _update_goal(self, command, control_dict):
...
-            ee_pos=cb.as_float32(ee_pos),
-            ee_mat=cb.as_float32(cb.T.quat2mat(ee_quat)),
-            ee_lin_vel=cb.as_float32(ee_vel[:3]),
-            ee_ang_vel_err=cb.as_float32(
-                cb.T.quat2axisangle(
-                    cb.T.quat_multiply(cb.T.axisangle2quat(-ee_vel[3:]), cb.T.axisangle2quat(base_ang_vel))
+        kd = 2 * cb.sqrt(kp) * self.damping_ratio
+
+        # Batched joint state reads
+        # Keep all_q as full (N_view, n_dof) to preserve shape for downstream offset computation.
```
wensi-ai: just to confirm, n_dof should inherently be the same within the same controller group?

OmniGibson/tests/test_controllers.py
```diff
@@ -317,3 +388,199 @@ def check_ori_error(curr_orientation, init_orientation, tol=0.1):
...
+
+    assert gk_fetch_a == gk_fetch_b
+    assert gk_fetch_a != gk_franka
+
+    ctrl_fetch = ControllerView.get_controller(gk_fetch_a)
+    ctrl_franka = ControllerView.get_controller(gk_franka)
+    assert ctrl_fetch.n_members >= 2
+    assert ctrl_franka.n_members == 1
+
+    og.clear()
```
wensi-ai, line 587: nice!

OmniGibson/omnigibson/controllers/osc_controller.py
Andorexad: yes

OmniGibson/omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -182,14 +183,14 @@ def __init__(
             for arm_name in self.robot.arm_names:
                 eef = f"eef_{arm_name}"
                 arm = f"arm_{arm_name}"
-                arm_ctrl = self.robot.controllers[arm]
-                if isinstance(arm_ctrl, InverseKinematicsController):
+                arm_gk, arm_ci = self.robot.controllers[arm]
```
cgokmen: General comment: there's no reason to shorten variable names to the point of unintelligibility anymore since everyone uses IDEs with autocomplete. Let's make an effort to find a middle ground here (common words like index can be shortened to idx, but should go with controller_idx instead of ci or group_key or w/e.

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,264 @@
...
+        - Locate the shared controller group.
+        - Mark controller_idx as a tombstoned (unregistered) slot.
+        - Keep the group if any active members remain; delete it only when all members
+        are tombstoned.
+
+        Why tombstones:
+        - Preserve stable controller_idx assignments for remaining robots.
+        - Avoid shifting arrays / remapping indices across robots in the same group.
+
+        Tombstoned slots are never reused.
```
cgokmen: Why not? What's the downside of using a tombstoned slot?

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,264 @@
...
+        Args:
+            group_key (str): key identifying the controller group
+            controller_idx (int): index of the controller within the group
+        """
+        cls._controller_groups[group_key].reset(controller_idx)
+
+    @classmethod
+    def clear(cls):
+        """
+        Remove all registered controller groups. Call this when the scene is reset.
```
cgokmen: When the scene is reset? I don't think so - maybe when the simulator is cleared or something?

OmniGibson/omnigibson/controllers/dd_controller.py
```diff
@@ -88,41 +88,40 @@ def __init__(
...
 
         Returns:
-            Array[float]: outputted (non-clipped!) velocity control signal to deploy
-                to the [left, right] wheel joints
+            Tensor: (N, 2) outputted (non-clipped!) velocity control signal to deploy to the [left, right] wheel joints
         """
-        lin_vel, ang_vel = goal_dict["vel"]
+        vel_batch = goals["vel"]  # (N, 2)
+        lin_vel = vel_batch[:, 0:1]
+        ang_vel = vel_batch[:, 1:2]
```
cgokmen: Why slice this if you're going to squeeze it later, instead of indexing it? Am I misunderstanding

OmniGibson/omnigibson/controllers/ik_controller.py
cgokmen: As long as we merge this PR first I think the work you'll have to do is the same?

OmniGibson/omnigibson/controllers/ik_controller.py
```diff
@@ -143,12 +137,12 @@ def limiter(target_pos: Array[float], target_quat: Array[float], control_dict: D
 
         self.mode = mode
         self.workspace_pose_limiter = workspace_pose_limiter
-        self.task_name = task_name
         self.reset_joint_pos = reset_joint_pos[dof_idx]
         self.condition_on_current_position = condition_on_current_position
 
-        # Other variables that will be filled in at runtime
-        self._fixed_quat_target = None
+        self._link_name = None  # eef/trunk link name (same for all members in the group)
```
cgokmen: This is actually known at the time this controller is created. I don't like that we wait to get it and lazy-save it. Can it be included at creation time? Through the config or the group key?

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @classmethod
+    def is_grasping(cls, group_key: str, controller_idx: int):
+        return cls._controller_groups[group_key].is_grasping(controller_idx)
+
+    @classmethod
+    def dump_state(cls, group_key: str) -> dict:
+        return cls._controller_groups[group_key].dump_state()
+
+    @classmethod
```
cgokmen, line 282: These functions should dump the state of just one controller idx (e.g. what used to be all the data that existed in one instance of the pre-refactor Controller class).

OmniGibson/omnigibson/controllers/ik_controller.py
```diff
@@ -195,42 +189,72 @@ def limiter(target_pos: Array[float], target_quat: Array[float], control_dict: D
...
+            int: controller_idx
+        """
+        idx = super().add_member(articulation_root_path, control_enabled=control_enabled)
+        if self._link_name is None:
+            self._link_name = link_name
+        self._fixed_quat_targets.append(None)
+        self._control_filters.append(
+            None
+            if self._smoothing_filter_size in {None, 0}
+            else MovingAverageFilter(obs_dim=self._filter_obs_dim, filter_width=self._smoothing_filter_size)
```
cgokmen: Switch this to have one instance of the filter per controller, the filter ALSO tracks multiple indices, uses add_member, tombstone, etc.

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -1111,17 +1145,84 @@ def get_relative_angular_velocity(self, prim_path, estimate=False):
...
+
+    def get_all_link_relative_angular_velocity(self, link_name, estimate=False):
+        """Returns (N, 3) link angular velocities for all robots."""
+        cache_key = f"all_link_rel_ang_vel{'_est' if estimate else ''}_{link_name}"
+        if cache_key not in self._read_cache:
+            link_idx = self._link_idx[0][link_name]
+            prim_paths_sorted = sorted(self._idx.keys(), key=lambda p: self._idx[p])
+            self._read_cache[cache_key] = cb.stack(
+                [self._get_relative_velocities(p, estimate=estimate)[link_idx, 3:] for p in prim_paths_sorted], dim=0
+            )
```
cgokmen: These can and should be batched also. Link name is same for everyone.

OmniGibson/omnigibson/controllers/ik_controller.py
cgokmen: And we'll need to move this into JointController (along with the dump/load logic related to it.)

OmniGibson/omnigibson/controllers/ik_controller.py
```diff
@@ -290,74 +326,106 @@ def _update_goal(self, command, control_dict):
...
+            self.routing_path, link_name
+        )  # (N_view, 3), (N_view, 4)
+        ee_pos_all = ee_pos_all[rows]
+        ee_quat_all = ee_quat_all[rows]
+        ee_mat_all = cb.as_float32(cb.T.quat2mat(ee_quat_all))  # (N, 3, 3)
+
+        q_lower = self._control_limits[ControlType.get_type("position")][0][self.dof_idx]  # (ctrl_dim,)
+        q_upper = self._control_limits[ControlType.get_type("position")][1][self.dof_idx]  # (ctrl_dim,)
+        q_lower_batch = cb.stack([q_lower] * N, dim=0)  # (N, ctrl_dim)
+        q_upper_batch = cb.stack([q_upper] * N, dim=0)  # (N, ctrl_dim)
```
cgokmen: Just use q_lower.unsqueeze()

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -197,6 +217,43 @@ def __init__(
...
+
+        Args:
+            articulation_root_path (str): articulation root prim path of the new group member
+            link_name (None or str): EEF link name — ignored by default; overridden by IK/OSC.
+
+        Returns:
+            int: controller_idx — index into this group's member arrays for the added controller,
+                used to access goal, control, and articulation_root_path
+        """
+        controller_idx = len(self._articulation_root_paths)
```
cgokmen: Here, update the logic to reuse the first-available unregistered controller. If none is available, only then, grow the tensors, and use the last index.

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -135,9 +141,23 @@ def __init__(
...
+        # Batched goals: key -> (N, *shape) tensor, built lazily as members are added
+        self._goals = {}
+        # Per-member flag indicating whether goal has been set this step
+        self._goal_set = []
+        # Per-member control enabled mask (1 enabled, 0 disabled)
+        self._control_enabled = []
+        # Per-member last deployed control tensor (set in _write_control)
+        self._controls = []
+        # Per-member tombstone mask: 0 = active, 1 = unregistered (permanently ignored)
+        self._unregistered_controllers = []
```
cgokmen: These should become tensors.

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -135,9 +141,23 @@ def __init__(
...
+        # Lazy-init: row indices of each member in the shared view (None until first _write_control call)
+        self._view_row_indices = None
```
cgokmen: Remove this. Compute on demand with a @property def view_row_indices?

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
+        clipped_control = control.clip(lo, hi)
         idx = (
             self._dof_has_limits[self.dof_idx]
             if self.control_type == ControlType.POSITION
-            else [True] * self.control_dim
+            else (cb.ones(self.control_dim) > 0)
         )
-        control[idx] = clipped_control[idx]
+        control[:, idx] = clipped_control[:, idx]
         return control
```
cgokmen:
```
# Undo the clipping of unlimited position joints
if self.control_type == POSITION:
  idx = self._dof_has_limits[self.dof_idx]
  clipped_control[:, ~idx] = control[:, ~idx]

return clipped_control
```

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
-        Take a controller step.
-
-        Args:
-            control_dict (Dict[str, Any]): dictionary that should include any relevant keyword-mapped
-                states necessary for controller computation
+        Take a batched controller step across all member controller.
 
-        Returns:
-            Array[float]: numpy array of outputted control signals
+        For any controller that has not received a goal this step, a no-op goal is computed.
```
cgokmen: received a goal yet

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
+        The control is then computed for all controllers, clipped, written to Isaac, and the
+        goal_set flags are reset.
         """
-        # Generate no-op goal if not specified
-        if self._goal is None:
-            self._goal = self.compute_no_op_goal(control_dict=control_dict)
+        N = self.n_members
+        # If all active (non-unregistered) members are disabled, early return
+        if all(self._control_enabled[i] == 0 or self._unregistered_controllers[i] == 1 for i in range(N)):
+            return
```
cgokmen: Tensorize these.

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
-        return self._control
+        # Fill in no-op goals for any controllers that haven't received a goal.
+        for i in range(N):
+            # Skip control_disabled or unregistered members
+            if self._control_enabled[i] == 0 or self._unregistered_controllers[i] == 1:
+                continue
+            if not self._goal_set[i]:
+                no_op = self.compute_no_op_goal(i)
+                for k, v in no_op.items():
+                    self._goals[k][i] = v
```
cgokmen: First, find the indices of non-initialized controllers using cb.argwhere(?) then just loop over those indices

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
+        control_output = self.clip_control(control_output)
+        for i in range(N):
+            if self._control_enabled[i] != 0 and self._unregistered_controllers[i] == 0:
+                ctrl = control_output[i]
+                self._controls[i] = (
+                    cb.from_torch(ctrl)
+                    if isinstance(ctrl, th.Tensor)
+                    else (ctrl if isinstance(ctrl, cb.arr_type) else cb.array(ctrl))
+                )
+        self._write_control()
```
cgokmen: Just put the contents of this here.

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
+        unregistered = state.get("unregistered_controllers", [0] * self.n_members)
+        for i, u in enumerate(unregistered):
+            self._unregistered_controllers[i] = int(u)
+        for name, val in state["goals"].items():
+            if name in self._goals:
+                if isinstance(val, th.Tensor):
+                    val = cb.from_torch(val)
+                elif not isinstance(val, cb.arr_type):
+                    val = cb.array(val)
+                self._goals[name][:] = val
```
cgokmen: Only save/load one index at once. The caller should pass in a separate controller_idx.

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +346,278 @@ def _reverse_preprocess_command(self, processed_command):
...
```
cgokmen: Remove the code above that does the weird override of the goal setting.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -310,7 +310,7 @@ def __init__(
         self.dof_names_ordered = None
         self._control_enabled = True
 
-        class_name = self.kinematic_tree_identifier.lower()
+        class_name = self.robot_type.lower()
         if relative_prim_path:
```
cgokmen: Let's remove this option. Make sure it's not used anywhere

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,264 @@
...
+                - str: group_key identifying the controller group
+                - int: controller_idx — the member's index within that group
+        """
+        # Build a unique string key for a controller group.
+        group_key = cls._make_key(articulation_root_path, body_part, controller_cfg)
+
+        if group_key not in cls._controller_groups:
+            from omnigibson.controllers import create_controller
+
+            cls._controller_groups[group_key] = create_controller(**controller_cfg)
```
cgokmen: Just add link_name= here.

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,264 @@
...
+
+        controller = cls._controller_groups[group_key]
+        controller_idx = controller.add_member(
+            articulation_root_path,
+            link_name,
```
cgokmen: Remove link_name here.

OmniGibson/omnigibson/controllers/multi_finger_gripper_controller.py
```diff
@@ -179,91 +181,103 @@ def _preprocess_command(self, command):
...
+        target_batch = goals["target"]  # (N, command_dim)
+
+        self._ensure_view_row_indices()
+        rows = self._view_row_indices
+        all_joint_pos = ControllableObjectViewAPI.get_all_joint_positions(self.routing_path)[rows, :][
+            :, self.dof_idx
+        ]  # (N, ctrl_dim)
+
+        u_list = []
+        for i in range(N):
```
cgokmen: Tensorize this.

OmniGibson/omnigibson/controllers/multi_finger_gripper_controller.py
```diff
@@ -179,91 +181,103 @@ def _preprocess_command(self, command):
...
-        # Update whether we're grasping or not
-        self._update_grasping_state(control_dict=control_dict)
+            u_list.append(u)
 
-        # Return control
-        return u
+        return cb.stack(u_list, dim=0)  # (N, control_dim)
 
-    def _update_grasping_state(self, control_dict):
+    def _update_grasping_state(self, controller_idx, joint_pos, control):
```
cgokmen: Tensorize this too

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -598,14 +601,37 @@ def _load_controllers(self):
...
-            controller = create_controller(**cb.from_torch_recursive(cfg))
+            # Determine link_name for arm or trunk controllers (used by IK/OSC)
+            link_name = None
+            if self.is_manipulation:
+                for arm in self.arm_names:
+                    if name == f"arm_{arm}":
+                        try:
+                            link_name = self.eef_link_names[arm]
+                        except (ValueError, AttributeError, KeyError):
+                            pass
```
cgokmen: Don't do try/except

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -598,14 +601,37 @@ def _load_controllers(self):
...
+            if self.is_articulated_trunk:
+                if name == "trunk":
+                    try:
+                        link_name = self.joints[self.trunk_joint_names[-1]].body1.split("/")[-1]
+                    except (ValueError, AttributeError, KeyError):
+                        pass
```
cgokmen: Here too, no try/except. This should exist.

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -1389,7 +1183,7 @@ def _initialize(self):
             robot_name_components[0] == "controllable"
         ), "Third component of articulation root path (robot name) must start with 'controllable'"
         assert (
-            robot_name_components[1] == self.kinematic_tree_identifier.lower()
+            robot_name_components[1] == self.robot_type.lower()
         ), "Third component of articulation root path (robot name) must contain the class name as the second part"
         # Run super
```
cgokmen: Remove this checking mechanism

OmniGibson/omnigibson/utils/backend_utils.py
```diff
@@ -54,6 +54,7 @@ class _ComputeBackend:
     arange = None
     where = None
     squeeze = None
+    stack = None
```
cgokmen: Remove

OmniGibson/omnigibson/controllers/dd_controller.py
```diff
@@ -88,41 +88,40 @@ def __init__(
...
 
-        # Convert to wheel velocities
         left_wheel_joint_vel = (lin_vel - ang_vel * self._wheel_axle_halflength) / self._wheel_radius
         right_wheel_joint_vel = (lin_vel + ang_vel * self._wheel_axle_halflength) / self._wheel_radius
```
cgokmen: Try to compute both of these at once, without the later stack, by making self._wheel_axle_halflength be a [-wah, wah] tensor we compute/stack ahead of time, during initialization

OmniGibson/omnigibson/controllers/osc_controller.py
```diff
@@ -246,11 +243,11 @@ def limiter(target_pos: Array[float], target_quat: Array[float], control_dict: D
         # Other values
         self.decouple_pos_ori = decouple_pos_ori
         self.workspace_pose_limiter = workspace_pose_limiter
-        self.task_name = task_name
         self.reset_joint_pos = reset_joint_pos[dof_idx]
 
-        # Other variables that will be filled in at runtime
-        self._fixed_quat_target = None
+        # member state that will be filled in at runtime
+        self._link_name = None  # eef/trunk link name (same for all members in the group)
```
cgokmen: Dont forget to move this into init and remove from add_member

OmniGibson/omnigibson/controllers/osc_controller.py
```diff
@@ -311,27 +326,23 @@ def _update_variable_gains(self, gains):
...
-        pos_relative = cb.copy(control_dict[f"{self.task_name}_pos_relative"])
-        quat_relative = cb.copy(control_dict[f"{self.task_name}_quat_relative"])
+        prim_path = self._articulation_root_paths[controller_idx]
+        link_name = self._link_name
+
+        # Get current EEF pose
+        pos_relative, quat_relative = ControllableObjectViewAPI.get_link_relative_position_orientation(
+            prim_path, link_name
+        )
+        pos_relative = cb.copy(pos_relative)
```
cgokmen: Dont copy these here, only copy them below if you're storing them directly.

OmniGibson/omnigibson/controllers/osc_controller.py
```diff
@@ -371,107 +386,158 @@ def _update_goal(self, command, control_dict):
...
-            rest_qpos=self.reset_joint_pos,
+            )
+        )  # (N, 3)
+
+        # Broadcast scalar gains to (N, *) for the batch JIT
+        kp_batch = cb.stack([kp] * N, dim=0)  # (N, 6)
+        kd_batch = cb.stack([kd] * N, dim=0)  # (N, 6)
+        kp_null_batch = cb.stack([self.kp_null] * N, dim=0)  # (N, ctrl_dim)
+        kd_null_batch = cb.stack([self.kd_null] * N, dim=0)  # (N, ctrl_dim)
+        rest_qpos_batch = cb.stack([self.reset_joint_pos] * N, dim=0)  # (N, ctrl_dim)
```
cgokmen: Make this work with broadcasting - don't do *N and stack

cgokmen (changes requested):

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,274 @@
...
+        Register a controller into the appropriate group.
+
+        If no group exists for the given key, one is created. A new member is then
+        added to that group's controller via add_member().
+
+        Args:
+            body_part (str): name of the body part being controlled (e.g., "arm_right", "base")
+            controller_cfg (dict): controller configuration dict (must include "name" key)
+            articulation_root_path (str): articulation root prim path of the new group member
+            link_name (None or str): if specified, the name of the EEF or trunk link (for IK/OSC controllers)
```
wensi-ai, line 61: nit: Per @cremebrule, the link can technically be any part of the robot (e.g. hand, trunk, head, etc.). We can probably keep it for now, just FYI

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,274 @@
...
+        Returns:
+            2-tuple:
+                - str: group_key identifying the controller group
+                - int: controller_idx — the member's index within that group
+        """
+        # Build a unique string key for a controller group.
+        group_key = cls._make_key(articulation_root_path, body_part, controller_cfg)
+
+        if group_key not in cls._controller_groups:
+            from omnigibson.controllers import create_controller
```
wensi-ai, line 73: Is there a reason not importing this at the top? I remember the linter would complain about this but if not I guess it's fine.

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,274 @@
...
+
+            controller_name = controller_cfg.get("name")
+            if link_name is not None and controller_name in {
+                "InverseKinematicsController",
+                "OperationalSpaceController",
+            }:
+                cls._controller_groups[group_key] = create_controller(**controller_cfg, link_name=link_name)
+            else:
+                cls._controller_groups[group_key] = create_controller(**controller_cfg)
```
wensi-ai, line 83: For ik/osc, if link_name is None, does it have a default value in controller_cfg? If not maybe we should assert here, if so please ignore this.

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,274 @@
...
+        # Fallback for custom / array-like objects
+        try:
+            hash(value)
+            return value
+        except TypeError:
+            return str(value)
+
+    @classmethod
+    def _make_key(cls, articulation_root_path: str, body_part: str, controller_cfg: dict) -> str:
+        from omnigibson.utils.usd_utils import get_robot_kinematic_tree_pattern
```
wensi-ai, line 317: same here for import

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -135,9 +141,25 @@ def __init__(
...
+        self._articulation_root_paths = []
+        # Batched goals: key -> (N, *shape) tensor, built lazily as members are added
+        self._goals = {}
+        # Per-member flag indicating whether goal has been set this step
+        self._goal_set = th.zeros(0, dtype=th.bool)
+        # Per-member control enabled mask (1 enabled, 0 disabled)
+        self._control_enabled = th.zeros(0, dtype=th.long)
+        # Per-member last deployed control tensor (N, control_dim)
+        self._controls = th.zeros((0, self.control_dim))
+        # Per-member tombstone mask: 0 = active, 1 = unregistered (permanently ignored)
```
wensi-ai: this seems wrong?

Andorexad: yeah the comment is wrong, i forgot to change it. will update in next commit

OmniGibson/omnigibson/controllers/controller_view.py
Andorexad: if move to top there'll be a circular import

OmniGibson/omnigibson/controllers/controller_view.py
Andorexad: same as above

OmniGibson/omnigibson/controllers/controller_view.py
Andorexad: in previous code link_name is called "task_name", and it could only be trunk or arm inside ik/osc. so I didn't consider other possibilities here. do we need to consider other possibilities?

cgokmen:
@copilot review

OmniGibson/omnigibson/controllers/joint_controller.py
```diff
@@ -153,6 +159,95 @@ def __init__(
...
+        state = super()._dump_state(controller_idx=controller_idx)
+        if self._control_filter is None:
+            state["control_filter"] = None
+        else:
+            filter_state = self._control_filter.dump_state(serialized=False)
+            state["control_filter"] = {
+                "past_samples": filter_state["past_samples"][controller_idx].clone(),
+                "current_idx": filter_state["current_idx"][controller_idx].clone(),
+                "fully_filled": filter_state["fully_filled"][controller_idx].clone(),
+            }
```
cgokmen: This breaks abstraction. Pass the idx to the filter too. Here just do `state["control_filter"] = self._control_filter.dump_state(idx=controller_idx)`

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -792,124 +800,15 @@ def control_enabled(self):
...
-            control_type (k- or n-array): control types for each DOF. Each entry should be one of ControlType.
-                 This should be n-DOF length if all joints are being set, or k-length (k < n) if specific
-                 indices are being set. In this case, the length of @control must be the same length as @indices!
-
-        Returns:
-            2-tuple:
-                - n-array: raw control signals to send to the object's joints
-                - list: control types for each joint
+        Run per-robot assisted grasping logic after ControllerView.step_all() writes controls to the Isaac buffer
+        but before ControllableObjectViewAPI.flush_control() flushes them.
```
cgokmen: Question: is it important that this happens before the control gets flushed?

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -1206,19 +970,22 @@ def set_joint_positions(self, positions, indices=None, normalized=False, drive=F
...
-            controller_states[controller_name] = controller.dump_state()
-
-        state["controllers"] = controller_states
+        # Dump controllers
+        controller_states = {}
+        for controller_name, (group_key, controller_idx) in self._controllers.items():
+            print(
+                f"[Robot._dump_state] robot={self.name} "
+                f"controller={controller_name} group_key={group_key} idx={controller_idx}"
+            )
```
cgokmen: Remove debug print. Alternatively just do log.debug() here in case it's useful later (then it can be seen in debug mode only)

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -1240,10 +1007,16 @@ def _load_state(self, state):
...
-            controller.load_state(state=controller_states[controller_name])
+        # Load controllers
+        for controller_name, (group_key, controller_idx) in self._controllers.items():
+            controller_state = state["controller_groups"].get(controller_name)
+            if controller_state is not None:
+                print(
+                    f"[Robot._load_state] robot={self.name} "
+                    f"controller={controller_name} group_key={group_key} idx={controller_idx} "
+                    f"incoming_goal_set={controller_state.get('goal_set', None)}"
+                )
```
cgokmen: More debug print. Alternatively just do log.debug() here in case it's useful later (then it can be seen in debug mode only)

OmniGibson/omnigibson/controllers/controller_view.py
cgokmen: The comment here is inaccurate now, right?

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -4025,25 +3795,27 @@ def q_to_action(self, q):
         """
         if not self.is_holonomic_base:
             action = []
-            for name, controller in self.controllers.items():
+            for name, (group_key, _) in self.controllers.items():
+                ctrl = ControllerView.get_controller(group_key)
```
cgokmen: Is this the only place where we expose the internal controller object outside the view? If so, let's remove this API, and for this line add a get_use_delta_commands API at the view level (and raise an error for anything that's not a jointcontroller)

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -4058,7 +3830,7 @@ def q_to_action(self, q):
                     canonical_pos, th.tensor([0.0, 0.0, 0.0, 1.0]), body_pos, body_quat
                 )[0]
                 command = th.tensor([local_pos[0], local_pos[1], delta_q])
-            action.append(controller._reverse_preprocess_command(command))
+            action.append(ctrl._reverse_preprocess_command(command))
```
cgokmen: Expose this on the view too

OmniGibson/omnigibson/controllers/ik_controller.py
```diff
@@ -193,76 +192,60 @@ def limiter(target_pos: Array[float], target_quat: Array[float], control_dict: D
...
+            self._fixed_quat_targets[idx] = None
+        else:
+            self._fixed_quat_targets.append(None)
+        return idx
 
-        return state
+    def reset(self, controller_idx):
+        # Call super first
+        super().reset(controller_idx)
+        self._fixed_quat_targets[controller_idx] = None
```
cgokmen: Question: is there any reason this is not part of the goals (and such automatically managed by the parent class, and included in the no-op goal, etc.?)

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +362,266 @@ def _reverse_preprocess_command(self, processed_command):
...
 
-        # Compute control, then clip and return
-        control = self.compute_control(goal_dict=self._goal, control_dict=control_dict)
-        assert (
-            len(control) == self.control_dim
-        ), f"Control signal must be of length {self.control_dim}, got {len(control)} instead."
-        self._control = self.clip_control(control=control)
-        return self._control
+    def unregister_member(self, controller_idx):
+        """Mark member at controller_idx as a tombstone (permanently ignored)."""
```
cgokmen: Inaccurate comment

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -1018,13 +1019,61 @@ def set_joint_efforts(self, prim_path, efforts, indices):
         # Add this index to the write cache
         self._write_idx_cache["dof_actuation_forces"].add(idx)
 
-    def get_root_transform(self, prim_path):
+    def get_member_view_indices(self, prim_paths):
+        """Return view row index for each prim_path (in input order)."""
+        return [self._idx[p] for p in prim_paths]
+
+    def set_all_joint_position_targets(self, enabled_rows, controls, dof_idx):
```
cgokmen, line 1298: Why is enabled_rows not already the correctly shaped int array in all these functions? I think we should make it be that and remove the int_array step.

Andorexad: In controller_base line 476, `enabled_rows` is init as a flat 1D tensor.  it's also passed to                           `self._write_idx_cache[...].update(enabled_rows)` in usd_utils.py, which needs a flat iterable of hashable ints. If we want to have a (-1,1) shape passed in, we need to flatten().
The `cb.int_array() + reshape` and `flatten()` are both O(1) view ops.
Is the flatten approach the most efficient way to handle this?

OmniGibson/omnigibson/robots/robot.py
Andorexad: It's not important. This function won't change control data. I just followed the code before refactor. Is it better to move it after `flush_control`?

OmniGibson/omnigibson/controllers/controller_view.py
Andorexad: I added assert

OmniGibson/omnigibson/robots/robot.py
cgokmen: No, its okay. It just sounded important :)

cgokmen (changes requested):
All open comments have been addressed and this looks good in general. Unfortunately - the profiling message shows significant slowdowns (50x FPS reduction = everything's 2x slower). The interesting thing is that the slowed-down part seems to be not the control step but the non-physics part? Let's discuss tomorrow how to debug. I'll be at the lab.

OmniGibson/omnigibson/action_primitives/starter_semantic_action_primitives.py
```diff
@@ -182,14 +183,14 @@ def __init__(
             for arm_name in self.robot.arm_names:
```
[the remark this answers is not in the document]
cgokmen: @Andorexad apply this

OmniGibson/omnigibson/controllers/joint_controller.py
```diff
@@ -199,86 +276,106 @@ def _update_goal(self, command, control_dict):
...
+                ).float()
 
-            # Add Coriolis / centrifugal compensation
             if self._use_cc_compensation:
-                u += control_dict["cc_force"][self.dof_idx]
+                u += cb.to_torch(
+                    ControllableObjectViewAPI.get_all_coriolis_and_centrifugal_compensation_forces(self.routing_path)[
+                        rows, :
+                    ][:, self.dof_idx]
+                ).float()
```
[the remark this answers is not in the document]
cgokmen: @Andorexad confirm and apply this

OmniGibson/omnigibson/controllers/controller_view.py
```diff
@@ -0,0 +1,284 @@
...
+
+        pattern = get_robot_kinematic_tree_pattern(articulation_root_path)
+        cfg_hash = hash(cls._freeze_for_hash(controller_cfg))
+        return f"{pattern}__{body_part}__{cfg_hash}"
```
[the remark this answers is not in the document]
cgokmen: @Andorexad apply this, but in other places in our code I think we use md5 (confirm this) so consider that instead of sha.

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -1018,13 +1019,61 @@ def set_joint_efforts(self, prim_path, efforts, indices):
...
+            enabled_rows: list[int] — view row indices for enabled members (pre-filtered)
+            controls: (N_enabled, len(dof_idx)) array — pre-stacked by controller
+            dof_idx: DOF column indices (cb.arr_type)
+        """
+        if "dof_position_targets" not in self._read_cache:
+            self._read_cache["dof_position_targets"] = cb.from_torch(self._view.get_dof_position_targets())
+        targets = self._read_cache["dof_position_targets"]
+        row_idx = cb.int_array(enabled_rows).reshape(-1, 1)
+        targets[row_idx, dof_idx] = cb.from_torch(controls) if isinstance(controls, th.Tensor) else controls
+        self._write_idx_cache["dof_position_targets"].update(enabled_rows)
```
cgokmen, line 1310: .tolist

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -1018,13 +1019,61 @@ def set_joint_efforts(self, prim_path, efforts, indices):
...
 
-    def get_root_transform(self, prim_path):
+    def get_member_view_indices(self, prim_paths):
+        """Return view row index for each prim_path (in input order)."""
+        return [self._idx[p] for p in prim_paths]
+
+    def set_all_joint_position_targets(self, enabled_rows, controls, dof_idx):
+        """
+        Args:
+            enabled_rows: list[int] — view row indices for enabled members (pre-filtered)
```
cgokmen, line 1301: Change these enabled_rows arg types to tensor

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -1018,13 +1019,61 @@ def set_joint_efforts(self, prim_path, efforts, indices):
...
+            controls: (N_enabled, len(dof_idx)) array — pre-stacked by controller
+            dof_idx: DOF column indices (cb.arr_type)
+        """
+        if "dof_position_targets" not in self._read_cache:
+            self._read_cache["dof_position_targets"] = cb.from_torch(self._view.get_dof_position_targets())
+        targets = self._read_cache["dof_position_targets"]
+        row_idx = cb.int_array(enabled_rows).reshape(-1, 1)
```
cgokmen, line 1308: Don't do int_array

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +362,266 @@ def _reverse_preprocess_command(self, processed_command):
...
+
+        # Clip, store per-member controls, then write batched
+        control_output = self.clip_control(control_output)
+        self._controls[active_mask] = control_output[active_mask]
+
+        # Write batched control signals to Isaac via a single tensor op per control type.
+        if not active_mask.any():
+            return
+
+        all_view_rows = th.tensor(self.view_row_indices, dtype=th.long)
```
cgokmen: maybe this should be stored as a tensor all along?

OmniGibson/omnigibson/controllers/controller_base.py
```diff
@@ -289,180 +362,272 @@ def _reverse_preprocess_command(self, processed_command):
...
+            no_op_goal = self.compute_no_op_goal(controller_idx)
+            for k, v in no_op_goal.items():
+                self._goals[k][controller_idx] = v
+        command = self._compute_no_op_command(controller_idx)
+        # Ensure the no-op command matches the active controller backend before preprocessing.
+        # Some controllers return torch.Tensors even when the controller backend is NumPy,
+        # which would cause type mismatches inside _reverse_preprocess_command (e.g., when
+        # comparing or clamping against NumPy limits). Convert to a NumPy array in that case.
+        if isinstance(command, th.Tensor) and cb is not th:
+            command = command.detach().cpu().numpy()
```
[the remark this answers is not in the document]
cgokmen: This should not be a thing. We should only return, and use, tensors that are in the type of the compute backend.

PR #2000, opened 2026-03-11
Move light object and primitive object to inherit from USDObject

stefren:


OmniGibson/omnigibson/objects/light_object.py
```diff
@@ -99,19 +106,19 @@ def __init__(
...
+        tempdir_path = tempfile.mkdtemp(name, dir=og.tempdir)
+        usd_path = os.path.join(tempdir_path, f"{name}.usd")
+        side_stage = lazy.pxr.Usd.Stage.CreateNew(usd_path)
+        root = side_stage.DefinePrim("/object", "Xform")
+        side_stage.SetDefaultPrim(root)
+        side_stage.DefinePrim("/object/base_link", "Xform")
+        getattr(lazy.pxr.UsdLux, f"{light_type}Light").Define(side_stage, "/object/base_link/light")
+        side_stage.Save()
+        del side_stage
+        return usd_path
```
cgokmen, line 120: Nice. One consideration: this object actually ends up with no visual meshes and no collision meshes. In fact it doesn't even have a rigid body API attached to its base link. This violates a bunch of assumptions. Can you check what the object's `links` looks like after creation? Is it a RigidKinematicPrim? What APIs does it end up with? What puts those APIs on there?

cgokmen: If it is kinematic-only, let's actually enforce it to be that by pushing the kinematic_only=True load config that BaseObject uses for enforcement

OmniGibson/omnigibson/objects/primitive_object.py
```diff
@@ -120,33 +127,44 @@ def __init__(
...
+        col_geom = create_primitive_mesh(
+            prim_path="/object/base_link/collisions", primitive_type=primitive_type, stage=side_stage
         )
+        lazy.pxr.UsdPhysics.CollisionAPI.Apply(col_geom.GetPrim())
+        lazy.pxr.UsdPhysics.MeshCollisionAPI.Apply(col_geom.GetPrim())
+        lazy.pxr.PhysxSchema.PhysxCollisionAPI.Apply(col_geom.GetPrim())
+
+        side_stage.Save()
+        del side_stage
+        return usd_path
```
cgokmen, line 153: This also doesn't apply the rigid body API, and assumes that it will get applied. Is it entity_prim's post_load that applies this by creating a rigid prim? If so, is it a dynamic prim? What's e.g. its mass, etc?

OmniGibson/omnigibson/objects/light_object.py
stefren: Ok, I think I'm still a bit shaky on all the assumptions that are being made.. but I think previously it was becoming a `RigidDynamicPrim` with APIs set by the `rigid_prim` `_post_load` function since `fixed_base` was default False. 
I'm leaning towards making a change that makes `fixed_base=True` and `kinematic_only=True` for light objects, unless there's a case where users actually want to configure it to have a non fixed base? (I don't know if they are allowed to move light objects??)

OmniGibson/omnigibson/objects/primitive_object.py
stefren: yes, I think it becomes a dynamic prim. In `rigid_prim.py` `_post_load` it looks at the `load_config` but i'm not sure if that's set correctly in this case. It seems like we're ignoring it if `mass` is passed in as a keyword argument to `PrimitiveObject`, maybe that's a bug? I'm also still not super familiar what this is used for.

OmniGibson/omnigibson/objects/light_object.py
cgokmen: In this case since the object is just a light and nothing else (e.g. there actually isn't a mesh that a robot can interact with) I agree that this is the best way to go.

PR #2004, opened 2026-03-12
Fix / disable a few tests from the examples as tests

stefren:


OmniGibson/tests/test_examples.py
```diff
@@ -55,6 +52,9 @@
     "teleoperation.robot_teleoperate_demo",
     "teleoperation.vr_robot_control_demo",  # does not support headless mode
     "teleoperation.vr_scene_tour_demo",  # does not support headless mode
+    "robots.curobo_example",  # requires CuRobo and CUDA support
+    "robots.import_custom_robot",  # CLI conversion tool, requires demo / test asset file
+    "objects.import_custom_object",  # CLI conversion tool, requires demo / test asset files
```
cgokmen: Lets not disable these, let them fail. We can add in a test asset file to run this on later.

cgokmen: Actually nvm lets leave them disabled. I'll assign somebody the task of fixing these and we can reenable then.

PR #2009, opened 2026-03-13
Move articulation root determination to preload step

stefren:


OmniGibson/omnigibson/objects/object_base.py
```diff
@@ -250,29 +250,7 @@ def _post_load(self):
...
-        # Potentially add articulation root APIs and also set self collisions
-        root_prim = (
-            None
-            if self.articulation_root_path is None
-            else lazy.isaacsim.core.utils.prims.get_prim_at_path(self.articulation_root_path)
-        )
-        if root_prim is not None:
-            lazy.pxr.UsdPhysics.ArticulationRootAPI.Apply(root_prim)
-            lazy.pxr.PhysxSchema.PhysxArticulationAPI.Apply(root_prim)
+        if self.articulation_root_path is not None:
```
cgokmen, line 249: It's still true that all objects require this articulation root path logic, so you need to merge the UsdObject class into BaseObject first prior to doing this. Otherwise it looks like we support a class of objects (which by the way doesn't actually exist right now) that don't have articulation roots set properly.

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -15,6 +17,186 @@
...
+        usd_path (str): Original USD path; its basename is reused in the temp directory.
+
+    Returns:
+        str: Path to the exported temp file.
+    """
+    basename = os.path.basename(usd_path)
+    tempdir_path = tempfile.mkdtemp(basename, dir=og.tempdir)
+    temp_usd_path = os.path.join(tempdir_path, basename)
+    stage.Export(temp_usd_path)
+    return temp_usd_path
```
cgokmen: I think its better to keep this code in the (single) place it's called to make it clear where it's saving, instead of having this helper function.

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -15,6 +17,186 @@
...
+        children.extend(child_prim.GetChildren())
+        prim_type = child_prim.GetPrimTypeInfo().GetTypeName().lower()
+        if "joint" in prim_type:
+            if "fixed" in prim_type:
+                n_fixed_joints += 1
+            else:
+                n_joints += 1
+        if "attachment" in child_prim.GetName().lower():
+            has_attachment = True
+    return n_joints, n_fixed_joints, has_attachment
```
cgokmen: Lets throw this into USD utils

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -15,6 +17,186 @@
...
+        native_bb_attr = default_prim.GetAttribute("ig:nativeBB")
+        if native_bb_attr.IsValid():
+            native_bb = th.tensor(list(native_bb_attr.Get()), dtype=th.float32)
+            bb = th.tensor(bounding_box, dtype=th.float32)
+            scale = th.ones(3)
+            valid = native_bb > 1e-4
+            scale[valid] = bb[valid] / native_bb[valid]
+            return scale
+
+    return th.ones(3)
```
cgokmen: I'm noticing that this code is not actually replacing any code, so this logic (divide bbox size by nativebb size, etc.) must exist in at least one other place, no? We shouldn't have that kind of duplication. Hopefully that other place is happening before this one and we can just use the scale here?

Also, code at this class's level should not rely on things like the ig:nativeBB flag which is only available for DatasetObject.

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -15,6 +17,186 @@
...
+    if not fixed_base:
+        return False
+    if load_config.get("kinematic_only", None) is False:
+        return False
+    scale = _resolve_scale(load_config, default_prim)
+    return (
+        n_joints == 0
+        and (th.all(th.isclose(scale, th.ones_like(scale), atol=1e-3)).item() or n_fixed_joints == 0)
+        and not has_attachment
+    )
```
cgokmen: Where did the original version of this go? I don't see it removed in object_base, is it repeated? Let's make sure it's not repeated, and the existing forcing mechanism is respected. Also don't make this into a helper function like this, let's just have the code be where it's used.

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -15,6 +17,186 @@
...
+                continue
+            if len(body0_rel.GetTargets()) > 0:
+                joint_children.add(body1_rel.GetTargets()[0].pathString.split("/")[-1])
+
+    valid_roots = list(set(link_names) - joint_children)
+    assert len(valid_roots) == 1, (
+        f"Exactly one root link should have been found for {default_prim.GetName()}, "
+        f"but found none/multiple instead: {valid_roots}"
+    )
+    return valid_roots[0]
```
cgokmen: Lets move this into usd_utils

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -15,6 +17,186 @@
...
+
+    if kinematic_only or (not has_articulated and not has_fixed):
+        return None
+
+    if not fixed_base and has_articulated:
+        root_link_name = _find_root_link_name(default_prim)
+        target = default_prim.GetPrimAtPath(root_link_name)
+        return target
+
+    return default_prim
```
cgokmen: This can also go into usd_utils.

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -127,8 +310,34 @@ def _update_path(asset_path):
...
+        n_joints, n_fixed_joints, has_attachment = _count_joints(default_prim)
+        articulation_root_prim = _determine_articulation_root_prim(
+            default_prim, self.fixed_base, self._load_config, n_joints, n_fixed_joints, has_attachment
+        )
+
+        if articulation_root_prim is not None:
+            lazy.pxr.UsdPhysics.ArticulationRootAPI.Apply(articulation_root_prim)
+            lazy.pxr.PhysxSchema.PhysxArticulationAPI.Apply(articulation_root_prim)
+
+        return _export_stage_to_temp(stage, usd_path)
```
cgokmen: This can all go into _prepare_to_load, I think.

OmniGibson/omnigibson/objects/usd_object.py
stefren: yeah ok, the review makes sense to me haha. I think I was confused about one version of the code being on the USD directly and one version on the already created `EntityPrim` object looking at various instance variables that are pulled out, but I'll unify this and clean it up.

OmniGibson/omnigibson/objects/object_base.py
stefren: oh ok, do you prefer that i leave this code here or should i merge USDObject into BaseObject?

Issue #880, opened 2024-09-13
Remove kinematic only pose cache, use PoseAPI directly


cgokmen:
This can output wrong values if the scene is moved.

PR #2020, opened 2026-03-18
Object State Tensorization 1/2

Andorexad:
Changes in this PR
- Merge global update state into tensorized
- Update state `toggle`, `open`, `aabb` to be tensorized
- Add `ArticulationObjectViewAPI` and `RigidBodyViewAPI`. Both of their computation are on GPU
- Move tensorized object_states value compute step onto GPU
- Move part of `RigidContactViewAPI` computation to GPU
- Remove kinematic only pose cache, use PoseAPI directly #880


OmniGibson/omnigibson/simulator.py
```diff
@@ -67,9 +49,28 @@ def wrapper(self, *args, **kwargs):
     ControllableObjectViewAPI,
     PoseAPI,
     RigidContactAPI,
+    clear as clear_usd_utils,
+    triangularize_mesh,
```
cgokmen, line 55: Why are these added? Why doesn't it cause any issues to not have these?

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,146 @@
...
+    # Named with underscore to avoid collision with the @property requires_closed below.
+    _requires_closed = None
+
+    # list[GeomPrim]: visual toggle-button markers, one per tracked object (same idx as VALUES).
+    # Populated during _initialize(); used in _check_overlap and color updates.
+    visual_markers = None
+
+    # list[float | th.Tensor]: per-object marker scale (same idx as VALUES).
+    # Populated during _initialize(); used in _check_overlap for sphere radius.
+    scales = None
```
cgokmen: Might as well have these start with _ too, these are all private after all.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,146 @@
...
+    """
+    Boolean state representing whether a button-like object has been toggled on.
+
+    Tensorized storage
+    ------------------
+    Inherits TensorizedValueState. All per-object mutable values are stored in class-level
+    tensors and lists (same index as VALUES):
+
+        VALUES[:, 0]     — toggle on/off (float32, 0.0 = off / 1.0 = on)
+        VALUES[:, 1]     — robot_can_toggle_steps counter (float32, used as int)
```
cgokmen: After reading this PR I realized that the cleanest convention for this is "only put it on VALUES if it's going to be included as the return value of get_value()", which this second element fails. So the state size for this particular state should be 1, and the robot_can_toggle_steps variable should be a separate tensor.

That way you can flip the state type to bool, too.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,146 @@
...
+    def value_type(cls):
+        # float32 for both fields; toggle_state is 0.0/1.0 (bool semantics),
+        # robot_can_toggle_steps is an integer stored as float.
+        return th.float32
+
+    @classproperty
+    def value_name(cls):
+        # Used by the parent's _dump_state / _load_state for the raw tensor column.
+        # Individual fields are accessed by column index, not by this key directly.
+        return "toggle_state"
```
cgokmen: I wonder what this does, tbh. I'd think "values" would be good enough for all classes so long as they're already keyed by class name, but for the sake of backwards compatibility lets leave it in.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -92,38 +218,163 @@ def global_update(cls):
...
+        new_values = values.clone()
+
+        # Column index constants (defined locally for readability; no object creation overhead).
+        COL_OPEN, COL_FORCE_OFF, COL_ACTIVE, COL_IN_CONTACT, COL_CAN_TOGGLE, COL_FLIP = range(6)
+
+        # Reset all scratch columns for this step (in-place fill, no allocation).
+        cls._scratch_masks.fill_(False)
+
+        # Step 1: requires_closed guard.
+        # Open is not tensorized; loop only over the _requires_closed subset.
```
cgokmen: TODO: It should be!

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,146 @@
...
+class ToggledOn(TensorizedValueState, BooleanStateMixin, LinkBasedStateMixin):
+    """
+    Boolean state representing whether a button-like object has been toggled on.
+
+    Tensorized storage
+    ------------------
+    Inherits TensorizedValueState. All per-object mutable values are stored in class-level
+    tensors and lists (same index as VALUES):
+
+        VALUES[:, 0]     — toggle on/off (float32, 0.0 = off / 1.0 = on)
```
cgokmen: Lets change the tensorized states such that VALUES is an array like (NUM_SCENES, NUM_OBJECTS_PER_SCENE). The 2nd dimension index can no longer come from an object-keyed dictionary since it refers to a different object in every scene. So instead key the OBJ_IDXES dict by obj.relative_prim_path (which is scene-relative, and will be the same for all instances of the same object across different scenes). The reverse lookup should be OBJS_BY_IDX[SCENE, OBJ_IDX] = obj  (this is a 2d list)

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -92,38 +218,163 @@ def global_update(cls):
             for scene in og.sim.scenes
         ]
```
cgokmen: Precompute the per-scene robot fingers mask. (S, 1, R). This should be computed ahead of time at initialization.

Also precompute the per-scene toggleable objects mask. This should look like (S, N, C).

Now switch the loop below this comment to use is_in_contact_batch. Update is_in_contact_batch to allow an arbitrary number of dimensions before the R/C dimension. It should be able to broadcast the Sx1xR to be SxNxR, then compare it with the SxNxC and return a SxN tensor of booleans.

For this, also update the RigidContactAPI to use a tensor across scenes instead of a dictionary. Assume (and assert at init time) that the set of sensor_paths and filter_paths are the same across scenes at initialization time, and track a single relative-prim-path-to-row and relative-prim-path-to-col dictionary across different scenes. Key it again by obj.relative_prim_path

cgokmen: Note: let's remove the _add_obj and _remove_obj API. There should be a function that's called after any object changes, just like RigidContactAPI's initialize_view. Just like function, the initialization function should be able to take into account what indices existed previously (and retain their data).

cgokmen (changes requested):

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+        )
+        if not has_rigid_bodies:
+            self.clear()
+            return
+
+        # --- Phase A: create per-scene PhysX views, build flat _PATH_TO_IDX, collect poses ---
+        poses_list = []
+        physx_offsets = [0]  # cumulative physx link counts per scene
+
+        og.sim.pi.update_simulation(elapsedStep=0, currentTime=og.sim.current_time)
```
cgokmen: Remove this after syncing from main

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+        with suppress_omni_log(channels=["omni.physx.tensors.plugin"]):
+            for scene_idx, scene in enumerate(og.sim.scenes):
+                view = og.sim.physics_sim_view.create_rigid_body_view(pattern=f"/World/scene_{scene_idx}/*/*")
```
cgokmen: Use a single view across all scenes

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+        self._IDX_TO_PATH = []  # list[link_absolute_prim_path]
+
+        # Raw poses from PhysX — flat across all scenes, physx_tracked links only.
+        # shape: (N_links_physx_tracked, 7) — [px, py, pz, qx, qy, qz, qw]
+        # N_links_physx_tracked = sum_s(P_physx_s); does NOT grow when untracked links are added.
+        self._POSES = None
+
+        # Scene boundary indices into _POSES (physx_tracked links only).
+        # _SCENE_PHYSX_OFFSETS[s] = flat start index of scene s; shape: (S+1,) int64
+        self._SCENE_PHYSX_OFFSETS = None
```
cgokmen: This will be gone after you move to single-view

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+            surviving = [
+                (old_idx, self._PATH_TO_IDX[p]) for p, old_idx in prev_path_to_idx.items() if p in self._PATH_TO_IDX
+            ]
+            if surviving:
+                old_idxs = th.tensor([o for o, _ in surviving], dtype=th.long)
+                new_idxs = th.tensor([n for _, n in surviving], dtype=th.long)
+                self._POSES[new_idxs] = prev_poses[old_idxs]
+
+        _pose_matrices_cpu = self._poses_to_matrices(self._POSES)  # (N_links_physx_tracked, 4, 4) — CPU
+        self._POSE_MATRICES = _pose_matrices_cpu.cuda()  # GPU — will be kept in sync via async_copy_to_gpu()
```
cgokmen: Remove phase B, instead read directly from the rigid body view.

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+                    abs_path = link.prim_path
+
+                    # Register articulated kinematic-only objects' links
+                    if is_untracked and abs_path not in self._PATH_TO_IDX:
+                        idx = len(self._PATH_TO_IDX)
+                        self._PATH_TO_IDX[abs_path] = idx
+                        self._IDX_TO_PATH.append(abs_path)
+
+                    # Now if path not registered, meaning it's an articulated non-kinematic link (e.g. Robot)
+                    # Skip, should not be covered by this view
```
cgokmen: Not true. Robot links should be included

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+        N_physx_total = len(self._PATH_TO_IDX)  # boundary between physx_tracked and physx_untracked
+        raw_local_points = {}  # {flat_idx: (V, 4) tensor}
+        kinematic_poses_to_fill = []  # [(flat_idx, pose_7)]
+
+        for scene_idx, scene in enumerate(og.sim.scenes):
+            for obj in scene.objects:
+                is_untracked = obj.kinematic_only and obj.prim_type != PrimType.CLOTH
+                for link in obj.links.values():
+                    abs_path = link.prim_path
+
```
cgokmen, line 314: Check that your guess for is_untracked is _exactly_ correct

`assert is_untracked == (abs_path not in self._PATH_TO_IDX)`

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+                    # Register articulated kinematic-only objects' links
+                    if is_untracked and abs_path not in self._PATH_TO_IDX:
```
cgokmen: if is_untracked

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+                        idx = len(self._PATH_TO_IDX)
+                        self._PATH_TO_IDX[abs_path] = idx
+                        self._IDX_TO_PATH.append(abs_path)
+
+                    # Now if path not registered, meaning it's an articulated non-kinematic link (e.g. Robot)
+                    # Skip, should not be covered by this view
+                    path_idx = self._PATH_TO_IDX.get(abs_path)
+                    if path_idx is None:
+                        continue
```
cgokmen: Remove

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,408 @@ def create_joint(
...
+
+                    # Collect poses for physx_untracked links
+                    if is_untracked and path_idx >= N_physx_total:
+                        pos, quat_wxyz = link.get_position_orientation()
+                        quat_xyzw = th.cat([quat_wxyz[1:], quat_wxyz[:1]])
+                        kinematic_poses_to_fill.append((path_idx, th.cat([pos, quat_xyzw])))
```
cgokmen: Move this into the is_untracked branch above

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -2389,6 +2833,10 @@ def scene_relative_prim_path_to_absolute(scene, relative_prim_path):
     if relative_prim_path.startswith("/OmniGraph"):
         return relative_prim_path
 
+    # Special case for global floor plane collision prim — already an absolute path
+    if relative_prim_path.endswith("collisionPlane"):
```
cgokmen: Check the full path equality

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -2415,6 +2863,10 @@ def absolute_prim_path_to_scene_relative(scene, absolute_prim_path):
     if absolute_prim_path.startswith("/OmniGraph"):
         return absolute_prim_path
 
+    # Special case for global floor plane collision prim — not scene-scoped, return unchanged
+    if absolute_prim_path.endswith("collisionPlane"):
```
cgokmen: Check the full path equality

OmniGibson/omnigibson/simulator.py
```diff
@@ -1150,6 +1171,7 @@ def _non_physics_step(self):
                     # For this same reason, after we finish the loop, we keep any objects that are yet to be initialized
                     # First call zero-physics step update, so that handles are properly propagated
                     self.refresh_physics()
+                    initialize_tensorized_states()
```
cgokmen: Don't do this here. Defer to the next edited area where the object states are updated

OmniGibson/omnigibson/object_states/__init__.py
```diff
@@ -34,6 +34,21 @@
 from omnigibson.object_states.touching import Touching
 from omnigibson.object_states.under import Under
 
+
+def initialize_tensorized_states():
```
cgokmen: Remove this. See simulator.py for how to do this

OmniGibson/omnigibson/object_states/__init__.py
```diff
@@ -77,4 +92,5 @@
     "Under",
     "Unfolded",
     "VerticalAdjacency",
+    "initialize_tensorized_states",
```
cgokmen: Remove

OmniGibson/omnigibson/simulator.py
```diff
@@ -1176,10 +1198,17 @@ def _non_physics_step(self):
 
                 # Propagate states if the feature is enabled
                 if gm.ENABLE_OBJECT_STATES:
-                    # Step the object states in global topological order (if the scene exists)
+                    # TensorizedValueState global_updates (GPU computation)
                     for state_type in self.object_state_types_requiring_update:
-                        if issubclass(state_type, GlobalUpdateStateMixin):
+                        if issubclass(state_type, TensorizedValueState):
                             state_type.global_update()
```
cgokmen: Call your initialize_view functions like this. Make sure object_state_types_requiring_update is sorted topologically (it already should be)

OmniGibson/omnigibson/simulator.py
```diff
@@ -1321,6 +1350,19 @@ def step_physics(self):
             # Accumulate contact data from this physics step and then flush to cache.
             # We normally do this in _non_physics_step, but step_physics bypasses that so we do it here.
             RigidContactAPI.update_contact_cache()
+            RigidBodyViewAPI.update_pose_cache()
+            ArticulatedObjectViewAPI.update_dof_cache()
+
+            RigidBodyViewAPI.async_copy_to_gpu()
+            ArticulatedObjectViewAPI.async_copy_to_gpu()
+            self.CPU_TO_GPU.synchronize()
```
cgokmen: Is this exactly equal to the chunk called in non_physics_step? If so, put these in a utility function on Simulator. update_apis or something. call here and there.

OmniGibson/omnigibson/simulator.py
```diff
@@ -1321,6 +1350,19 @@ def step_physics(self):
...
+
+            # TODO (andi) is this ideal?
+            # Keep AABB VALUES fresh so callers (e.g. Inside._set_value after sample_kinematics) can read
+            # up-to-date AABBs without waiting for the next full _non_physics_step.
+            if gm.ENABLE_OBJECT_STATES and AABB.VALUES is not None and AABB.VALUES.numel() > 0:
+                AABB.global_update()  # issues async copy on GPU_TO_CPU inside global_update()
+                self.GPU_TO_CPU.synchronize()  # wait before caller reads VALUES_CPU
```
cgokmen: This definitely can't be here. What is the solution? sample_kinematics shouldn't use the AABB _state_ for its AABB computations. Use the object's AABB property instead.

OmniGibson/omnigibson/simulator.py
Andorexad: Here it can cover cases when new objects get initialized. what about the case of obj get removed? Can we keep the `initialize_tensorized_state()` and call it inside `removing_objects()`?

OmniGibson/omnigibson/simulator.py
```diff
@@ -1113,7 +1124,10 @@ def update_handles(self):
 
             # Finally update any unified views
             RigidContactAPI.initialize_view()
+            RigidBodyViewAPI.initialize_view()
+            ArticulatedObjectViewAPI.initialize_view()
             ControllableObjectViewAPI.initialize_view()
+            initialize_tensorized_states()
```
cgokmen: Convert this to a loop that goes through all tensorized object states like in non_physics_step, and initialize them. Don't have this initialize_tensorized_states function.

OmniGibson/omnigibson/simulator.py
cgokmen: Remove this, see what happens.

OmniGibson/omnigibson/object_states/aabb.py
```diff
@@ -1,8 +1,134 @@
...
+    # list[int] — O indices of cloth objects (per-object fallback)
+    CLOTH_OBJ_IDXS = None
+
+    # (S*O, 3) scratch buffers — pre-allocated in initialize_view(), reused each step via fill_()
+    _AABB_LO = None
+    _AABB_HI = None
+
+    @classproperty
+    def value_shape(cls):
+        return (6,)
```
cgokmen, line 44: Maybe change this into a 2,3

OmniGibson/omnigibson/object_states/aabb.py
```diff
@@ -1,8 +1,134 @@
...
+
+        # Initialize new VALUE slots for objects that just appeared
+        for rel_path, obj_idx in cls.OBJ_IDXS.items():
+            if rel_path not in prev_rel_paths:
+                for s_idx in range(S):
+                    if cls.IDX_OBJS[s_idx][obj_idx] is not None:
+                        cls.VALUES[s_idx, obj_idx] = 0.0  # will be correct after first _update_values
+
+    @classmethod
+    def _update_values(cls, values):
```
cgokmen, line 113: This should just mutate values in-place. Make it not return anything on any of the classes. Instead, make TensorizedValueState retain a separate prev-value tensor, which it uses to compare against this.

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -190,6 +190,361 @@ def create_joint(
...
+                        cls._IDX_TO_PATH.append(abs_path)
+
+                        pos, quat_wxyz = link.get_position_orientation()
+                        quat_xyzw = th.cat([quat_wxyz[1:], quat_wxyz[:1]])
+                        poses_list.append(th.cat([pos, quat_xyzw]))
+
+                    # Collect local collision points for AABB
+                    # This is require for all registered links
+                    pts = link.collision_boundary_points_local  # (V, 3) or None
+                    if pts is not None:
```
cgokmen: Make sure things still work when pts is none / empty

OmniGibson/omnigibson/object_states/slicer_active.py
```diff
@@ -37,83 +37,78 @@ def global_initialize(cls):
...
+
+        # Base class rebuilds OBJ_IDXS, IDX_OBJS, VALUES (with value carry-over for survivors)
+        super().initialize_view()
+
+        S = len(cls.IDX_OBJS)
+        N = len(cls.OBJ_IDXS)
+
+        # Rebuild SLICER_LINK_PATHS indexed by obj_idx.
+        # Use relative prim paths so that is_in_contact lookups work with the shared path mapping.
+        cls.SLICER_LINK_PATHS = [[] for _ in range(N)]
```
cgokmen: Here, follow the same setup as in aabb, store all slicer links and sliceable links indices as found in rigid contact API, and then later you can use is_in_contact_batch

OmniGibson/omnigibson/object_states/open_state.py
```diff
@@ -169,36 +187,90 @@ def _find_articulated_joints(prim):
...
+
+        cls.OPENABLE_MASK = th.zeros(O, max_dof, dtype=th.bool, device="cuda")
+        cls.THRESHOLDS_S1 = th.zeros(O, max_dof, device="cuda")
+        cls.DIRECTIONS_S1 = th.zeros(O, max_dof, device="cuda")
+        cls.THRESHOLDS_S2 = th.zeros(O, max_dof, device="cuda")
+        cls.DIRECTIONS_S2 = th.zeros(O, max_dof, device="cuda")
+        cls.BOTH_SIDES = th.zeros(O, dtype=th.bool, device="cuda")
+
+        for _rel_path, obj_idx in cls.OBJ_IDXS.items():
+            obj = next(row[obj_idx] for row in cls.IDX_OBJS if row[obj_idx] is not None)
```
cgokmen: Why would this be None? Don't allow this

OmniGibson/omnigibson/object_states/open_state.py
```diff
@@ -115,20 +118,35 @@ def _get_relevant_joints(obj):
...
+
+    # (O,) bool — whether each object uses both-sides open logic
+    BOTH_SIDES = None
 
-        # Check the metadata info to get relevant joints information
-        self.relevant_joints_info = _get_relevant_joints(self.obj)
-        assert self.relevant_joints_info[1], f"No relevant joints for Open state found for object {self.obj.name}"
+    # (S*O,) int64 — pre-built row indices into ArticulatedObjectViewAPI._POSITIONS
+    # rows [obj_idx*S .. (obj_idx+1)*S-1] = scenes 0..S-1 for object obj_idx
+    OBJ_IDXES_IN_ARTICULATION_VIEW = None
```
cgokmen: This all needs to be (S, O, ...) because the same object across different scenes will have different numbers of joints. Come up with a unit test for this and put it in test_multiple_envs. For example: same named object in scene 0 has 1 joint, in scene 1 has 2 joints, if you open the 2nd joint in scene 2, what happens?

OmniGibson/omnigibson/object_states/open_state.py
```diff
@@ -169,36 +187,90 @@ def _find_articulated_joints(prim):
...
+        open_s1 = ((pos - t1) * d1 > 0) & mask  # (S, O, max_dof)
+        any_s1 = open_s1.any(dim=2)  # (S, O)
+
+        open_s2 = ((pos - t2) * d2 > 0) & mask
+        any_s2 = open_s2.any(dim=2)
+
+        both = cls.BOTH_SIDES.unsqueeze(0)  # (1, O)
+        is_open = th.where(both, any_s1 & any_s2, any_s1)  # (S, O)
+
+        return is_open.float()
```
cgokmen: save as boolean

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+    Named (S, N) bool tensors allocated in initialize_view(); reused in-place every step to
+    avoid per-call allocations:
+        _mask_force_off, _mask_active, _mask_in_contact, _mask_can_toggle, _mask_flip
+
+    Per-instance state:
+        self.scale      — temporary constructor arg; used during _initialize() to populate
+                          cls.scales[obj_idx], not accessed at runtime
+    """
+
+    # th.Tensor (S, N) float32: robot hand-in-marker step counter per (scene, object).
```
cgokmen: S, O - standardize across all files (another refers them to object_count)

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+
+    # th.Tensor (S, N) bool: whether each tracked object requires Open state to be False
+    # before it can be toggled. Construction-time constant.
+    _requires_closed = None
+
+    # (N,) long: index into Open.VALUES columns for each toggle object (-1 if no Open state).
+    # Built once in initialize_view(); used to vectorize the requires_closed guard.
+    _open_col_idx = None
+
+    # list[GeomPrim]: visual toggle-button markers, one per tracked object type (N index).
```
cgokmen: maybe index S, N?

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+    # Populated during _initialize(); used in _check_overlap and color updates.
+    visual_markers = None
+
+    # list[float | th.Tensor]: per-object marker scale (N index).
+    # Populated during _initialize(); used in _check_overlap for sphere radius.
+    scales = None
+
+    # Pre-computed contact masks built in initialize_view(); reused every step.
+    _finger_contact_query_mask = None  # list[Tensor(1, R_s) | None] — finger row mask per scene
+    _obj_contact_with_mask = None  #  list[Tensor(N, C_s) | None] — toggle-object col masks per scene
```
cgokmen: This needs to be S, N, C_s

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -77,55 +296,189 @@ def get_optional_dependencies(cls):
...
+                scene_idx=scene_idx,
+                query_masks=finger,  # (1, R_s)
+                with_masks=object,  # (N, C_s)
+                ignore_masks=None,
+                current_only=False,
+            )  # (N,) bool
+            cls._finger_contact_result[scene_idx].copy_(result)
+
+        # Step 2 - batch value update (calls _update_values, then fires state_updated)
+        super().global_update()
```
cgokmen: All this just needs to go inside update_values?

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+    # list[float | th.Tensor]: per-object marker scale (N index).
+    # Populated during _initialize(); used in _check_overlap for sphere radius.
+    scales = None
+
+    # Pre-computed contact masks built in initialize_view(); reused every step.
+    _finger_contact_query_mask = None  # list[Tensor(1, R_s) | None] — finger row mask per scene
+    _obj_contact_with_mask = None  #  list[Tensor(N, C_s) | None] — toggle-object col masks per scene
+
+    # (S, N) bool tensor: result of is_in_contact_batch from previous global_update call.
+    # Written by global_update(), read by _update_values().
```
cgokmen: fix comment when moved correctly into update values

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+
+    # Class-level color constants: defined once, reused every flip and _set_value call.
+    # Avoids allocating new th.tensor([...]) on every color update.
+    COLOR_ON = th.tensor([0, 1.0, 0])  # green  — toggle is on
+    COLOR_OFF = th.tensor([1.0, 0, 0])  # red    — toggle is off
+
+    @classproperty
+    def value_type(cls):
+        # float32; toggle_state is 0.0/1.0 (bool semantics)
+        return th.float32
```
cgokmen: bool?

OmniGibson/omnigibson/object_states/tensorized_value_state.py
```diff
@@ -2,145 +2,170 @@
...
+        Carry-over: values for objects whose relative_prim_path still exists are preserved;
+        new slots are initialized to zero (subclasses override to set correct defaults).
+        """
+        # Snapshot for carry-over (OBJ_IDXS / VALUES are None on the very first call)
+        prev_obj_idxs = dict(cls.OBJ_IDXS) if cls.OBJ_IDXS is not None else {}
+        prev_values = cls.VALUES.clone() if cls.VALUES is not None and cls.VALUES.numel() > 0 else None
+
+        # Reset
+        cls.global_initialize()
+
```
cgokmen, line 83: Add assertion that every scene has the same number of objects, and the same object at the same relative prim path has this state with the exact same settings.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+        cls._open_col_idx = th.full((N,), -1, dtype=th.long, device="cuda")
+        for obj_idx, obj in enumerate(cls.IDX_OBJS[0] if cls.IDX_OBJS else []):
+            if obj is not None and Open in obj.states:
+                open_idx = Open.OBJ_IDXS.get(obj.relative_prim_path, -1)
+                cls._open_col_idx[obj_idx] = open_idx
+
+        # Rebuild visual_markers and scales lists indexed by N
+        # Carry over existing markers/scales for surviving objects; new slots stay None
+        cls.visual_markers = [None] * N
+        cls.scales = [None] * N
```
cgokmen: S, N

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+            cls._mask_flip = th.zeros((0, 0), dtype=th.bool, device="cuda")
+            return
+
+        # Build per-scene contact masks
+        #     broadcasting against (N, C_s) in is_in_contact_batch gives one result per object.
+        cls._finger_contact_query_mask = []
+        cls._obj_contact_with_mask = []
+
+        for scene_idx, scene in enumerate(og.sim.scenes):
+            # Early return if this scene doesn't have contact view
```
cgokmen: It should

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+    # list[GeomPrim]: visual toggle-button markers, one per tracked object type (N index).
+    # Populated during _initialize(); used in _check_overlap and color updates.
+    visual_markers = None
+
+    # list[float | th.Tensor]: per-object marker scale (N index).
+    # Populated during _initialize(); used in _check_overlap for sphere radius.
+    scales = None
+
+    # Pre-computed contact masks built in initialize_view(); reused every step.
+    _finger_contact_query_mask = None  # list[Tensor(1, R_s) | None] — finger row mask per scene
```
cgokmen: These should be tensors

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -77,55 +296,189 @@ def get_optional_dependencies(cls):
...
+        cls._mask_force_off.fill_(False)
+        cls._mask_active.fill_(False)
+        cls._mask_in_contact.fill_(False)
+        cls._mask_can_toggle.fill_(False)
+        cls._mask_flip.fill_(False)
+
+        # Step 1: requires_closed guard — vectorized via Open.VALUES.
+        # _open_col_idx (N,): column index into Open.VALUES for each toggle object (-1 = no Open state).
+        if cls._requires_closed.any():
+            valid = cls._requires_closed & (cls._open_col_idx >= 0).unsqueeze(0)  # (S, N)
```
cgokmen: You don't need this, because we check and assert that everything that has requires_closed already is guaranteed to have a nonzero open col idx. Assume this to be true.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -21,28 +21,247 @@
...
+
+        # Build _open_col_idx: (N,) long — maps each toggle-object N-index to its column in
+        # Open.VALUES.  -1 for objects that have no Open state.
+        # Uses scene-0 objects as representative (all scenes share the same object types).
+        cls._open_col_idx = th.full((N,), -1, dtype=th.long, device="cuda")
+        for obj_idx, obj in enumerate(cls.IDX_OBJS[0] if cls.IDX_OBJS else []):
+            if obj is not None and Open in obj.states:
+                open_idx = Open.OBJ_IDXS.get(obj.relative_prim_path, -1)
+                cls._open_col_idx[obj_idx] = open_idx
+
```
cgokmen: Assert here that anything that has requires_closed also has an open col idx

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -77,55 +296,189 @@ def get_optional_dependencies(cls):
...
+        # Single clone — required so the base class can detect changes via new_values != cls.VALUES.
+        new_values = values.clone()
+        S = values.shape[0]
+
+        # Reset all scratch masks for this step (in-place fill, no allocation).
+        cls._mask_force_off.fill_(False)
+        cls._mask_active.fill_(False)
+        cls._mask_in_contact.fill_(False)
+        cls._mask_can_toggle.fill_(False)
+        cls._mask_flip.fill_(False)
```
cgokmen: Name these so it's super clear what each mask does

cgokmen (approved):

PR #2022, opened 2026-03-18
Temporarily disable failing tests on main and add a few more test fixes

stefren:


stefren:
tracked in https://github.com/StanfordVL/BEHAVIOR-1K/issues/2023

PR #2051, opened 2026-03-27
Fix RigidContactAPI: don't create None views if scene contains no objects

cgokmen:


stefren:
where's the regression test though :sweat_smile:

PR #2094, opened 2026-04-09
Enforce USD-Fabric sync via editing_usd() context

cgokmen:
With the migration from OmniHydra to Fabric Scene Delegate (FSD), USD edits are no longer automatically propagated to Fabric. This PR introduces a systematic mechanism to ensure every USD edit is properly synchronized, and to detect any that aren't.

- **New `editing_usd()` context manager on `Simulator`** — wraps any block of code that mutates the USD stage. On exit, calls `SynchronizeToFabric()` to flush changes (much faster than rendering!). Asserts no nesting (all USD edits for a logical operation belong in one context). Currently we do this at a very low level (e.g. sync near every line of code that we don't manage), which causes a lot of extra syncs than necessary, but as a reminder up until this version EVERY usd edit was causing a sync so this is still strictly better.
- **USD edit guard** — a `Tf.Notice.ObjectsChanged` listener that raises `RuntimeError` with full backtrace if any USD edit occurs outside an `editing_usd()` block or simulator lifecycle method. Enabled after simulator init.
- **`set_position_orientation` scale fix** — fixed a bug where writing position/orientation via the Fabric hierarchy was losing scale information.
- **ProfilingEnv redo** — instead of having a profiling env that replicates much of Environment and Simulator's functionality (which is very brittle and possibly inaccurate), we hook directly into the appropriate spots in simulator using a Profiler class.

With this, we can also start an effort to remove random og.sim.render() calls scattered across the codebase.


OmniGibson/omnigibson/sensors/vision_sensor.py
```diff
@@ -219,7 +209,7 @@ def _post_load(self):
         else:
             viewport = lazy.omni.kit.viewport.utility.create_viewport_window()
             # Take a render step to make sure the viewport is generated before docking it
-            render()
+            og.sim.render()
```
wensi-ai, line 214: are we sure this is the same thing as render()?

OmniGibson/omnigibson/sensors/vision_sensor.py
```diff
@@ -962,7 +956,7 @@ def clear(cls):
             sensor.remove()
 
         # Render to update
-        render()
+        og.sim.render()
 
         cls.SEMANTIC_REMAPPER = Remapper()
         cls.INSTANCE_REMAPPER = Remapper()
```
wensi-ai, line 966: Do you think for the property setter (e.g. image_height, we should wrap those into editing_usd() context?

OmniGibson/omnigibson/sensors/vision_sensor.py
```diff
@@ -876,7 +870,7 @@ def active_camera_path(self, path):
         self._viewport.viewport_api.set_active_camera(path)
         # Requires 6 updates to propagate changes
         for i in range(6):
```
wensi-ai, line 876: Hmm these numbers are still very adhoc, we should probably investigate this further in the future.

OmniGibson/tests/conftest.py
```diff
@@ -34,8 +34,8 @@ def env(request, stopped_env):
         if name not in ("env", "stopped_env", "request"):
             request.getfixturevalue(name)
 
-    og.sim.step()
     og.sim.play()
+    og.sim.step()
```
wensi-ai: Is this intended?

stefren: why do we need this?

.github/workflows/profiling.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Convert deep profile to HTML
         if: matrix.deep_profile
         run: |
-          for prof in load.prof pre_physics_step.prof post_physics_step.prof non_physics_step.prof; do
+          for prof in load.prof step.prof pre_physics_step.prof post_physics_step.prof non_physics_step.prof; do
```
stefren, line 88: nit; should we pull this into a list of args and expand it here and below

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -26,6 +25,25 @@
...
+        api: The USD API class (e.g. lazy.pxr.UsdPhysics.RigidBodyAPI)
+
+    Returns:
+        The API wrapper for the prim
+    """
+    if prim.HasAPI(api):
+        return api(prim)
+    with og.sim.editing_usd():
+        return api.Apply(prim)
+
```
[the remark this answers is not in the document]
stefren: yeah i think this is a good point. i'm not sure I generally understand the point of this function, it's because we don't know if some APIs are available in fabric or not? If so it would be better to test and statically use `editing_usd` in cases where it's not available, rather than trying to do this dynamically at runtime. also now we have two editing_usd wraps when we apply the API and then modify it again as in some callsites

OmniGibson/omnigibson/objects/light_object.py
```diff
@@ -131,7 +131,9 @@ def _post_load(self):
         self._light_link.load(self.scene)
 
         # Apply Shaping API and set default cone angle attribute
-        lazy.pxr.UsdLux.ShapingAPI.Apply(self._light_link.prim).GetShapingConeAngleAttr().Set(180.0)
+        shaping_api = ensure_usd_api(self._light_link.prim, lazy.pxr.UsdLux.ShapingAPI)
+        with og.sim.editing_usd():
+            shaping_api.GetShapingConeAngleAttr().Set(180.0)
```
stefren, line 136: e.g. i think this would be cleaner without the `ensure_usd_api` call? we should just know if it's available in fabric or not

OmniGibson/omnigibson/prims/xform_prim.py
```diff
@@ -113,37 +114,37 @@ def _set_xform_properties(self):
...
+
+            if "xformOp:orient" not in prop_names:
+                xform_op_rot = xformable.AddXformOp(
+                    lazy.pxr.UsdGeom.XformOp.TypeOrient, lazy.pxr.UsdGeom.XformOp.PrecisionDouble, ""
+                )
+            else:
+                xform_op_rot = lazy.pxr.UsdGeom.XformOp(self._prim.GetAttribute("xformOp:orient"))
+            xformable.SetXformOpOrder([xform_op_translate, xform_op_rot, xform_op_scale])
+
+        XFormPrim.set_position_orientation(self, position=current_position, orientation=current_orientation)
```
stefren, line 148: does this need to be separated from the above usd edit? would maybe add a comment

OmniGibson/omnigibson/sensors/vision_sensor.py
```diff
@@ -252,7 +244,7 @@ def _post_load(self):
 
         # Requires 4 render updates to propagate changes
         for i in range(4):
-            render()
+            og.sim.render()
```
stefren, line 247: is this still necessary?

OmniGibson/omnigibson/sensors/vision_sensor.py
```diff
@@ -876,7 +872,7 @@ def active_camera_path(self, path):
         self._viewport.viewport_api.set_active_camera(path)
         # Requires 6 updates to propagate changes
         for i in range(6):
-            render()
```
stefren, line 879: why are the number of render calls needed so arbitrary lol

OmniGibson/omnigibson/systems/macro_particle_system.py
```diff
@@ -1248,13 +1248,12 @@ def process_particle_object(self):
         self._particle_offset = particle_offset
         self._particle_radius = particle_radius
 
-    def refresh_particles_view(self):
+    def update_handles(self):
```
stefren, line 1251: how does this function differ from `og.sim.update_handles` and when do we use one or the other? Can you update the docstring?

OmniGibson/omnigibson/utils/profiling_utils.py
```diff
@@ -1,13 +1,75 @@
...
 
+class Profiler:
+    """
+    Lightweight profiler that tracks wall-clock timing across enable/disable cycles.
+
+    Used by the Simulator's `with_profiler` decorator. Each enable/disable pair records
+    one timing sample. Callers can read cumulative or per-call averages at any time.
+
+    Optionally wraps a cProfile.Profile for deep (function-level) profiling when
+    gm.ENABLE_PROFILING is set.
```
stefren: nit but we should probably change that flag to be called `ENABLE_DEEP_PROFILING`? otherwise feels a bit confusing to me

OmniGibson/omnigibson/utils/profiling_utils.py
```diff
@@ -16,104 +78,30 @@ def gpu_in_use_by_this_process(gpu_handle, pid: int) -> bool:
...
-    }
+    compute_pids = {process.pid for process in pynvml.nvmlDeviceGetComputeRunningProcesses(gpu_handle)}
+    graphics_pids = {process.pid for process in pynvml.nvmlDeviceGetGraphicsRunningProcesses(gpu_handle)}
 
     pids_using_device = compute_pids | graphics_pids
 
     return len(pids_using_device & our_pids) > 0
 
 
-class ProfilingEnv(og.Environment):
```
stefren, line 38: are we not using this anymore?

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -11,7 +11,6 @@
 
 import omnigibson as og
 import omnigibson.lazy as lazy
-import omnigibson.utils.transform_utils as T
```
stefren, line 14: huh i think this is being used? don't you need to replace the callsites?

OmniGibson/omnigibson/utils/usd_utils.py
```diff
@@ -125,66 +143,68 @@ def create_joint(
...
+            joint_prim.GetAttribute("physics:breakForce").Set(break_force)
+        if break_torque is not None:
+            joint_prim.GetAttribute("physics:breakTorque").Set(break_torque)
 
-    # Possibly exclude this joint from the articulation
-    joint_prim.GetAttribute("physics:excludeFromArticulation").Set(exclude_from_articulation)
+        # Possibly (un-/)enable this joint
+        joint_prim.GetAttribute("physics:jointEnabled").Set(enabled)
 
-    # We update the simulation now without stepping physics if sim is playing so we can bypass the snapping warning from PhysicsUSD
```
stefren, line 184: hm, we don't need an extra sync (closing the context) here?

OmniGibson/omnigibson/simulator.py
```diff
@@ -433,15 +427,24 @@ def __init__(
...
 
             self._floor_plane = None
             self._skybox = None
             self._last_scene_edge = None
             self._stage_id = None
 
+            # USD edit guard: detects edits outside editing_usd() context
+            self._editing_usd = False
+            self._editing_usd_caller = None
+            self._in_sim_lifecycle = 0
```
stefren, line 446: why can't we use `True` and `False` here?

OmniGibson/omnigibson/simulator.py
```diff
@@ -433,15 +427,24 @@ def __init__(
...
+            self._deferred_usd_guard_error = None
+            self._usd_guard_enabled = False
+            self._usd_guard_listener = None
```
stefren, line 449: also should we add a way to turn off this mechanism? seems like it could impact performance and we mostly need it for debugging right?

stefren (commented):
most changes look good

OmniGibson/omnigibson/objects/usd_object.py
```diff
@@ -634,102 +634,109 @@ def _create_emitter_apis(self, emitter_type):
         # as the object moves. We put it under a dummy mesh so as not to force write synchronization to the actual
         # physx-tracked links (required when using Fabric), which causes physics issues
         dummy_mesh_path = f"{link.prim_path}/emitter"
-        lazy.pxr.UsdGeom.Sphere.Define(og.sim.stage, dummy_mesh_path)
+        with og.sim.editing_usd():
+            lazy.pxr.UsdGeom.Sphere.Define(og.sim.stage, dummy_mesh_path)
         relative_dummy_mesh_path = absolute_prim_path_to_scene_relative(self._scene, dummy_mesh_path)
         mesh = GeomPrim(relative_prim_path=relative_dummy_mesh_path, name=f"{self.name}_emitter")
         mesh.load(self._scene)
-        mesh.visible = False
```
stefren, line 640: oops i think you need to put this back

cgokmen: I really wonder how I did this lmao. thanks again

OmniGibson/omnigibson/sensors/vision_sensor.py
cgokmen: Unsure. Will test in follow-up.

OmniGibson/omnigibson/simulator.py
cgokmen: They are nested (play calls render and so on) so it prematurely comes out if you just use True/False.

OmniGibson/omnigibson/simulator.py
cgokmen: Doesn't seem to have much of an impact. I'm leaving it in for now, can remove later.

OmniGibson/omnigibson/systems/macro_particle_system.py
cgokmen: og.sim.update_handles calls this on everything it updates handles on. This already exists on objects, prims, etc - it was just named differently here. Docstring updated.

OmniGibson/omnigibson/utils/profiling_utils.py
cgokmen: done

OmniGibson/tests/conftest.py
cgokmen: I don't know. It didnt make any sense in its previous form (it was calling step on a nonplaying env...) so I reordered them. But in hindsight it must have meant this is pointless, so I removed it now.

OmniGibson/omnigibson/sensors/vision_sensor.py
cgokmen: We didn't understand this well at the time, but the renders are async. So effectively it could show up in one, two, five, ten steps based on your GPU. We should follow up to solve this.

OmniGibson/omnigibson/utils/profiling_utils.py
cgokmen: Seems outdated - I don't see this in the current state?

OmniGibson/omnigibson/utils/usd_utils.py
cgokmen: Where are you finding these versions of these files that you commented on? This doesn't exist in the final state of this PR.

.github/workflows/profiling.yml
cgokmen: ignoring for now so I can merge today

cgokmen (commented):
Completed self-review

wensi-ai (approved):
Thanks Cem!

PR #2095, opened 2026-04-09
Add documentation links to claude/agents.md

stefren:
FYI in general while testing Claude often still has to be prompted explicitly to use the documentation, otherwise it does not always default to looking it up even if it's quite relevant. But now you can e.g. ask it to follow links to the documentation or look at Omniverse or IsaacSim docs. I initially ran into an issue where it would add an extra `docs/` part to the URL which is why I added the extra sentence at the end, seems like it could be a bit finnicky still.


wensi-ai (approved):
Looks good, perhaps we can have a dedicated related doc section in the md file?

cgokmen (approved):

PR #2099, opened 2026-04-11
cache link.volume once and reuse after

yalcintur:
Cache link.volume per link in _post_load, so we ensure it's computed once instead of twice when initializing per-category mass


cgokmen (changes requested):
Check if `points` etc are cached properties on RigidPrim. If it is, change this also to a cached_property on RigidPrim instead of this adhoc caching

cgokmen (approved):

PR #2116, opened 2026-04-17
Per-robot RigidContactView for assisted-grasp contact positions

cgokmen:
- The `RigidContactAPI` refactor no longer exposes contact positions, which forced the assisted-grasp code path in `robot.py` to fall back to the legacy contact sensor interface — slow and vestigial.
- Each manipulation robot now owns its own `rigid_contact_view` (rows: any rigid body in its scene, columns: its finger links), rebuilt inside `Robot.update_handles` alongside the articulation view, and reads contact positions from it for `_create_assisted_grasp_joint_from_sim_state` — matching the v3.7.2 design but scoped per-robot instead of a global `GripperRigidContactAPI`.
- Drops the now-unused `acquire_contact_sensor_interface()` acquisition and `contact_sensor` property on the simulator, and a vestigial no-op `get_rigid_body_raw_data` call in `RigidPrim._initialize` (a holdover from the old `dc` handle era).


stefren (approved):

wensi-ai (approved):
lgtm

PR #2119, opened 2026-04-18
Serialize assisted grasping state

### LINKED ISSUE #1983: Assisted grasping serialization and unit tests
Currently we don't serialize AG state (we just dump it in a dict). We should serialize it and also add roundtrip dump/save tests, in multiple envs, and add golden tests also.

stefren:
close #1983


cgokmen (approved):
I made some changes, so do another review pass yourself before merging please. Thanks!

PR #2122, opened 2026-04-19
Fix - Non-fixed-base articulation objects with internal fixed joints (e.g. cart) get teleported to garbage poses at sim.play()

### LINKED ISSUE #2121: Non-fixed-base articulation objects with internal fixed joints (e.g. cart) get teleported to garbage poses at sim.play()
**To Reproduce**
create_stable_scene_json on house_double_floor_upper
- serving_cart_vvtcby_0 fails the stability check with lin_vel z = -98.67
- cart is in free-fall from z ≈ 273,248 m. 
- Its xformOp:translate ends up at (-2371081.75, -1273490.75, 273248.06). The value is deterministic across runs.

**Additional context**
The cart's USD contains a PhysicsFixedJoint connecting base_link to a nearly-massless meta link meta__base_link_fillable_0_0_link (mass=0.001, inertia=1e-5, breakForce=3.4e38). 
Because n_fixed_joints=1 and fixed_base=False, _preapply_articulation_root sets articulation_root_path = base_link (not the entity).                                           In EntityPrim.set_position_orientation (prims/entity_prim.py:1010-1050), this hits the else branch, which only calls self.root_link.set_position_orientation(...) on base_link and leaves every other body in the articulation at its USD default.

After _load_scene_prim_with_objects runs:                                                                                                                               
  - base_link xformOp:translate = (-1.0554, 1.6549, 0.3544) 
  - meta__base_link_fillable_0_0_link xformOp:translate = (0, 0, 0)

Explosion happens when `physx_fabric_interface.force_update` is fired after this. 

As temporary fix, we can re-applying each object's _init_state pose after env creation, as PhysX maintains the articulation tree post-creation.

kmy17518:
Addresses #2121 

Move all links when updating an articulation’s pose, not just the root

For non-fixed-base articulations, `EntityPrim.set_position_orientation` used to update only the root link’s transform. Other links, including fixed-jointed meta links, remained at their default USD poses. When `sim.play()` initialized the articulation from these inconsistent poses, joint constraints were immediately violated. With infinite break force and near-zero-mass meta links, this caused the PhysX solver to diverge and send the articulation to invalid world coordinates.

This change applies the same rigid transform delta to every link in the articulation. It aligns the behavior with the fixed-base case, where moving the entity prim implicitly moves all links together.


cgokmen:
I'm closing this in favor of #2127 which contains the more principled fix for this issue (letting PhysX take care of the joints)

PR #2131, opened 2026-04-21
Update inside setter

wensi-ai:
image

### DIFF
--- a/OmniGibson/omnigibson/object_states/inside.py
+++ b/OmniGibson/omnigibson/object_states/inside.py
@@ -18,12 +18,34 @@ class Inside(RelativeObjectState, KinematicsMixin, BooleanStateMixin):
         return deps
 
     def _set_value(self, other, new_value, reset_before_sampling=False):
+        """
+        Set the Inside state for this object with respect to another object (container).
+
+        This samples a random position inside the container's fillable volume, places the object,
+        lets it settle via physics, and verifies it's still inside.
+
+        The sampling strategy uses a two-phase approach:
+        1. First half of attempts: Sample from inset AABB (container bounds minus object extent).
+           This ensures the sampled position is at the object's centroid, not near the edges,
+           which makes it more likely to fit inside.
+        2. Second half of attempts: Sample from full container AABB. This is a fallback for
+           cases where the object is too large to fit entirely within the inset bounds.
+
+        Args:
+            other: The container object to place this object inside.
+            new_value: True to set Inside state (only True is supported).
+            reset_before_sampling: If True, reset this object before sampling.
+
+        Returns:
+            True if successfully placed inside, False otherwise.
+        """
         if not new_value:
             raise NotImplementedError("Inside does not support set_value(False)")
 
         if other.prim_type == PrimType.CLOTH:
             raise ValueError("Cannot set an object inside a cloth object.")
 
+        # Find the container's fillable meta link (fillable or openfillable)
         container_link = None
         for link in other.links.values():
             if link.is_meta_link and link.meta_link_type in macros.object_states.contains.CONTAINER_META_LINK_TYPES:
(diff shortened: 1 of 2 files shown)


OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                pos = aabb_low + th.rand(3) * (aabb_high - aabb_low)
+
+                if not container_link.check_points_in_volume(pos.unsqueeze(0)).item():
+                    continue
+
+                pos[2] += 0.05  # Add a small offset to ensure the object is above the bottom of the container
+                self.obj.set_position_orientation(position=pos, orientation=orientation)
+                self.obj.keep_still()
+
+                n_steps_max = int(0.5 / og.sim.get_physics_dt())
```
stefren: why 0.5?

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                        current_only=False,
+                    )
+                    and i < n_steps_max
+                ):
+                    og.sim.step_physics()
+                    i += 1
+                self.obj.keep_still()
+                other.keep_still()
+
+                for i in range(5):
```
stefren:
```suggestion
                for _ in range(5):
```
nit this is unused in this loop and less confusing if we don't override the other variable

stefren (approved):

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
         deps.update({AABB})
         return deps
 
-    def _set_value(self, other, new_value, reset_before_sampling=False, use_trav_map=False):
+    def _set_value(self, other, new_value, reset_before_sampling=False):
```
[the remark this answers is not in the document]
cgokmen: There should not be a need for this, we should just remove any such callers if they exist.

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
         if other.prim_type == PrimType.CLOTH:
             raise ValueError("Cannot set an object inside a cloth object.")
 
+        container_link = None
+        for link in other.links.values():
+            if link.is_meta_link and link.meta_link_type in macros.object_states.contains.CONTAINER_META_LINK_TYPES:
+                container_link = link
+                break
+
+        assert container_link is not None, f"Container object {other.name} must have a fillable meta link"
```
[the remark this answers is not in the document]
cgokmen: We're OK with assertions.

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+        aabb_low, aabb_high = container_link.visual_aabb
+
         for _ in range(os_m.DEFAULT_HIGH_LEVEL_SAMPLING_ATTEMPTS):
-            if sample_kinematics("inside", self.obj, other, use_trav_map=use_trav_map) and self.get_value(other):
-                return True
-            else:
-                og.sim.load_state(state, serialized=False)
+            pos = None
+            orientation = (
+                self.obj.sample_orientation()
```
cgokmen, line 92: What does this do? We should only sample an orientation that rotates the object around the Z axis. Compare to the approach seen in the original shadow casting.

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                if (hasattr(self.obj, "orientations") and self.obj.orientations is not None)
+                else th.tensor([0, 0, 0, 1.0])
+            )
+
+            self.obj.set_position_orientation(position=th.tensor([100, 100, 10]), orientation=orientation)
+            self.obj.keep_still()
+            og.sim.step_physics()
```
cgokmen: What's the point of this? I think it's unnecessary, hardcodes a bad position, steps the physics for no reason, and the object obviously won't be still if its falling

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+
+                n_steps_max = int(0.5 / og.sim.get_physics_dt())
+                i = 0
+                while (
+                    not RigidContactAPI.is_in_contact(
+                        scene_idx=self.obj.scene.idx,
+                        query_set=[self.obj],
+                        with_set=None,
+                        ignore_set=None,
+                        current_only=False,
```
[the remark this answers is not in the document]
cgokmen: Do this

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+            self.obj.keep_still()
+            og.sim.step_physics()
+
+            for _ in range(os_m.DEFAULT_LOW_LEVEL_SAMPLING_ATTEMPTS):
+                pos = aabb_low + th.rand(3) * (aabb_high - aabb_low)
+
+                if not container_link.check_points_in_volume(pos.unsqueeze(0)).item():
+                    continue
+
+                pos[2] += 0.05  # Add a small offset to ensure the object is above the bottom of the container
```
cgokmen: 1cm is enough imo. 5cm fall will add a bunch of momentum

OmniGibson/omnigibson/object_states/inside.py
cgokmen: +1

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                    i += 1
+                self.obj.keep_still()
+                other.keep_still()
+
+                for i in range(5):
+                    og.sim.step_physics()
+                i = 0
+                while th.norm(self.obj.get_linear_velocity()) > 1e-3 and i < n_steps_max:
+                    og.sim.step_physics()
+                    i += 1
```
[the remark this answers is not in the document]
cgokmen: +1

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+
+                og.sim.render()
+
```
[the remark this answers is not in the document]
cgokmen: Agreed. Also, does sample_kinematics ever call anything that would cause a non-physics step? Because without it, this get_value cannot be True I think, because the contacts, touching state, adjacency state etc wont update?

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                if self.get_value(other):
+                    return True
+
+                break
+
+            og.sim.load_state(state, serialized=False)
```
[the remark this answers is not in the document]
cgokmen: High prio

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                    og.sim.step_physics()
+                i = 0
+                while th.norm(self.obj.get_linear_velocity()) > 1e-3 and i < n_steps_max:
+                    og.sim.step_physics()
+                    i += 1
+
+                og.sim.render()
+
+                if self.get_value(other):
+                    return True
```
cgokmen: I wonder if it makes sense to still restore the original state here, but just place this object relative to its parent. That way motion of the parent object and other side effects could prevented. e.g.

```
if self.get_value(other):
  obj_relative_to_container = T.relative_pose_transform(*obj.get_position_orientation(), *container.get_position_orientation())   
  og.sim.load_state(state, serialized=False)
  obj_world_pose = T.pose_multiply(*container.get_position_orientation(), *obj_relative_to_container)  
  obj.set_position_orientation(*obj_world_pose)
```

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
             self.obj.reset()
 
+        aabb_low, aabb_high = container_link.visual_aabb
+
         for _ in range(os_m.DEFAULT_HIGH_LEVEL_SAMPLING_ATTEMPTS):
-            if sample_kinematics("inside", self.obj, other, use_trav_map=use_trav_map) and self.get_value(other):
-                return True
-            else:
-                og.sim.load_state(state, serialized=False)
+            pos = None
```
cgokmen: not necessary

OmniGibson/omnigibson/object_states/inside.py
```diff
@@ -17,24 +17,82 @@ def get_dependencies(cls):
...
+                if (hasattr(self.obj, "orientations") and self.obj.orientations is not None)
+                else th.tensor([0, 0, 0, 1.0])
+            )
+
+            self.obj.set_position_orientation(position=th.tensor([100, 100, 10]), orientation=orientation)
+            self.obj.keep_still()
+            og.sim.step_physics()
+
+            for _ in range(os_m.DEFAULT_LOW_LEVEL_SAMPLING_ATTEMPTS):
+                pos = aabb_low + th.rand(3) * (aabb_high - aabb_low)
```
cgokmen, line 107: I wonder if, for the first half of the low-level sampling attempts, we could make a wiser guess: inset the container AABB dimensions by half this object's AABB dimensions. So for example if the object is 10cm tall, it doesn't make any sense to make guesses in the bottom or top 5cm of the container volume. This will make things significantly faster for certain oblong container shapes. But it's sometimes infeasible and we still want to attempt things in that case, so we should fall back to the current approach if it doesn't work.

Also, in the first half's case, you should use the position obtained here as the AABB centroid, not as the base position.

cgokmen (changes requested):

OmniGibson/omnigibson/object_states/inside.py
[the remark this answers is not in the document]
wensi-ai: sample_kinematics does call og.sim.render() at the very end

OmniGibson/omnigibson/object_states/inside.py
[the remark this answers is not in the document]
wensi-ai: fixed

OmniGibson/omnigibson/object_states/inside.py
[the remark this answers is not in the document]
wensi-ai: fixed

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: fixed

OmniGibson/omnigibson/object_states/inside.py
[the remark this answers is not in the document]
wensi-ai: fixed

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: This is exactly the original code

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: It was 5cm before, I think we want to set a higher one to prevent from overlapping with the ground?

OmniGibson/omnigibson/object_states/inside.py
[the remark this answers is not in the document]
wensi-ai: Although I don't know about the non-physics step situation. Looks like it's not called anywhere but getter still succeeds?

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: direct copy from sample_kinematics. Added a comments to explain this.

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: fixed

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: removed

OmniGibson/omnigibson/object_states/inside.py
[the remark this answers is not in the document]
wensi-ai: Confirmed that render is not needed, removed

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: fixed

OmniGibson/omnigibson/object_states/inside.py
wensi-ai: Can you elaborate more on this?

cgokmen (approved):

PR #2132, opened 2026-04-21
Fix unresponsive goal-condition text color updates

yalcintur:
Goal-condition labels used to only change color when the window was resized. This PR switches the color update mechanism to omni.ui's native :selected pseudo-state: base style holds the red color, :selected overrides to green, and update_goal_status just flips label.selected.


stefren (approved):

wensi-ai (approved):
lgtm!

PR #2144, opened 2026-04-24
Apply replay transitions after load_state, not before

yalcintur:
What's happening in replaying the hdf5 of halve_an_egg: after touching the egg with a knife, the transition rule runs by which the whole egg disappears and two halved eggs appear in its place.

The bug: At step 1020 the replay code first removed the whole egg and added the halves so the scene no longer has the whole egg. Then it tried to load the recorded snapshot for step 1020, which still mentions the whole egg's ID.

Crash: AssertionError: Could not find object while deserializing with hash_key uuid: 27912356 (This UUID 27912356 is the whole egg.)

 The fix is swapping the order. First load the snapshot (scene has the whole egg). Then apply the "remove whole egg, add halves" change. Now the scene matches what the next step expects.


wensi-ai:
See #2151

PR #2150, opened 2026-04-24
Additional fixes for autogenerate script

stefren:
1. Adds additional checking that is used by `bddl_utils` so that it only shows models that should be valid at runtime. The only discrepancy now is that we don't check for required attachments because that requires a running simulator instance.
2. Only shows scenes that match the BDDL.
3. Briefly shows what models, if any, were used in 2025 (should be known-good). 
4. Fixes a bug with parent synsets and properly digs down into children synsets now.


wensi-ai (approved):
Looks good, in general though if there are multiple good objects for a synset we should try to diversify instead of using the same ones from last year.

PR #2152, opened 2026-04-25
Add profiling for vector environment

Andorexad:
Restacked onto #2329 (`vec/merge-vector-into-main`). Previously targeted `vector`.

Adds a profiling harness for the vectorized environment.

### Matrix — 20 combos

| axis | values |
|---|---|
| `num_envs` | **1**, **10** |
| rendering | on / off |
| object states | on / off |
| task type | simple / behavior / navigation |

Runs on the **titanrtx** self-hosted runners.

### Included fix: `/rtx/descriptorSets`

Rendering combos at `num_envs > 1` died on the titanrtx runners with:

```
[gpu.foundation.plugin]  Unable to allocate descriptor sets.
[rtx.denoising.plugin]   Failed to allocate ParameterBlock resources
```

Deterministic — 5/5 at 10 envs, and **still failing after dropping to 8 envs** (24 render
products), so it is not a matter of shaving the env count. Not a VRAM issue either: those combos
peak around 12 GB on a 32 GB card.

OmniGibson launches a trimmed custom `.kit` experience, so it never picks up the
`/rtx/descriptorSets=360000` that Isaac Sim's own extensions set (71 occurrences across 66
extensions) and instead inherits Kit's built-in default of **10000**. r1pro carries 3 cameras, so
the budget runs out once enough render products exist. A 5090 happens to fit 30 products within
10000; Turing does not.

The setting caps the descriptor pool rather than preallocating it, so raising it 36x costs well
under a percent of VRAM — measured on a 5090, same combo, clean run:

| | 10000 | 360000 |
|---|---|---|
| 1 env, render, simple, states on | 10.42 GB | 10.48 GB |
| 10 envs, render, simple, states on | 12.20 GB | 12.36 GB |

### Contents
- `OmniGibson/tests/benchmark/vector_profiling.py` — runs one config, emits load time, FPS,
  Isaac vs non-Isaac step time, RAM and VRAM to `vector_output.json`
- `.github/workflows/profiling_vector.yml` — the matrix, plus a report job that diffs against a
  cached baseline and posts a sticky PR comment. Multi-env rows carry an `[xN.N]` ratio against
  the same scenario at 1 env, so the table shows how each metric scales with env count.
- `OmniGibson/omnigibson/omnigibson_5_1_0.kit` — one-line `descriptorSets` fix, see below
- `OmniGibson/omnigibson/tasks/point_navigation_task.py` — one-line fix, see below

### Included fix: PointNavigationTask landing pose
`_sample_initial_pose_and_goal_pos` returns scene-local coordinates, but the pose was applied as
a world pose. With `num_envs > 1` the holonomic base joints get driven far out of range and the
articulation diverges to NaN. Now passes `frame="scene"`. No-op at `num_envs=1`. Mirrors the
same fix in `behavior_task.py`.

Measured, not assumed — 10-env navigation, paired local runs:

| | pass rate |
|---|---|
| without fix | 0 / 5 |
| with fix | 3 / 4 |

Fisher exact one-tailed p = 0.048.

### Two BehaviorTask config defects found while validating
Both made `--task-type behavior` at high env counts impossible, and both are fixed by matching
how the challenge evaluator configures its env:

1. **Load only the rooms the activity needs**, from the same `B100_task_misc.csv` metadata
   `omnigibson/eval/utils/eval_utils.py` reads. `house_single_floor` has 21 room instances /
   595 objects; `bringing_water` needs 7. Loading the full house × `num_envs` exhausted GPU
   memory. Falls back to the full scene if the metadata is unavailable.
2. **Stop seeding BehaviorTask with the Rs_int robot pose.** `[-1.3, 0.5, 0.0]` was hand-picked
   for `Rs_int` beside the breakfast_table and was applied to every task type, landing the robot
   ~7.5 m from `house_single_floor`'s presampled pose (`[6.062, -0.888, 0.005]`) — inside
   geometry. `Environment.load()` runs `post_play_load` → `scene.reset()` → `step_physics()`
   *before* `BehaviorTask.reset()` applies the presampled pose, so that first physics step
   diverged to NaN (surfacing as the misleading `mat2quat` "Rotation matrix must not be scaled"
   assert). BehaviorTask now places the robot itself, as `omnigibson/eval/r1pro.yaml` does.

### Local validation (RTX 5090)
The full 20-job matrix was run sequentially and passed **20/20**. Rendering combos were briefly
dropped to 8 envs while diagnosing the descriptor failures and have since been restored to 10; the
final 10-env rendering config was re-run locally after the `.kit` change and completed cleanly
(`rc=0`, zero `Unable to allocate descriptor sets`), but its FPS/VRAM numbers are not reported
here because another job was occupying ~20 GB of the GPU at the time. CI is the first clean
measurement of that combo.

| scenario | load (s) | FPS | Isaac (ms) | VRAM (GB) |
|---|---|---|---|---|
| 1 env, norender, simple, states on | 26 | 19.9 | 35.97 | 10.56 |
| 1 env, render, simple, states on | 25 | 17.9 | 36.91 | 10.42 |
| 1 env, norender, behavior, states on | 103 | 10.3 | — | 18.19 |
| 8 envs, render, simple, states on | 122 | 66.5 | 84.68 | 11.94 |
| 10 envs, norender, simple, states on | 108 | 179.2 | 36.42 | 10.29 |
| 10 envs, norender, navigation, states on | 114 | 177.4 | 37.28 | 10.34 |
| 10 envs, norender, behavior, states on | 1254 | 38.6 | — | 21.29 |

1 → 10 envs gives **9.0×** throughput without rendering; Isaac step time stays flat at ~36 ms,
and VRAM barely moves with env count for the small scenes.

### Known CI failures — not from this PR
The three `behavior` jobs currently fail on `vec/merge-vector-into-main` with:

```
interactive_traversable_scene.py:122, in get_scene_loading_info
    return os.path.join(get_task_instance_path(scene_model, scene_instance))
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

The CI runners have `2026-challenge-task-instances/metadata/` but not `scenes/`, so
`get_task_instance_path` returns `None`. This surfaced because #2329 removed the 2025 fallback
(commit `8ce48195b`), whose comment read *"Remove this fallback after the self-hosted CI runners
download the 2026 challenge task instances"*. The same `TypeError` is failing
`test_multiple_envs_behavior_scene` on #2329 itself, independently of this PR. Needs either the
data on the runners or the fallback restored.

Because of this, whether 8 envs is enough for the *behavior* rendering combo is still unverified
— descriptor sets are also consumed by materials and textures, and behavior loads 435 objects per
env versus a handful for simple. Those jobs die on the task instance lookup before they get far
enough to tell.

### Notes for reviewers
- The `10 envs, norender, behavior` job takes ~21 min, almost entirely scene load.
- 10-env navigation is intermittently flaky (~25%) from r1pro random-action NaN divergence,
  independent of this PR; `fail-fast: false` is set.


stefren (approved):

wensi-ai (approved):

OmniGibson/omnigibson/tasks/point_navigation_task.py
```diff
@@ -347,8 +347,12 @@ def _reset_agent(self, env, env_indices):
             # stepping the simulator and checking for collisions. We should consider alternatives.
             initial_pos, initial_quat, goal_pos = self._sample_initial_pose_and_goal_pos(env, env_idx)
 
-            # Land the robot
-            robot.set_position_orientation(position=initial_pos, orientation=initial_quat)
+            # Land the robot. The sampled pose comes from scene.get_random_point() → trav_map.map_to_world,
+            # which returns coordinates in the scene's own local frame. With multi-scene envs, scene_N's
+            # scene_prim is offset from world origin, so we must mark the pose as scene-relative and let
+            # set_position_orientation convert to world — otherwise the holonomic base joints get set to
+            # reach a world pose well outside their reasonable range and the articulation diverges to NaN.
```
stefren, line 354: can we remove this comment? I don't think it's useful. in the vector world we basically just always need to pass `frame="scene"` right

stefren (approved):

PR #2164, opened 2026-04-30
Fix robot keep still

wensi-ai:
Currently we don't stabilize task relevant objects (robot is one of them) for seed instances. This introduces instabilities when instancing the seed instance (e.g. robot wobbles on start). We fix this by moving the code black outside the for loop. Also we add extra robot keep still to make sure the robot is stable on task start.

### DIFF
--- a/joylo/gello/robots/og_robot.py
+++ b/joylo/gello/robots/og_robot.py
@@ -271,6 +271,7 @@ class OGRobotServer:
         # Take a single step
         action = self.get_action()
         self.env.step(action)
+        self.robot.keep_still()
 
         # Set up keyboard handlers
         self._setup_keyboard_handlers()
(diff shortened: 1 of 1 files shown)


joylo/gello/robots/og_robot.py
```diff
@@ -1098,20 +1098,23 @@ def reset(self, increment_instance=True):
...
+        # They should already be stable from the sampled instance, but there is some issue where loading the state
+        # causes some jitter (maybe for small mass / thin objects?)
+        for _ in range(25):
+            og.sim.step_physics()
+            self.robot.keep_still()
+            for entity in self.env.task.object_scope.values():
+                if entity is not None and not isinstance(entity, BaseSystem):
+                    entity.keep_still()
+        self.env.scene.update_initial_file()
+
```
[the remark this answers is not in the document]
wensi-ai: We want the update here so env.reset() will reset to the stabilized scene.

joylo/gello/robots/og_robot.py
```diff
@@ -1098,22 +1099,25 @@ def reset(self, increment_instance=True):
...
+        # Try to ensure that all task-relevant objects are stable
+        # They should already be stable from the sampled instance, but there is some issue where loading the state
+        # causes some jitter (maybe for small mass / thin objects?)
+        for _ in range(25):
+            og.sim.step_physics()
+            for entity in self.env.task.object_scope.values():
+                if entity is not None and not isinstance(entity, BaseSystem):
+                    entity.keep_still()
+        self.env.scene.update_initial_file()
```
stefren, line 1116: what's the reasoning for moving this logic? it doesn't seem to change anything (only reorders with the util to update the instance label)

joylo/gello/robots/og_robot.py
```diff
@@ -1098,22 +1099,25 @@ def reset(self, increment_instance=True):
...
+
         # Reset env
         self.env.reset()
+        self.robot.keep_still()
```
stefren, line 1120: i think we should put this above with stabilizing the other objects so it gets caught in the update_initial_state and the reset? At least that seems more consistent to me.

stefren (approved):

joylo/gello/robots/og_robot.py
wensi-ai: see comments above

joylo/gello/robots/og_robot.py
wensi-ai: see comments above

yalcintur (approved):
Tested for the task that resulted in the mentioned behavior. With this branch, the behavior is stable

PR #2173, opened 2026-05-01
Fix JoyLo Home reset after slicing objects

yalcintur:
Home reset in JoyLo was updating the scene initial file on every reset, so after a sliceable object transitioned, the sliced scene could become the new reset baseline. Only refresh the baseline during initial setup or when loading a new sampled task instance.

Also wrap inactive prim reactivation in an editing_usd context so restoring the original object after a transition stays synchronized with Fabric.

### DIFF
--- a/joylo/gello/robots/og_robot.py
+++ b/joylo/gello/robots/og_robot.py
@@ -271,6 +271,7 @@ class OGRobotServer:
         # Take a single step
         action = self.get_action()
         self.env.step(action)
+        self.robot.keep_still()
 
         # Set up keyboard handlers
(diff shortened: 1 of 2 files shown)


joylo/gello/robots/og_robot.py
wensi-ai: you should remove this I think

wensi-ai (changes requested):

joylo/gello/robots/og_robot.py
yalcintur: done!

cgokmen (approved):

wensi-ai (approved):

PR #2174, opened 2026-05-03
Stabilize Rs_int profiling benchmark startup

yalcintur:
This PR makes the profiling benchmark more robust for the Rs_int + GPU dynamics + 1 robot case. It adds validation of robot articulation handles before entering the measured benchmark loop and a short unmeasured warm-up path if articulation handles are not ready immediately.


stefren (approved):

cgokmen (changes requested):
Don't merge before we discuss this - let's find out whats happening

PR #2212, opened 2026-05-15
Refresh PhysX-Fabric bindings when restoring re-added scene objects

yalcintur:
This PR rebuilds live PhysX/Fabric bindings after `Scene.restore` re-adds missing objects, before loading saved poses. This is done by a helper function that stop/plays the raw sim context, renders, and refreshes handles. Before this, readded rigid objects could have valid PhysX tensor handles while Fabric/render transforms stayed stale at the origin. A normal PhysX-to-Fabric update was not enough. The PhysX/Fabric publisher needed to be rebuilt before `load_state()` applied saved poses.

### DIFF
--- a/OmniGibson/omnigibson/simulator.py
+++ b/OmniGibson/omnigibson/simulator.py
@@ -978,7 +978,6 @@ def _launch_simulator(*args, **kwargs):
             """
             playing = self.is_playing()
             if playing:
-                self.update_handles()
                 state = self.dump_state()
 
                 # Omniverse has a strange bug where if GPU dynamics is on and the object to remove is in contact with
--- a/OmniGibson/omnigibson/simulator.py
+++ b/OmniGibson/omnigibson/simulator.py
@@ -1186,35 +1186,6 @@ def _launch_simulator(*args, **kwargs):
             RigidContactAPI.initialize_view()
             ControllableObjectViewAPI.initialize_view()
 
-        def refresh_physics_after_stage_update(self):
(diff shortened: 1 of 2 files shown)


OmniGibson/omnigibson/simulator.py
```diff
@@ -1185,6 +1186,35 @@ def update_handles(self):
...
+            self.render()
+            self.update_handles()
+
+            for scene in self.scenes:
+                for robot in scene.robots:
+                    if robot.initialized:
+                        robot.update_controller_mode()
+
+                if gm.ENABLE_TRANSITION_RULES:
+                    scene.transition_rule_api.refresh_all_rules()
```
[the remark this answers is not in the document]
stefren: hmm yeah maybe I'm missing something, why do we need to e.g. set the robot controller mode here?

OmniGibson/omnigibson/simulator.py
stefren, line 970: hm i'm kind of skeptical that we need this to be set up as a contextmanager in the first place :\

stefren (approved):

OmniGibson/omnigibson/simulator.py
yalcintur: Makes sense! Cem also suggested to remove it and make the stop play include remove objects in restore which fixed the problem that required adding update handles to removing objects.

wensi-ai (approved):

stefren (approved):

PR #2215, opened 2026-05-16
import usd pipeline fixes

asattiraju1:
- OmniGibson/omnigibson/examples/robots/import_custom_robot.py
  - Forward cfg.dataset_root from the config YAML into the import_og_asset_from_urdf call.

- OmniGibson/omnigibson/utils/asset_conversion_utils.py
  - Replace hardcoded GetChild("World").GetChild("mesh") with a generic "single Xform child" lookup.
  - Isaac names the wrapper differently per mesh format (World for OBJ, Scene for DAE, node_STL_BINARY_ for STL)

- OmniGibson/omnigibson/utils/asset_conversion_utils.py
  - Drop the per-iteration source delete after Sdf.CopySpec.
  - When multiple links share a mesh file (e.g. 4 fingers → one STL, panda left/right fingers → finger.dae), only the first link gets geometry; the source was deleted for the rest


yalcintur:
We have used this PR in the final version of the Robot import PR (#2292). Thanks Abhinav!

PR #2217, opened 2026-05-18
Decrease trash_can mass to 6.0

yalcintur:
Decrease trash_can mass to 6.0 from 9.5 as the latter is hard to be carried with R1Pro robot, causes arm bending and undesired movements that lead to collisions.


stefren (approved):
can you make sure to update the density and then also overwrite the override in the assets spreadsheet?

stefren:
actually i'll just revert my PR

PR #2220, opened 2026-05-18
Fix out-of-step refresh issue for time-dependent object states

Andorexad:
`_refresh_state_caches` is called both inside `non_physic_step` but also `step_physics`, which doesn't involve time advancement, so `toggle`, `slicer_actice` and `temperature` should not advance.

### DIFF
--- a/OmniGibson/omnigibson/object_states/slicer_active.py
+++ b/OmniGibson/omnigibson/object_states/slicer_active.py
@@ -150,7 +150,7 @@ class SlicerActive(TensorizedAbsoluteState, BooleanStateMixin):
         # Base class rebuilds OBJ_IDXS, IDX_OBJS, VALUES (with value carry-over for survivors)
         super().initialize_view()
 
-        cls._LAST_UPDATE_STEP_INDEX = -1
+        cls._LAST_UPDATE_STEP_INDEX = og.sim.step_call_index
 
         S = len(cls.IDX_OBJS)
         O = len(cls.OBJ_IDXS)
--- a/OmniGibson/omnigibson/object_states/slicer_active.py
+++ b/OmniGibson/omnigibson/object_states/slicer_active.py
@@ -308,7 +308,7 @@ class SlicerActive(TensorizedAbsoluteState, BooleanStateMixin):
 
     def _dump_state(self):
         if self.OBJ_IDXS is None or self.obj.relative_prim_path not in self.OBJ_IDXS:
-            return dict(value=True, previously_touching=False, delay_counter_seconds=0.0)
+            return dict(value=True, previously_touching=False, delay_counter=0.0)
         state = super()._dump_state()
         scene_idx = self.obj.scene.idx
         obj_idx = self.OBJ_IDXS[self.obj.relative_prim_path]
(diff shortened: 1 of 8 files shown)


OmniGibson/omnigibson/simulator.py
```diff
@@ -455,6 +455,11 @@ def __init__(
             self.pre_step_exception = None
             self.post_step_exception = None
 
+            # Monotonically advances exactly once per og.sim.step() call.
+            # Time-dependent tensorized states gate on this so their
+            # _update_values doesn't tick during step_physics(). Reset to 0 in stop().
+            self._step_call_index = 0
```
cgokmen: There's already a current_time_step_index available on og.sim, lets just use that.

OmniGibson/omnigibson/object_states/slicer_active.py
```diff
@@ -252,6 +259,11 @@ def _append_none_for_scene():
     def _update_values(cls, values):
         if cls.PREVIOUSLY_TOUCHING_WP is None:
             return
+        # Out-of-step update
+        step = og.sim.step_call_index
+        if step == cls._LAST_UPDATE_STEP_INDEX:
+            return
+        cls._LAST_UPDATE_STEP_INDEX = step
```
cgokmen: I don't quite understand the logic of this. Why is it true that we can just return when this is the case? If we can just return, why is this function getting called in the first place? What states do, or don't, need this check? It's a little confusing. Also, nothing takes into account the time difference since the last step. But doesn't that matter for things like toggle (e.g. step_physics() vs full step() time) and temperature? If you just implement the take-into-account-how-much-time-passed logic, then implicitly the zero-time-has-passed case will get covered too. Let's also move the last update step tracking into the base tensorized state class, and pass a `dt` argument into update_values (tensorized state can compute the timesteps since the last update and multiply it with the physics dt). Also lets make all time-dependent math use actual time (seconds) rather than timestep count.

cgokmen (changes requested):

wensi-ai (approved):

OmniGibson/omnigibson/object_states/tensorized_state.py
```diff
@@ -112,6 +141,16 @@ def pre_update(cls):
             return
         cls.PREV_VALUES.copy_(cls.VALUES_CPU)
 
+        # Refresh dt for time-dependent kernels. max(0, …) guards against a stop→play cycle,
+        # which resets Isaac's current_time to 0 — without the clamp the next dt would be negative.
+        import omnigibson as og  # local import to avoid module-level cycle
+
+        curr_time = og.sim.current_time
+        dt_value = max(0.0, curr_time - cls._last_update_time)
```
cgokmen: One concern here is that it's hard to tell how current_time moves when we do things like og.sim.step_physics(). Previously those steps did NOT advance any of the non-physics states, and they also did not cause later non-physics steps to advance further. Now they do the latter (because current_time does move when we do step_physics).

I think the correct way of implementing this is simply allowing the caller of the update function to provide the dt rather than trying to compute it this way. Then, for step_physics we can use a dt of 0, and for the regular non-physics step we can use the render dt. This would be in line with the existing behavior, and it would also avoid the issue where the last_update_time will not be a multiple of render dt.

OmniGibson/omnigibson/object_states/toggle.py
```diff
@@ -559,22 +563,24 @@ def _update_values(cls, values):
...
         if cls._mask_can_toggle_flat is None:
             return
         S, O = values.shape[:2]
 
         mask_flat = cls._mask_can_toggle_flat
         values_flat_wp = wp.from_torch(values.view(-1).view(th.uint8), dtype=wp.uint8)
-        steps_flat = cls._robots_can_toggle_steps.reshape((S * O,))
+        time_flat = cls._robots_can_toggle_time.reshape((S * O,))
+
+        threshold_seconds = m.CAN_TOGGLE_STEPS * og.sim.get_sim_step_dt()
```
cgokmen: Lets change this macro to also be in seconds directly (and assert _somewhere_ that it's more than 1 sim step dt)

cgokmen (approved):
Please address comments and merge

PR #2235, opened 2026-05-21
Actually build knowledgebase in pull-sheets workflow check

cgokmen:
## Summary
- The `Test if knowledgebase loads OK` step in [.github/workflows/pull-sheets-bddl.yml](.github/workflows/pull-sheets-bddl.yml) was just running `from bddl.knowledge_base import *`, which used to trigger eager loading but no longer does — population now happens inside `KnowledgeBase.__init__`. The step was passing without actually validating anything.
- Now explicitly instantiates `KnowledgeBase(verbose=False, load_wordnet=True)`, which both builds the knowledgebase and enables the deeper WordNet-based synset validation (the same flag the test fixture uses in `bddl3/tests/test_knowledgebase.py`).

## Test plan
- [ ] Workflow run completes successfully and catches knowledgebase build errors before commit.

### DIFF
--- a/.github/workflows/pull-sheets-bddl.yml
+++ b/.github/workflows/pull-sheets-bddl.yml
@@ -89,10 +89,11 @@ jobs:
       working-directory: bddl3
       run: python -m bddl.data_generation.generate_datafiles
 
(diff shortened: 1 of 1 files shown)


stefren (approved):
r

PR #2236, opened 2026-05-21
Clean up duplicate wp and th.tensor

Andorexad:
As title. Only keep wp.array version for object state class property.


cgokmen (approved):
Great work!

PR #2254, opened 2026-06-01
Brenda/sampling bddl

13RENDA:
bddl for task 99, 100, 101


bddl3/bddl/activity_definitions/installing_a_fax_machine/problem0.bddl
```diff
@@ -13,15 +14,28 @@
...
+        (future carrel.n.02_1)
+        (future carrel.n.02_2)
+        (future carrel.n.02_3)
+        (future carrel.n.02_4)
+        (future carrel.n.02_5)
+        (future carrel.n.02_6)
+        (future carrel.n.02_7)
+        (future carrel.n.02_8)
+        (future carrel.n.02_9)
+        (future carrel.n.02_10)
```
stefren: what are these futures? shouldn't they just be specified inroom?

wensi-ai: fyi future refers to an object doesn't exist now and will be generated in the future (e.g. through transition rules), which is not the case here.

wensi-ai: also what's the reason of having these many carrels?

wensi-ai (changes requested):

wensi-ai (approved):

PR #2261, opened 2026-06-05
Speed up replay obs state loading

yalcintur:
Summary
- Add batched_usd_edits() to batch nested USD edits into one Fabric sync
- Add batch_state_load to playback wrappers
- Enable batched replay state loading in replay_obs

Benchmark
- store_honey.hdf5, 3495 steps
- Batched replay loop: ~241.5s / 14.47 steps/s
- Unbatched replay loop: ~295.0s / 11.85 steps

### DIFF
--- a/OmniGibson/omnigibson/simulator.py
+++ b/OmniGibson/omnigibson/simulator.py
@@ -446,6 +446,7 @@ def _launch_simulator(*args, **kwargs):
 
             # USD edit guard: detects edits outside editing_usd() context
             self._editing_usd = False
+            self._allow_nested_usd_edits = False
             self._editing_usd_caller = None
             self._in_sim_lifecycle = 0
             self._deferred_usd_guard_error = None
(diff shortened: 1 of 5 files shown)


OmniGibson/omnigibson/simulator.py
```diff
@@ -1596,6 +1600,23 @@ def editing_usd(self, stage=None):
...
+            """
+            Batch nested USD edits into a single Fabric synchronization.
+
+            This is useful for state replay, where many low-level setters each wrap their USD writes in editing_usd().
+            The outer context keeps the USD edit guard active while deferring Fabric sync until all nested edits finish.
+            """
+            assert not self._editing_usd, "Cannot start batched_usd_edits() while another editing_usd() context is open"
+            old_allow_nested_usd_edits = self._allow_nested_usd_edits
+            self._allow_nested_usd_edits = True
+            try:
```
stefren, line 1614: i don't think we need anything allowing nested usd edits. Instead I would just call this like "defer" or "skip" synchronize to fabric and you can just set a different flag that says to skip synchronization and use it in the existing editing_usd context manager above. 
I would explicitly SynchronizeToFabric in load_replay_state in the batch case after running load state with skipping fabric sync and not create another new context manager, since I think this is definitely not a pattern we want to encourage or use widely. 
Honestly I'd even consider explicitly passing a sync_to_fabric boolean to any of the lower level setters and propagate them through here. I think the current implementation is going to be very prone to bugs. Also are we confident that this already doesn't cause any bugs where certain state is dependent on other state being already synced correctly?

PR #2262, opened 2026-06-05
Allow skipping keep_still during replay obs

yalcintur:
Summary
- Add keep_still_on_replay to playback wrappers
- Add replay_obs --skip_keep_still flag
- Keep existing behavior as the default

Benchmark
- store_honey.hdf5, 3495 steps
- keep_still + batched replay loop: ~241.5s
- skip keep_still + batched replay loop: ~196.9s


wensi-ai:
We still need keep still per @cremebrule

PR #2268, opened 2026-06-10
Let eval script work on vec env

Andorexad:
## Goal
This PR makes the BEHAVIOR evaluation code able to run several task setups at the same time inside a single simulator, instead of one at a time. The simulator can already hold multiple copies of the same scene; this change lets the evaluator actually use them, so a batch of task instances can be scored together rather than sequentially. 

  ## Changes

  **Run multiple instances at once (`omnigibson/eval/adapter.py`)**
  - The evaluator now keeps one robot, one policy, one set of scores, and one observation per parallel slot, instead of a single fixed one. Asking for one slot behaves exactly like before.
  - Loading a task setup is split into two steps: first put each instance's objects/robot into its slot, then let physics settle for the whole group at once. This is needed because settling physics moves all scenes together, so we can't settle them one by one.
  - Added `BehaviorEvalVectorEnv`, a standard environment object so an outside evaluation harness can drive the whole batch through the usual reset/step calls. A slot that finishes early waits until the rest of its group is done (we don't swap in a new instance mid-group, because loading one would disturb the others still running).
  - `make_env(...)` now hands back this batched environment.

  **Per-instance scoring (`omnigibson/metrics/*`, `omnigibson/tasks/behavior_task.py`)**
  - Scores are now computed per slot, so each parallel run reports its own result instead of everything reading the first slot's state.
  - Added a helper func on the behavior task that checks goal progress for one specific slot.

  **Make the camera setup handle more than one scene (`omnigibson/eval/wrappers/*`)**
  - The camera-resolution wrappers now set every scene's robot cameras, not just the first.

  ## Remaining Work
  - Run with a real policy


cgokmen (approved):

Andorexad:
move to #2318

PR #2274, opened 2026-06-11
Do not mark object as sleeping if it has moved

stefren:
Check whether the pose was changed or not, if the pose was changed then don't set the object to sleeping as this could cause us not to calculate the last frame where the object makes contact with another object. I think this could also possibly imply that we have an off-by-one with how we're storing / processing sleep state, but this should fix the issue without requiring re-recording.

### DIFF
--- a/OmniGibson/omnigibson/prims/entity_prim.py
+++ b/OmniGibson/omnigibson/prims/entity_prim.py
@@ -1544,6 +1544,13 @@ class EntityPrim(XFormPrim):
         return state
 
     def _load_state(self, state):
+        recorded_sleep = state.get("is_asleep", False)
+        if recorded_sleep:
+            pos, orn = state["root_link"]["pos"], state["root_link"]["ori"]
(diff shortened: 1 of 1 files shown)


OmniGibson/omnigibson/prims/entity_prim.py
```diff
@@ -1544,6 +1544,26 @@ def _dump_state(self):
         return state
 
     def _load_state(self, state):
+        recorded_asleep = state.get("is_asleep", False)
+
+        # If the object was asleep when this state was recorded, it was not moving, so its current
+        # pose/joints are already correct. Skip re-writing the kinematic state, which would otherwise
+        # flush this body's PhysX contact manifold and make contact-based states (Touching -> OnTop)
+        # read False during state playback. See gm.SKIP_ASLEEP_POSE_ON_LOAD.
```
stefren:
```suggestion
        # read False during state playback.
```

wensi-ai (approved):
lgtm

PR #2286, opened 2026-06-22
Fix lightning for turning off all lights before sleep

wensi-ai:


OmniGibson/scripts/learning/replay_obs.py
```diff
@@ -80,6 +83,38 @@ def _load_room_instances(task_name: str) -> list[str]:
...
+    light_synchronizer = LightToggleSynchronizer(env.scene)
+
+    def load_state_and_sync(*args, **kwargs):
+        result = original_load_state(*args, **kwargs)
+        light_synchronizer.sync_from_current_state()
+        return result
+
+    def step_and_sync(*args, **kwargs):
+        _, reward, terminated, truncated, info = original_step(*args, **kwargs)
+        n_render_iterations = kwargs.get("n_render_iterations", 1)
```
stefren: do we need this if we always use 1 for render_iterations?

OmniGibson/scripts/learning/replay_obs.py
```diff
@@ -80,6 +83,38 @@ def _load_room_instances(task_name: str) -> list[str]:
...
+        obs, _ = _render_and_get_obs(env, light_synchronizer, n_render_iterations=n_render_iterations)
+        return obs, reward, terminated, truncated, info
+
+    og.sim.load_state = load_state_and_sync
+    env.env.step = step_and_sync
```
stefren: i think this rewriting of methods feels pretty hacky, is there a reason we can't wire the lighting changes directly into `playback_episode` in `data_wrapper`?

OmniGibson/scripts/learning/replay_obs.py
wensi-ai: removed

OmniGibson/scripts/learning/replay_obs.py
wensi-ai: I think this is supposed to be a task-specific hot fix anyways, so it should stay in replay_obs.py rather than go directly inside data_wrapper.py

wensi-ai: Fixed.

stefren (approved):

PR #2311, opened 2026-07-16
Vectorize particle system

Andorexad:


OmniGibson/omnigibson/simulator.py
```diff
@@ -1214,11 +1216,35 @@ def sync_physx_to_fabric(self):
 
             self._sim_context._physx_fabric_interface.update(self.current_time, self.get_physics_dt())
 
-        def update_handles(self):
+        def refresh_physics_view_handles(self, reattach_unified_views=True):
```
cgokmen: I wonder - couldn't you do the same trick that you do here (don't refresh the full view if that view's objects didnt change) inside the views' initialize_view implementations themselves instead of branching this here? That way you wouldn't have to expose two different code paths in the views (one for refresh, one for initialize), the view would just pick which to do based on whether its prim paths changed?

e.g. undo the current change, and in xyzview.initialize_view check if prim path list is same as before, if it is, just refresh the view, if not, rebuild the whole thing

OmniGibson/omnigibson/scenes/scene_base.py
```diff
@@ -1104,6 +1104,13 @@ def get_system(self, system_name, force_init=True):
         if not system.initialized and force_init:
             system.initialize(scene=self)
             self.system_registry.add(system)
+            # Activating a system changes the tensorized (object, system) state layout (e.g.
+            # ContainedParticles' SYS_IDXS), so rebuild the unified/tensorized views once the current
+            # structural mutation is complete. During object initialization, rebuilding here would
+            # expose partially initialized object states; _non_physics_step calls update_handles()
+            # after the full initialization batch instead.
+            if not og.sim._objects_to_initialize:
+                og.sim.update_handles()
```
cgokmen: If it is generally true that handles should not be updated if we have objects to initialize, doesn't that mean that this check should just go in update_handles?

OmniGibson/omnigibson/systems/micro_particle_system.py
```diff
@@ -1185,6 +1211,14 @@ def _sync_particle_instancers(self, idns, particle_groups, particle_counts):
             particle_counts (list of int): Desired particle counts that should exist per instancer. Length of this
                 list should be the same length as @idns
         """
+        # A MicroPhysicalParticleSystem owns at most one instancer. Reject any state (dumped, loaded,
+        # or deserialized) that asks for more than one rather than silently reinterpreting it. A
+        # zero- or one-instancer request (including a legacy non-zero ID) is fine.
```
cgokmen, line 1216: What is the thing that _does_ need to own multiple instancers?

OmniGibson/omnigibson/object_states/covered.py
```diff
@@ -52,6 +52,12 @@ def _initialize(self):
         # Grab group name
         self._visual_particle_group = VisualParticleSystem.get_group_name(obj=self.obj)
 
+    def get_value(self, system):
+        # ContactParticles.VALUES_CPU (or the visual group count) is the source of truth;
+        # caching this threshold separately could hide a same-step particle change.
```
cgokmen: It's unclear to me what this (and its occurrences in the other files) is referring to, and what that has to do with the code below, and why this means the usual get_value cannot be used.

cgokmen (approved):

OmniGibson/omnigibson/scenes/scene_base.py
Andorexad: It actually should be "don't rebuild while there would read the mid-init objects' half-built states". Objects can sit queued between steps with nobody initializing them, and refreshing then is both safe and desirable. So instead of moving the check, I now made the invariant explicit: _non_physics_step now sets a sim.initializing_objects flag strictly around the obj.initialize() loop, and get_system defers only on that flag — refreshing eagerly in every other situation. The reason the check stays at the call site rather than inside update_handles: deferral is a property of the caller, not the function. get_system can defer because it knows the batch-end update_handles covers it, while update_handles's other callers need immediate effect.

OmniGibson/omnigibson/systems/micro_particle_system.py
Andorexad: currently there is nothing in repo that needs >1 instancers. but in future, if we want to have different collision filtering within the same substance, or per-burst set recycling for streaming sources, we would need multiple instancer. I think we can leave this now and revisit this in next pr?

PR #2318, opened 2026-07-24
Parallel eval

Andorexad:
## Summary

  This PR adds vectorized BEHAVIOR policy evaluation so multiple task instances can run concurrently in one OmniGibson process.

  It also removes redundant rendering and observation-processing work while preserving the policy’s per-step inputs.

  ## What changed

  ### Batched evaluation

  - Added `--num-envs` to evaluate multiple task instances in parallel.
  - Reworked the evaluator around one multi-environment OmniGibson environment.
  - Batch-loads, settles, resets, and steps task instances.
  - Sends one batched observation to the policy and receives one batched action tensor.
  - Tracks observations, metrics, videos, light synchronization, success, and termination independently for each environment slot.
  - Keeps finished slots inactive until the current group completes so loading another instance cannot disturb scenes that are still running.

  ### Correct per-environment metrics

  - Made task and agent metrics environment-aware.
  - Evaluates BDDL goal predicates against each environment’s own object scope.
  - Computes Q-score, simulator time, and agent distance independently for each slot.

  ### Camera and rendering improvements

  - Configures camera modalities and resolutions before environment construction, which is required because tiled-render dimensions are fixed when the sensor is created.
  - Uses the existing tiled renderer as the source of multi-environment robot-camera observations.
  - Pauses redundant individual robot-camera render products while tiled rendering is active.
  - Disables the unused 1280×720 debug viewer during headless evaluation.
  - Caches persistent Warp aliases and tiled-image reshape metadata.
  - Preserves opt-out environment variables for controlled A/B testing.

  ### Policy transport

  - Adds optional server-provided action-chunk replay.
  - Replay is disabled by default.
  - When enabled, the client validates chunk shape and verifies that the first chunk action exactly matches the server’s immediate action.
  - Servers that do not return a chunk fall back to one request per step.
  ## Why

  The previous evaluator processed task instances sequentially and repeated work that could be shared or batched:

  - simulation and rendering were run separately for every instance;
  - policy inference and websocket transport were called once per environment;
  - tiled rendering ran alongside unused individual camera products;
  - headless evaluation still constructed and rendered an unused debug viewer.

  These changes increase evaluation throughput while continuing to render and deliver fresh camera observations at every policy step. The earlier render-skipping experiment is not included because it changed policy inputs.

  ## Testing

  Local validation:

  - `OMNIGIBSON_HEADLESS=1 pytest -q tests/test_tiled_rendering.py`
    - 3 passed
  - `OMNIGIBSON_HEADLESS=1 pytest -q tests/test_multiple_envs_behavior_task.py`
    - 12 passed
  - Python compilation passed for all changed modules.
  - `git diff --check` passed.
  - Completed full-length real-GR00T rollout testing.
  - Compared tiled observations against temporarily re-enabled individual camera products for RGB and depth consistency.
  - Verified that the optimized path contains no skipped-render or stale-observation logic.

### DIFF
--- a/OmniGibson/omnigibson/eval/eval.py
+++ b/OmniGibson/omnigibson/eval/eval.py
@@ -71,6 +71,17 @@ def parse_args() -> argparse.Namespace:
         help="Number of parallel env slots; instances are evaluated this many at a time.",
     )
     parser.add_argument(
+        "--render-every",
+        type=int,
+        default=1,
+        help=(
+            "Render (and feed the policy fresh camera images) every N steps; step physics-only in "
+            "between. Set to the policy's action horizon K for chunked policies (skip-render mode) to "
+            "skip drawing on steps the policy replays. 1 = render every step (for policies that look "
+            "every step, e.g. temporal_ensemble). Ignored while writing video."
+        ),
+    )
+    parser.add_argument(
         "--max-steps",
         type=int,
         default=None,
@@ -142,6 +153,7 @@ def main() -> None:
             "mode": args.mode,
             "seed": seed,
             "num_envs": args.num_envs,
+            "render_every": args.render_every,
             "task": {"name": args.task_name},
             "robot": robot_config,
         }
--- a/OmniGibson/omnigibson/eval/eval.py
+++ b/OmniGibson/omnigibson/eval/eval.py
@@ -17,7 +17,6 @@ Example:
 """
 
 import argparse
-import json
 import logging
 import os
 from pathlib import Path
@@ -65,6 +64,12 @@ def parse_args() -> argparse.Namespace:
         help="Instance split to evaluate. Default: public_test.",
     )
     parser.add_argument("--num-rollouts", type=int, default=1, help="Rollouts per instance.")
+    parser.add_argument(
+        "--num-envs",
+        type=int,
+        default=1,
+        help="Number of parallel env slots; instances are evaluated this many at a time.",
+    )
     parser.add_argument(
         "--max-steps",
         type=int,
@@ -136,6 +141,7 @@ def main() -> None:
             "write_video": args.write_video,
             "mode": args.mode,
             "seed": seed,
+            "num_envs": args.num_envs,
             "task": {"name": args.task_name},
             "robot": robot_config,
         }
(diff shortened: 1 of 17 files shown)


kmy17518 (approved):

OmniGibson/omnigibson/eval/utils/network_utils.py
```diff
@@ -40,6 +42,7 @@ def __init__(
         scheme: str = "ws",
         api_key: Optional[str] = None,
         allow_reconnect: bool = False,
+        action_chunk_size: int = 0,
```
stefren, line 45: let's add a docstring for this below since users need to implement this policy

stefren: We also need to include clear docs about how to implement the action chunk request/response

OmniGibson/omnigibson/eval/evaluator.py
```diff
@@ -83,60 +81,112 @@ def resolve_instance_ids(task_name: str, instance_indices: list[int], mode: str
...
+        self.robot_camera_names = get_robot_camera_names(self.robots[0].name, self.robot_eval_config)
         self._validate_robot_eval_config()
         self._apply_robot_eval_settings()
         self.policy = self.load_policy()
         self.metrics = self.load_metrics()
-        self.obs = None
-        self.light_synchronizer = None
+        self.obs = [None] * self.num_envs
+        self._video_writers = [None] * self.num_envs
+        self.light_synchronizers = [None] * self.num_envs
```
stefren: hm, i'm wondering if it would be cleaner to keep a wrapper class for a single evaluator that holds these instance variables and functions and then a BatchEvaluator class that holds many evaluators, rather than changing all of the methods to take in an index

OmniGibson/omnigibson/eval/policies.py
```diff
@@ -52,17 +58,29 @@ def __init__(
         host: Optional[str] = None,
         port: Optional[int] = None,
         allow_reconnect: bool = False,
+        action_chunk_size: int = 0,
```
wensi-ai, line 64: Have we documented this somewhere?

OmniGibson/omnigibson/eval/eval.py
```diff
@@ -149,54 +171,18 @@ def main() -> None:
...
-                    )
-                    results.append(result)
-                except Exception:
-                    logger.exception(f"Instance {instance_id} rollout {rollout_id} failed.")
-                    raise
-                finally:
-                    if args.write_video:
-                        evaluator.stop_recording()
+        # Each run() pass evaluates every instance once, num_envs at a time (the slot batching lives in
+        # evaluator.run -> evaluate_instances_batched). Outer loop repeats for additional rollouts.
```
wensi-ai: Sry I don't quite get it here, do we still run batched eval if we are evaluation 10 instances once?

OmniGibson/omnigibson/metrics/metric_base.py
```diff
@@ -4,9 +4,17 @@ class MetricBase:
     each environment episode
     """
 
-    def __init__(self):
+    def __init__(self, env_idx=0):
```
stefren: hm, I think this pattern is pretty strange, because we're storing `env_idx` without having access to the environment so it's not really useful standalone information. I think it's a bad code smell that we have to keep on passing the `env` in every time we need to access the scene. I would prefer if we could even just store the scene itself instead of the index?

stefren (commented):
can we also add a test for the evaluator?

OmniGibson/omnigibson/eval/policies.py
Andorexad: Yes—documented now in 372e573d6. The websocket policy docstring points implementers to the full optional action-chunk protocol in `docs/challenge/evaluation.md`, including request/response fields, validation, fallback behavior, reset behavior, and the correctness restrictions.

OmniGibson/omnigibson/metrics/metric_base.py
Andorexad: Agreed with the ownership concern, and addressed it in 2de77e2ee. Metrics created by `BatchedEvaluator` are now bound to an `InstanceEnvAccessor`, so evaluator code calls `reset()`, `step(...)`, and `aggregate()` without repeatedly passing the shared environment. The accessor exposes the correct per-instance `scene`, `robot`, `success`, and goal-satisfaction data. Storing only the scene is insufficient for `TaskMetric`, because success and goal satisfaction live in shared vectorized task tensors and still require the logical-environment index. The old `env_idx` path remains only as a backward-compatible fallback for general `MetricsWrapper` and external callers; the batched evaluator no longer uses it.

OmniGibson/omnigibson/eval/utils/network_utils.py
Andorexad: Addressed in 372e573d6. I added the constructor docstring and a dedicated `docs/challenge/evaluation.md` section that specifies the `__action_chunk_size__` request, the `action` plus optional `action_chunk` response and shape, exact first-action equality, fallback/reset behavior, and the safety restriction to receding-horizon/open-loop action chunks (not temporal ensembling or repeated inference from stale observations).

OmniGibson/omnigibson/eval/eval.py
Andorexad: Yes. The contract is now explicit: `num_instances` must equal `num_envs`. Evaluating 10 instances with 10 environments runs one batched evaluation containing ten logical environments. An incomplete batch now fails early instead of silently changing the execution model.

OmniGibson/omnigibson/eval/evaluator.py
Andorexad: Addressed in 2de77e2ee, following the ownership intent of this suggestion. We cannot create one complete evaluator per item because `env.step`, simulator/render lifecycle, tiled sensors, and policy inference are intentionally shared and batched. Instead, `BatchedEvaluator` owns that shared work; each `InstanceEvaluationState` owns per-instance metric/video/config/runtime state; and its `InstanceEnvAccessor` exposes one logical environment. Evaluator methods now receive the per-instance state instead of parallel arrays or raw indexes.

Andorexad:
Added focused evaluator coverage in `OmniGibson/tests/test_evaluator.py`: batched state isolation, accessor-to-logical-environment mapping, metric ownership/aggregation, and the exact complete-batch contract. The file passes all four tests. I also reran real headless N=2 end-to-end integrations with the local policy, OpenPI, and GR00T; both external-policy runs used replay disabled.

stefren (approved):

PR #2327, opened 2026-08-13
Fix four multi-scene correctness bugs found by evaluating the 2025 challenge policy at num_envs=10

Andorexad:
## Summary

Ran the 2025 BEHAVIOR challenge winner's policy ([IliaLarchenko/behavior-1k-solution](https://github.com/IliaLarchenko/behavior-1k-solution)) against the vectorized evaluator at `num_envs=10` to check whether the merged parallel-eval work has correctness bugs.

**Headline: the vectorized eval framework is correct.** Once four genuine multi-scene bugs are fixed, we reproduce the winner's published success rates within noise across 9 tasks. This PR contains those four fixes, plus one environment-parity change that is explicitly flagged for decision.

## The four multi-scene bugs (please review as bug fixes)

**1. `0add17ec8` — OnFire load-state heats every flammable object to 249 °C**

`OnFire._set_value(False)` writes `Temperature = ignition_temperature − 1` (249 with the 250 default). That is correct for a user-level *extinguish* call, but `OnFire` inherits the generic `TensorizedAbsoluteState._load_state`, which routes through `_set_value(state["on_fire"])`. So restoring a saved `on_fire: False` **heats a cold flammable object to 249 °C**. `scene.reset()` → `restore()` → `load_state()` replays this on every episode reset, for every flammable object, in every scene. The popcorn bag is flammable, so it started every episode already `Heated`. Fixed as a no-op `_load_state`, matching `main`.

The same commit fixes `_link_flat_idx` being shaped `(N,)` instead of `(S, N)`, so heat-source link poses were read from the wrong scene's column.

**2. `5dc9f5359` — heat proximity used AABB-*center* distance**

The tensorized path measured distance to the target's AABB center rather than to the AABB itself, changing pre-vectorization semantics. Restored to a closest-point clamp.

**3. `16405b32c` — transition rules stored macro particle poses in world frame**

Correct for scene 0, wrong for every other scene. Now scene-relative. Adds `test_multiple_envs_transition_rules.py`.

**4. `ba105fc97` — scene-file objects baked a +100 m z offset in every `env > 0`**

The temporary scene-prim z offset was omitted from the world-frame init-pose write, so it got baked into each object's USD local transform. Later scene-relative `load_state` writes hid this for most objects, but any object that never receives one — e.g. scene furniture under the eval's partial-scene-load path, such as the stove burner in `cook_hot_dogs` — physically sat ~100 m above its scene in every env except env 0.

New tests: `test_multiple_envs_heat_states.py` (9 tests), `test_multiple_envs_transition_rules.py`.

## The AG window change — reviewer decision needed

`a21f93826` reverts #2098, restoring the assisted-grasp confirmation window. **This is not vector-specific and we are not claiming the current value is wrong.** `main`, `vector` and this branch's parent all carry the same 36-step window. .

Effective window in control steps (`_ag_grasp_counter` increments once per control step from `post_step()`, at 30 Hz control / 120 Hz physics):

| | value | effective |
|---|---|---|
| `684a8305` (2025 challenge build) | `3.0 * sim_step_dt` | **90 steps** |
| #2072 (Apr 9 2026) | `0.7 * physics_dt` | **84 steps** |
| #2098 (Apr 10 2026) | `0.3 * physics_dt` | **36 steps** |

#2072 changed the constant and the dt *together* and preserved the effective window — a deliberate re-expression, not a regression. #2098 the next day is the only real behavioural change: a 2.3× cut, one line, message "fix ag", no stated rationale. This PR reverts exactly that and leaves #2072's convention alone, so the diff is one line and grasp/release stay on one convention.

**Why we changed it:** the winner's published scores were measured against a 90-step window. Scoring the same policy against 36 and observing a gap cannot distinguish "the eval framework is wrong" from "the environment changed underneath the policy". On `make_microwave_popcorn`, restoring the AG grasp window from 36 back to 90 control steps took success from 2/5 to 5/5 on the same five instances on a trial run

## Results: 2025 policy on the vector branch, `num_envs=10`

Ten instances per task, compared against the winner's published per-task numbers.

| task | ours | winner |
|---|---|---|
| turning_on_radio | 6/10 | 6/10 |
| make_microwave_popcorn | 8/10 | 9/10 |
| moving_boxes_to_storage | 5/10 | 6/10 |
| picking_up_trash | 3/10 | 4/10 |
| wash_a_baseball_cap | 4/10 | 3/10 |
| spraying_fruit_trees | 0/10 (q 0.25), see notes below | 2/10 (q 0.30) |
| clean_boxing_gloves | 0/10, see notes below  | 2/10 |
| **total (7 tasks)** | **26/70 (37%)** | **32/70 (46%)** |
| cook_hot_dogs | see rendering note | 8/10 |
| cook_bacon | see rendering note | 7/10 |

The 7-task gap is ~1.5 SE — statistically indistinguishable. Coverage spans navigation, grasping, toggling, heavy-object transport, water/cleaning and particle spraying, across 3 different policy checkpoints.

Two results worth calling out:

- **`turning_on_radio` matched 6/10 vs 6/10 before any fix in this PR was applied.** Radio never grasps appliance doors, so it is unaffected by the AG window. That makes it the cleanest, fully unconfounded evidence that the vectorized framework itself is correct.
- **`make_microwave_popcorn` went 1/10 → 8/10** on these fixes, with all successes completing in 2224–3541 steps, inside the winner's published 2277–3541 band.

Two gap cases  `clean_boxing_gloves` and `spraying_fruit_trees`:
`clean_boxing_gloves` fails on `main` too (main 2/3 vs vector 0/4, Fisher p = 0.14, identical failure mode in instrumented rollouts)
`spraying_fruit_trees`: success rate is 0/10 but we match or beat the winner on 8/10</h2>

instance | winner q | ours q |  
-- | -- | -- | --
36 | 1.0 | 0.5 | winner better
92 | 1.0 | 0.5 | winner better
74 | 0.5 | 0.5 | tie
113 | 0.5 | 0.5 | tie
103 | 0.0 | 0.5 | we're better
145, 146, 160, 185, 219 | 0.0 | 0.0 | tie (all five)
mean q | 0.30 | 0.25 |  

## Known issue, not addressed here: tiled rendering

`cook_hot_dogs` and `cook_bacon` score 0 with tiled rendering but succeed with per-camera render products (`cook_hot_dogs` at n=3: tiled 0/12 vs per-camera 5/15, Fisher p ≈ 0.04).

This appears to be an upstream renderer issue rather than anything in `tiled_sensor.py`, which calls stock `omni.replicator.core.create.render_product_tiled`. NVIDIA documents the symptom under "Current Limitations" in Isaac Lab's tiled-rendering docs, and it matches open issues [IsaacLab #493](https://github.com/isaac-sim/IsaacLab/issues/493) (open since Jun 2024) and [IsaacLab #4431](https://github.com/isaac-sim/IsaacLab/issues/4431) (per-tile brightness differences at `num_envs > 1`, worse with more envs, gone with regular cameras).

Measured with zero actions (identical physics) and, importantly, against a repeat-run noise floor:

| comparison | signed bias | mean abs diff |
|---|---|---|
| noise floor (identical settings, repeat run) | +0.6 / +1.9 | 5.1 / 6.3 |
| tiled vs per-camera | **−4.6** | 11.1 |

The magnitude is only ~1.5× the noise floor, so magnitude alone is not convincing. What carries it is the **signed bias**: repeat runs sit near zero while tiled-vs-per-camera sits at −4.6 on all three cameras at every step. Tiled is systematically darker. Note also that the RTX renderer is not run-to-run deterministic — identical settings still differ by ~5 gray levels — so any pixel comparison without a noise floor is uninterpretable.

We tested and **refuted** the obvious explanation (the 265 px DLAA/DLSS threshold: per-camera products are 224×224 → DLAA, the tiled composite is 448×448 at n=3 → DLSS Performance). Forcing DLAA on both paths leaves the bias unchanged at −5.35. Also ruled out: staleness, geometry, gain/offset, the empty tile, indirect lighting, auto-exposure, and per-render-product settings. Mechanism remains unknown.

**Recommendation:** use per-camera render products for evaluation until this is understood. At n=10 that costs 36–60 min per task with no OOM.

## Testing

- `tests/test_multiple_envs_heat_states.py` — 9 passed (new)
- `tests/test_multiple_envs_transition_rules.py` — passed (new)
- `tests/test_ag_states.py` — 13 passed
- `tests/test_object_states.py` — full suite passed

https://claude.ai/code/session_01D25hNqSPXMZKZWgPyVvLEq

### DIFF
--- a/OmniGibson/omnigibson/object_states/temperature.py
+++ b/OmniGibson/omnigibson/object_states/temperature.py
@@ -31,7 +31,7 @@ def _incoming_heat_kernel(
     distance_thresholds: wp.array(dtype=wp.float32),  # (N_hss,)
     hss_self_temp_idx: wp.array(dtype=wp.int32),  # (N_hss,) into Temperature N — to skip self
     hss_self_inside_idx: wp.array(dtype=wp.int32),  # (N_hss,) into Inside N (for requires_inside)
-    link_flat_idx: wp.array(dtype=wp.int32),  # (N_hss,) into POSE_MATRICES — -1 if N/A
+    link_flat_idx: wp.array2d(dtype=wp.int32),  # (S_hss, N_hss) into POSE_MATRICES — -1 if N/A
     link_local_offset: wp.array(dtype=wp.vec3),  # (N_hss,) heat element offset in link local frame
     temp_to_aabb_idx: wp.array(dtype=wp.int32),  # (N_temp,) Temperature N → AABB N
     temp_to_inside_idx: wp.array(dtype=wp.int32),  # (N_temp,) Temperature N → Inside N
@@ -67,9 +67,10 @@ def _incoming_heat_kernel(
             return
         if inside_values[s, target_idx_in_inside, src_idx_in_inside] == wp.uint8(0):
             return
-    # point sources: target's AABB center must be within the distance threshold of the heat element
+    # point sources: target's AABB center must be within the distance threshold of the heat element.
+    # The element link is a different rigid body in every scene, so its pose index is per (s, h).
     else:
-        li = link_flat_idx[h]
+        li = link_flat_idx[s, h]
         if li < wp.int32(0):
             return
         link_pose = pose_matrices[li]
--- a/OmniGibson/omnigibson/object_states/temperature.py
+++ b/OmniGibson/omnigibson/object_states/temperature.py
@@ -67,7 +67,7 @@ def _incoming_heat_kernel(
             return
         if inside_values[s, target_idx_in_inside, src_idx_in_inside] == wp.uint8(0):
             return
-    # point sources: target's AABB center must be within the distance threshold of the heat element.
+    # point sources: the target's AABB must be within the distance threshold of the heat element.
     # The element link is a different rigid body in every scene, so its pose index is per (s, h).
     else:
         li = link_flat_idx[s, h]
@@ -79,11 +79,20 @@ def _incoming_heat_kernel(
             link_pose, wp.vec4(local_offset[0], local_offset[1], local_offset[2], wp.float32(1.0))
         )
         # Distance from the heat element to the CLOSEST POINT of the target's AABB (element
-        # position clamped into the box). This mirrors the pre-vectorization semantics, which
-        # used a PhysX overlap_sphere(threshold) collision query — any surface contact with the
-        # sphere counted. Using the AABB *center* instead silently shrinks every source's reach
-        # by the target's own extent (e.g. a sausage in a pan on a burner grazes the sphere but
-        # its center is beyond the threshold, so it would never cook).
+        # position clamped into the box).
+        #
+        # This is an APPROXIMATION of the pre-vectorization behaviour, which ran a PhysX
+        # overlap_sphere(threshold) scene query against the target's actual COLLISION GEOMETRY
+        # (heat_source_or_sink.py on main). Since a target's AABB always contains its geometry,
+        # distance-to-AABB <= distance-to-geometry, so this test is slightly MORE permissive than
+        # overlap_sphere — most visibly for long, thin or diagonally-oriented targets whose AABB
+        # is much larger than the mesh itself.
+        #
+        # It is nonetheless much closer than what it replaces: testing the AABB *center* shrinks
+        # every source's reach by the target's own extent (e.g. a sausage in a pan on a burner
+        # grazes the sphere, but its center is beyond the threshold, so it would never cook).
+        # An exact port would need per-(source, target) mesh queries inside this kernel, which is
+        # not available to Warp here; see the PR discussion for the trade-off.
         qx = wp.clamp(source_world_position[0], aabb_values[s, target_aabb_idx, 0], aabb_values[s, target_aabb_idx, 3])
         qy = wp.clamp(source_world_position[1], aabb_values[s, target_aabb_idx, 1], aabb_values[s, target_aabb_idx, 4])
         qz = wp.clamp(source_world_position[2], aabb_values[s, target_aabb_idx, 2], aabb_values[s, target_aabb_idx, 5])
--- a/OmniGibson/omnigibson/object_states/temperature.py
+++ b/OmniGibson/omnigibson/object_states/temperature.py
@@ -78,14 +78,18 @@ def _incoming_heat_kernel(
         source_world_position = wp.mul(
             link_pose, wp.vec4(local_offset[0], local_offset[1], local_offset[2], wp.float32(1.0))
         )
-        # get cx, cy, cz: center x, y, z of target's aabb bounding box
-        cx = (aabb_values[s, target_aabb_idx, 0] + aabb_values[s, target_aabb_idx, 3]) * wp.float32(0.5)
-        cy = (aabb_values[s, target_aabb_idx, 1] + aabb_values[s, target_aabb_idx, 4]) * wp.float32(0.5)
-        cz = (aabb_values[s, target_aabb_idx, 2] + aabb_values[s, target_aabb_idx, 5]) * wp.float32(0.5)
-        # get dx, dy, dz: delta of source position and target's center on 3 directions
-        dx = source_world_position[0] - cx
-        dy = source_world_position[1] - cy
-        dz = source_world_position[2] - cz
+        # Distance from the heat element to the CLOSEST POINT of the target's AABB (element
+        # position clamped into the box). This mirrors the pre-vectorization semantics, which
+        # used a PhysX overlap_sphere(threshold) collision query — any surface contact with the
+        # sphere counted. Using the AABB *center* instead silently shrinks every source's reach
+        # by the target's own extent (e.g. a sausage in a pan on a burner grazes the sphere but
+        # its center is beyond the threshold, so it would never cook).
+        qx = wp.clamp(source_world_position[0], aabb_values[s, target_aabb_idx, 0], aabb_values[s, target_aabb_idx, 3])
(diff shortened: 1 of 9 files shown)


OmniGibson/omnigibson/robots/robot.py
```diff
@@ -78,7 +78,32 @@
...
+# steps scales with the PHYSICS rate. At the eval's 30 Hz control / 120 Hz physics it is 90 steps.
+#
+# Why the window matters: non-root links of fixed-base objects are exempt from
+# ASSIST_GRASP_MASS_THRESHOLD (see _get_assisted_grasp_joint_type), so for appliance and cabinet
+# doors the window length is the ONLY gate against latching. At 36 steps a ~1.2 s incidental brush
+# while reaching past a microwave door establishes a SphericalJoint that tethers the arm there for
+# the rest of the episode. On make_microwave_popcorn, across 14 instrumented episodes, 6/6 failures
+# had spuriously AG-grasped the microwave and 0/8 successes had (Fisher exact p=0.0003); a
+# controlled A/B changing only this window went 2/5 -> 10/10 (Fisher exact p=0.022).
+m.GRASP_WINDOW = 0.75
```
stefren: hmm i don't think we should revert this here as part of this PR cc @wensi-ai since you landed the original PR?

OmniGibson/omnigibson/object_states/temperature.py
```diff
@@ -67,24 +67,29 @@ def _incoming_heat_kernel(
...
+        # used a PhysX overlap_sphere(threshold) collision query — any surface contact with the
+        # sphere counted. Using the AABB *center* instead silently shrinks every source's reach
+        # by the target's own extent (e.g. a sausage in a pan on a burner grazes the sphere but
+        # its center is beyond the threshold, so it would never cook).
+        qx = wp.clamp(source_world_position[0], aabb_values[s, target_aabb_idx, 0], aabb_values[s, target_aabb_idx, 3])
+        qy = wp.clamp(source_world_position[1], aabb_values[s, target_aabb_idx, 1], aabb_values[s, target_aabb_idx, 4])
+        qz = wp.clamp(source_world_position[2], aabb_values[s, target_aabb_idx, 2], aabb_values[s, target_aabb_idx, 5])
+        dx = source_world_position[0] - qx
+        dy = source_world_position[1] - qy
+        dz = source_world_position[2] - qz
```
stefren, line 104: is this correct? The overlap_sphere method seems like it would be calculating closest point to the object, rather than closest point to the object's AABB bounds?

OmniGibson/omnigibson/systems/system_base.py
```diff
@@ -359,8 +384,14 @@ def _load_state(self, state):
         self.max_scale = state["max_scale"]
 
         # Load the poses
-        setter = self.set_particles_local_pose if self._store_local_poses else self.set_particles_position_orientation
-        setter(positions=state["positions"], orientations=state["orientations"])
+        if self._store_local_poses:
```
stefren, line 403: should we be storing a boolean of whether we're using local poses in the dumped state? otherwise how would we confirm whether we know if the saved position state is local or non-local?

OmniGibson/omnigibson/robots/robot.py
```diff
@@ -78,7 +78,32 @@
 m.MAX_AG_DEFAULT_GRASP_POINT_PROP = 0.95
 m.AG_DEFAULT_GRASP_POINT_Z_PROP = 0.4
 m.CONSTRAINT_VIOLATION_THRESHOLD = 0.1
-m.GRASP_WINDOW = 0.3
+# REVIEWER DECISION NEEDED — environment parity for the 2025 challenge policy, not a bug fix.
```
wensi-ai: I think we should keep the 2026 number (0.3), but @stefren can confirm.

OmniGibson/omnigibson/robots/robot.py
wensi-ai: Let's still use 0.3

OmniGibson/omnigibson/systems/system_base.py
```diff
@@ -375,18 +425,20 @@ def serialize(self, state):
...
         state_dict["n_particles"] = n_particles
-        state_dict["min_scale"] = state[1:4]
-        state_dict["max_scale"] = state[4:7]
-        state_dict["positions"] = state[7 : 7 + len_positions].reshape(-1, 3)
-        state_dict["orientations"] = state[7 + len_positions : 7 + len_positions + len_orientations].reshape(-1, 4)
+        state_dict["uses_local_poses"] = bool(state[1])
+        state_dict["min_scale"] = state[2:5]
+        state_dict["max_scale"] = state[5:8]
+        state_dict["positions"] = state[8 : 8 + len_positions].reshape(-1, 3)
+        state_dict["orientations"] = state[8 + len_positions : 8 + len_positions + len_orientations].reshape(-1, 4)
```
stefren: this doesn't seem backwards compatible right? i think we'll no longer be able to read previously dumped states?

cgokmen: it's really important that this kind of breaking deserialization logic is gated and the old format also remains readable somehow

Andorexad: gated this in most recent commits

OmniGibson/omnigibson/object_states/temperature.py
Andorexad: agree. i modifed the kernel in the most recent commit

stefren (approved):

wensi-ai (approved):

PR #2329, opened 2026-08-18
Merge vector into main

Andorexad:


docs/omnigibson/object_states.md
```diff
@@ -75,10 +75,10 @@ These are object states that are agnostic to other objects in a given scene.
     </tr>
     <tr>
         <td valign="top" width="60%">
-            [**`VerticalAdjacency`** / **`HorizontalAdjacency`**](../reference/object_states/adjacency.md)<br><br>  
-            The nearby objects that are considered adjacent to the object, either in the +/- global Z axis or +/- global XY plane.<br><br>
+            [**`Adjacency`**](../reference/object_states/adjacency.md)<br><br>  
+            Pairwise relation indicating whether `other` is reachable from `self`'s AABB center via a ray in any of 22 directions: 2 vertical (+Z, −Z) and 20 horizontal (5 equally-spaced coordinate planes × 2 orthogonal axes × +/− signs).<br><br>
             <ul>
-                <li>`get_value()`: returns `AxisAdjacencyList`, a namedtuple with `positive_neighbors` and `negative_neighbors` each of which are lists of nearby objects</li>
+                <li>`get_value(other)`: returns a 22-element bool tensor; `value[k]` is True iff `other` is hit when ray-casting from `self`'s AABB center along direction `k`. Layout: `k=0` is +Z, `k=1` is −Z, `k=2..11` are the 10 horizontal axes' positive directions, `k=12..21` are the 10 horizontal axes' negative directions.</li>
```
stefren: why are there so many horizontal planes but only one vertical? also how does this compare to previous definition of HorizontalAdjacency?

joylo/gello/robots/og_robot.py
```diff
@@ -758,7 +764,7 @@ def _process_button_inputs(self):
         if button_a_state and not self._button_toggled_state["a"]:
             for obj in self.task_irrelevant_objects:
                 obj.visible = not obj.visible
-            task_objects = [obj for obj in self.env.task.object_scope.values() if obj is not None]
+            task_objects = [obj for obj in self.env.task.object_scope[0].values() if obj is not None]
```
stefren: should we rename this to be `object_scopes` or `object_scopes_per_env` so that it's clear that it's indexed across environments?

OmniGibson/omnigibson/tasks/task_base.py
```diff
@@ -173,130 +179,146 @@ def _create_reward_functions(self):
...
-        # By default, reset reward, done, and info
-        self._reward = None
-        self._done = False
-        self._success = False
-        self._info = None
+        if self._reward is None:
+            # First-time init
+            self._reward = th.zeros(self._num_envs, dtype=th.float32)
+            self._done = th.zeros(self._num_envs, dtype=th.bool)
+            self._success = th.zeros(self._num_envs, dtype=th.bool)
```
stefren, line 216: nit, we should just set this in the init right?

OmniGibson/omnigibson/termination_conditions/predicate_goal.py
```diff
@@ -22,23 +25,29 @@ def __init__(self, check_goal_fn):
         # Run super
         super().__init__()
 
-    def reset(self, task, env):
+    def reset(self, task, env, env_indices):
         # Run super first
-        super().reset(task, env)
+        super().reset(task, env, env_indices)
```
stefren, line 30: what's the use case for only resetting certain indices?

OmniGibson/omnigibson/utils/asset_utils.py
```diff
@@ -143,7 +143,17 @@ def get_task_instance_path(scene_name, instance_name, mode="train"):
...
-    return task_instances_path_2026 if os.path.exists(task_instances_path_2026) else None
+    if os.path.exists(task_instances_path_2026):
+        return task_instances_path_2026
+
+    if mode != "train":
+        return None
+
+    # TODO: Remove this fallback after the self-hosted CI runners download the 2026 challenge task instances.
+    task_instances_path_2025 = os.path.join(
+        gm.DATA_PATH, "2025-challenge-task-instances", "scenes", scene_name, "json", f"{instance_name}.json"
```
stefren: why did we need to change this in this PR?

OmniGibson/omnigibson/utils/deprecated_utils.py
```diff
@@ -1055,3 +1055,59 @@ def colorize_bboxes(bboxes_2d_data, bboxes_2d_rgb, num_channels=3):
         rgb_img_draw.rectangle([(bbox_2d[1], bbox_2d[2]), (bbox_2d[3], bbox_2d[4])], outline=outline, width=2)
     bboxes_2d_rgb = np.array(rgb_img)
     return bboxes_2d_rgb
+
+
+@wp.kernel(enable_backward=False)
```
stefren: i don't think we should put anything in this file called deprecated_utils? 😓

OmniGibson/scripts/sampling/sample_robot_pose.py
```diff
@@ -98,12 +98,12 @@ def sample_robot_poses(env) -> Dict[str, List[Dict]]:
     Returns:
         Dictionary mapping "robot" to list of pose dicts
     """
-    # Find the reference object from initial conditions
+    # Find the reference object from initial conditions (single-env script: scene 0)
```
stefren: should we assert that the environment is single-env?

OmniGibson/tests/test_multiple_envs_behavior_infra_api.py
```diff
@@ -0,0 +1,226 @@
+"""
+test_multiple_envs_behavior_infra_api.py
```
stefren: what does infra api mean here? should we just call this one the base `test_multiple_envs`? it's just creating and tearing down an environment

OmniGibson/tests/test_multiple_envs_behavior_infra_api.py
```diff
@@ -0,0 +1,226 @@
...
+# This test must run *after* every test that uses `behavior_env`, because it
+# tears down the shared module-scope env to build a 1-env one. Pytest collects
+# tests in source order, so keeping this class at the bottom of the file is
+# what guarantees the correct ordering.
+
+
+class TestBehaviorSingleEnv:
+    """BehaviorTask with num_envs=1. Cannot share the module-scope 2-env fixture."""
+
+    def test_single_env_behavior(self):
```
stefren: don't we have a lot of other tests that already test single env behavior?

OmniGibson/tests/test_multiple_envs_behavior_infra_api.py
```diff
@@ -0,0 +1,226 @@
...
+    def test_scenes_spatially_separated(self, behavior_env):
+        """BehaviorTask scenes occupy different spatial regions."""
+        env = behavior_env
+
+        scene_positions = [s.get_position_orientation()[0] for s in env.scenes]
+        for i in range(len(scene_positions)):
+            for j in range(i + 1, len(scene_positions)):
+                dist = th.norm(scene_positions[i] - scene_positions[j])
+                print(f"  Scene {i} <-> Scene {j} distance: {dist:.2f}")
+                assert dist > 1.0, f"Scenes {i} and {j} are too close: {dist:.2f}"
```
stefren: should we test anything more specfic about where the scenes are located relative to each other? maybe that the robot is being placed in the right location in each scene?

OmniGibson/tests/test_multiple_envs_behavior_task.py
```diff
@@ -0,0 +1,464 @@
+"""
```
stefren, line 1: can we combine some of these tests with the other suites? it seems like it's semi-expensive to instantiate these scenes and then each of the test suites actually tests very little

OmniGibson/tests/test_multiple_envs_dummy_api.py
```diff
@@ -0,0 +1,195 @@
...
+        num_envs = 3
+        env = setup_multi_environment(num_of_envs=num_envs)
+
+        scene_positions = [s.get_position_orientation()[0] for s in env.scenes]
+        for i in range(len(scene_positions)):
+            for j in range(i + 1, len(scene_positions)):
+                dist = th.norm(scene_positions[i] - scene_positions[j])
+                print(f"  Scene {i} <-> Scene {j} distance: {dist:.2f}")
+                assert dist > 1.0, f"Scenes {i} and {j} are too close: {dist:.2f}"
+        og.clear()
```
stefren: isn't this test replicated in the other suite?

stefren (commented):
review so far, i still need to look more deeply at the kernel implementations

.github/workflows/tests.yml
```diff
@@ -72,7 +72,17 @@ jobs:
...
+          - test_multiple_envs_grasp
+          - test_multiple_envs_point_navigation
+          - test_multiple_envs_point_reaching
+          - test_multiple_envs_behavior_infra_api
+          - test_multiple_envs_behavior_infra_scene
+          - test_multiple_envs_behavior_task
+          - test_multiple_envs_heat_narrow_phase
+          - test_multiple_envs_heat_states
+          - test_multiple_envs_transition_rules
           - test_object_removal
```
wensi-ai, line 80: Have we included the updated profiling test on CI?

OmniGibson/omnigibson/tasks/task_base.py
Andorexad: `_num_envs` isn't available in init. the task is constructed from config before the env exists, so it's None there (line 56) and only gets set in load() (line 115). So these can't be allocated in the constructor.
What do you think about allocating them in `load()` right after `self._num_envs = env.num_env` ?

docs/omnigibson/object_states.md
Andorexad: This is because of the downstream users of Adjacency. For horizontal ones, only OnTop / Under use it, and they need to know "is other directly above / directly below me", so we only need ±Z.
Horizontal used by NextTo, which asks "is other beside me at any bearing". we need a fan around the XY plane. 

Difference compared to previous: old = 5 planes × 2 axes × ± = 20 ray slots, but the linspace(0, π/2, 5) endpoints collide (0°/180° duplicated, 90° appears twice), so really 16 distinct directions at 22.5° for 20 casts. New = 10 distinct directions at 36°, no duplicates. Half the rays for a bit less angular resolution.

OmniGibson/tests/test_multiple_envs_behavior_infra_api.py
Andorexad: modified and refactored test files

OmniGibson/omnigibson/termination_conditions/predicate_goal.py
Andorexad: Envs in a vec env terminate at different times, so Environment.step() auto-resets only the done ones (self.reset(env_indices=done_indices)), and that propagates down to every termination condition and reward function.

.github/workflows/tests.yml
Andorexad: update in #2152

stefren (approved):

PR #2334, opened 2026-08-23
Add fail-closed pre-action guard wrapper

zoahdev:
## Summary

- add `ActionGuardWrapper`, a generic pre-action policy hook for OmniGibson environments
- prevent the wrapped environment from stepping on denial, malformed guard output, or guard exceptions
- preserve structured rejection reasons and metadata for evaluation and auditing
- add focused tests that exercise the wrapper without launching Isaac Sim

## Motivation

Embodied-agent evaluation increasingly needs a boundary where safety policies, human approval, workspace constraints, or external authorization can reject an action before it reaches physics. A generic wrapper provides that boundary without coupling OmniGibson to any policy vendor, network service, or cryptography implementation.

## API

The guard has signature `guard(env, action)` and may return either a boolean or a structural result exposing:

- `allowed: bool`
- `reason: str`
- `metadata: Mapping[str, Any]`

Rejected actions raise `ActionRejectedError`; the wrapped environment is never stepped. The normalized result is retained as `last_decision` for logging and evaluation.

## Validation

- Python source compilation passed
- standalone allow, deny, and guard-exception validation passed
- tests avoid launching Isaac Sim and do not require a GPU

Full OmniGibson integration tests were not run locally because the required Isaac Sim / NVIDIA environment is not available. The change is intentionally limited to the generic simulator integration point.

### DIFF
--- a/OmniGibson/omnigibson/envs/__init__.py
+++ b/OmniGibson/omnigibson/envs/__init__.py
@@ -1,3 +1,4 @@
+from omnigibson.envs.action_guard_wrapper import ActionGuardDecision, ActionGuardWrapper, ActionRejectedError
 from omnigibson.envs.data_wrapper import DataWrapper, DataPlaybackWrapper
 from omnigibson.envs.hdf5_data_wrapper import HDF5CollectionWrapper, HDF5PlaybackWrapper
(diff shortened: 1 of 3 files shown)


wensi-ai (commented):
Thanks for the contribution. We do not currently have an approved design or use case for an external pre-action authorization boundary, and the proposed in-process callback does not provide the stated isolation guarantees.

PR #2339, opened 2026-08-25
Fixes found by replaying recorded demos at num_envs > 1

Andorexad:
**1. `Scene._load_state` collapsed every scene into scene 0** — `scene_base.py`

It applied a state's recorded scene-pose header by *moving the scene prim to it*, then loaded object poses raw in world frame. Single-scene recordings carry the origin, so at `num_envs > 1` every scene's objects were injected into scene 0's cell, where they physically collide.

Now the recorded header is treated as the frame the state was captured in: the scene keeps its own prim pose and pose-bearing root-link fields (position/orientation, linear/angular velocity, cloth particle positions/velocities) are re-based by `rel = cur ∘ rec⁻¹`. Particle systems velocity are stored raw in world frame and are now rotated by rel_ori.

**2. Warmup contact capture at playback frequencies** — `simulator.py`

This bug only manifests at physics frequency == rendering frequency, which is why previous test with eval at 120/30 Hz never saw it.
At depth ≥ 2 the warmup captures are harmless only because `play()`'s following `update_handles()` zeroes `_PENDING_STEPS`. the invariant is protected by rebuild ordering, not by design, so any `play()` path without a following rebuild would silently fold warmup contacts into the first real step.

**3. / 4. Vacated tensorized columns** — `tensorized_state.py`, `toggle.py`

Mid-play `scene.remove_object` at `num_envs > 1` leaves `IDX_OBJS[scene][col] = None` while another scene keeps the column alive (`OBJ_IDXS` is keyed by `relative_prim_path`, shared across scenes); `TensorizedState.post_update` then dereferenced it. `ToggledOn.post_update` overrides the base instead of extending it, so the base guard does not protect that path — and `visual_markers[scene][col]` is legitimately `None` for assets with no togglebutton meta link. Both now match the sibling guards in `slicer_active.py` and `aabb.py`. 

**5. Transition-added objects deserialized with the wrong state schema** — `data_wrapper.py`

State is a flat float array read back through a shared cursor, so an object that consumes the wrong count corrupts every object after it. #2260's `_recorded_non_kin_state_names` shim aligns each object's schema with the recording's, but only for objects present in the recorded *initial* scene file. Objects created mid-episode by a recorded transition — an egg half, a wood chunk — are never stamped, so they deserialize with the current code's schema (5 states / 24 floats) against the recording's (2 / 16), and the walk fails on the first row containing them. Now they are stamped with the union of the recording's non-kin state vocabulary: a name absent from that union was never serialized by the collection code, so `deserialize` skips exactly what the recording could not contain. This blocked the entire slicing/chopping family from replaying at any `num_envs`.


OmniGibson/scripts/learning/vector_replay/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""Multi-env (vectorized) replay of recorded BEHAVIOR demos.
```
stefren, line 1: i'm not sure it's common to put this comment in an init file, or at least it's not clear to me that the editors of this module will see this comment here

stefren (commented):
can we go through the docstrings and comments added by this PR and make them shorter and more human-readable?

OmniGibson/omnigibson/object_states/adjacency.py
```diff
@@ -16,28 +16,31 @@
...
 #
-# VALUE is (S, N, N, 12) bool tensor populated by Warp ray casts against per-link wp.Mesh.
+# VALUE is (S, N, N, 18) bool tensor populated by Warp ray casts against per-link wp.Mesh.
 #
 # Axis layout:
 #   k=0       : +Z   (other above self)
 #   k=1       : -Z   (other below self)
-#   k=2..11   : 10 horizontal directions evenly spaced on the XY plane,
-#               direction k has angle (k-2) * 2π / 10 (so k=2 is +X, k=7 is -X).
+#   k=2..9    : positive directions of the 8 unique legacy horizontal axes
```
wensi-ai, line 33: what's the reason for this change?

OmniGibson/omnigibson/scenes/scene_base.py
```diff
@@ -1255,12 +1276,104 @@ def _load_state(self, state):
...
+                # match, this branch is skipped and the poses load unchanged.
+                inv_rec_pos, inv_rec_ori = T.invert_pose_transform(rec_pos, rec_ori)
+                rel_pos, rel_ori = T.pose_transform(cur_pos, cur_ori, inv_rec_pos, inv_rec_ori)
+                registry_state = self._rebase_registry_state_poses(registry_state, rel_pos, rel_ori)
+            self._registry.load_state(state=registry_state, serialized=False)
         else:
             self._registry.load_state(state=state, serialized=False)
 
+    @staticmethod
+    def _rebase_registry_state_poses(registry_state, rel_pos, rel_ori):
```
wensi-ai, line 1298: I wonder in the future whether we should just store everything w.r.t. scene origin instead.

wensi-ai (approved):

OmniGibson/omnigibson/object_states/adjacency.py
Andorexad: by having 8 axes we have 16 directions. now is the same as current main branch definition for horizontal adjacency. otherwise it will cause regression in Touching state check

OmniGibson/omnigibson/scenes/scene_base.py
Andorexad: yeah agree. this func is just a compatibility layer for existing recordings

PR #2346, opened 2026-09-11
Update data retention limits to 1 day

stefren:
We already dropped the settings for logs to 1 day retention, so there probably isn't a good use case for keeping these artifacts past then as well.

Also we seem to be hitting storage limit issues, likely due to the webpage build that we upload.


stefren:
hmmm actually this seems to already be the default behavior?

